PR #2, opened 2021-04-25
CLI Hardhat plugin

leomassazza:
- Imported interactive-ui as a hardhat plugin 
- Added core-js
- Removed any reference to synthetix v2
- Moved some functions to core-js


packages/cli/.gitignore
```diff
@@ -1 +1,112 @@
+# Logs
```
ajsantander, line 1: I'd remove this generic gitignore and replace it with a minimal version that only ignores what actually needs to be ignored.

packages/cli/README.md
```diff
@@ -0,0 +1,57 @@
+## hh-cli (Hardhat CLI)
```
ajsantander, line 1: This file can be deleted for now.

packages/cli/tasks-info.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  TASK_INTERACT_NAME: 'interact',
+  TASK_INTERACT_DESC: 'Spins up an interactive ui to interact with contracts.',
+
+  TASK_NOMINATE_OWNER_NAME: 'nominate-owner',
```
ajsantander, line 5: @leomassazza what do you think about removing all unimplemented tasks for now? Then we can focus on the interactive task fully and interpret anything present as functional.

packages/cli/tasks/accept_ownership.js
```diff
@@ -0,0 +1,18 @@
+/*global task*/
+/*eslint no-undef: "error"*/
+/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "[NAME,DESC]" }]*/
+
+const {
+  TASK_INTERACT_NAME,
```
ajsantander, line 6: I don't think that every task needs to know about every other task, so just import what's needed?

packages/core-js/utils/deploymentHelper.js
```diff
@@ -0,0 +1,28 @@
...
+
+function getSource({ deploymentData, contract }) {
+  if (contract) return deploymentData.sources[contract];
+  else return deploymentData.sources;
+}
+
+function getDeploymentFilePath({ artifactsPath, network, instance, jsonName }) {
+  console.log({ artifactsPath, network, instance, jsonName });
+  return path.join(artifactsPath, network, instance, jsonName);
+}
```
ajsantander, line 19: nit: linter not picking this up?

packages/core-js/utils/getEvents.js
```diff
@@ -0,0 +1,26 @@
+async function getPastEvents({ contract, eventName, provider, fromBlock, toBlock }) {
```
ajsantander, line 1: Is the interactive task using this? If not, I'd consider removing the file for now. We can always go back to the original CLI implementation and grab it.

packages/cli/.gitignore
leomassazza: Done

packages/core-js/utils/deploymentHelper.js
leomassazza: I think linter doesn't find any warning here :) Good catch

packages/core-js/utils/getEvents.js
leomassazza: I'll check and remove if not

packages/cli/tasks-info.js
leomassazza: ok

packages/cli/README.md
leomassazza: Done

packages/cli/tasks/accept_ownership.js
leomassazza: I removed not implemented tasks and the imports from the implemented ones

packages/core-js/utils/deploymentHelper.js
leomassazza: Fixed

packages/core-js/utils/getEvents.js
leomassazza: Don't needed. Removed.

packages/core-js/utils/prettyLog.js
```diff
@@ -0,0 +1,58 @@
+const { green, red, cyan, gray } = require('chalk');
```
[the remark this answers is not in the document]
leomassazza: Agree. I moved it as private utils for the cli package

ajsantander (approved):

PR #8, opened 2021-08-30
Add basic contracts to sample-project 

leomassazza:


packages/deployer/extensions/delete.me
```diff
@@ -0,0 +1 @@
+// Dummy file to include the folder in git
```
ajsantander, line 1: I wouldn't delete these. I think they're there to keep the folders around in version control.

packages/deployer/test/sample-project/contracts/libraries/UpgradeableProxy.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: MIT
```
ajsantander, line 1: This UpgradeableProxy.sol (as well as Proxy.sol) is too "OpenZeppelin-y". We want something less abstract, more to the point, that uses the namespace architecture directly, and that actually has an anti-bricking mechanism, like this: https://github.com/Synthetixio/synthetix-v3/blob/master/packages/deployer/test/sample-project/contracts/modules/UpgradeModule.sol

packages/deployer/test/sample-project/contracts/modules/AModule.sol
```diff
@@ -0,0 +1,21 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "./BModule.sol";
+import "../mixins/ModulesMixin.sol";
+
+contract AModule is ModulesMixin {
```
ajsantander, line 7: Can we stick to the modules in the MVP v3.0.0 spec?

ajsantander (changes requested):
@leomassazza let's discuss, I think there's several things here that wont work

packages/deployer/extensions/delete.me
leomassazza: Yes, I added them to get the folders in version control while we doesn't have any script there. As soon as we add something real we should delete the files. Otherwise hardhat will fail when our tasks scan for those folders.

mjlescano: For this case, is common practice to leave a `.gitkeep` file, which really is nothing more than something to track

packages/deployer/test/sample-project/contracts/mixins/ModulesMixin.sol
```diff
@@ -0,0 +1,17 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../storage/ProxyStorage.sol";
+import "../interfaces/IRouter.sol";
+
+contract ModulesMixin is ProxyStorageNamespace {
```
ajsantander, line 7: This is an experimental type 3 inter module implementation that we decided to drop. We should do research on this, but its not part of 3.0.0

packages/deployer/test/sample-project/contracts/storage/ProxyStorage.sol
```diff
@@ -0,0 +1,30 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+contract ProxyStorageNamespace {
```
ajsantander, line 4: Can we name these 'ProxyNamespace'? I think 'Storage' is a bit redundant and verbose.

packages/deployer/test/sample-project/contracts/Main.sol
```diff
@@ -0,0 +1,8 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "./libraries/UpgradeableProxy.sol";
+
+contract Main is UpgradeableProxy {
```
ajsantander, line 6: I would code a and use UUPS proxy locally

packages/deployer/test/sample-project/contracts/interfaces/IRouter.sol
```diff
@@ -0,0 +1,6 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+interface IRouter {
```
ajsantander, line 4: Delete

packages/deployer/test/sample-project/contracts/libraries/Address.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Collection of functions related to the address type
+ */
+library Address {
```
ajsantander, line 7: I would remove this for now and, if you wanna use it in the UpgradeModule, do this as an internal function.

packages/deployer/test/sample-project/contracts/libraries/Proxy.sol
```diff
@@ -0,0 +1,83 @@
...
+ * @dev This abstract contract provides a fallback function that delegates all calls to another contract using the EVM
+ * instruction `delegatecall`. We refer to the second contract as the _implementation_ behind the proxy, and it has to
+ * be specified by overriding the virtual {_implementation} function.
+ *
+ * Additionally, delegation to the implementation can be triggered manually through the {_fallback} function, or to a
+ * different contract through the {_delegate} function.
+ *
+ * The success and return data of the delegated call will be returned back to the caller of the proxy.
+ */
+abstract contract Proxy {
```
ajsantander, line 15: Do not use this and use custom more direct implementation of an UUPS proxy

packages/deployer/test/sample-project/contracts/modules/UpgradeModule.sol
```diff
@@ -0,0 +1,106 @@
...
+pragma solidity ^0.8.0;
+
+import "../mixins/OwnerMixin.sol";
+import "../storage/ProxyStorage.sol";
+import "../libraries/Address.sol";
+
+contract UpgradeModule is ProxyStorageNamespace, OwnerMixin {
+    /* MUTATIVE FUNCTIONS */
+
+    function upgradeTo(address newImplementation) public onlyOwnerOrProxy {
```
ajsantander, line 11: For sample project, just allow bricking with a small comment with a warning.

In the future, we may want to use this anti-bricking mechanism and/or investigate OpenZeppelin's mechanism in their official UUPS proxy implementation.

.solhintignore
```diff
@@ -1 +1,2 @@
 node_modules/
+artifacts/
```
leomassazza, line 2: Excluding artifacts. At compile time, solc can create folders ending in **.sol** that will match (and make fail) solhint contracts expansion filenames

package.json
```diff
@@ -17,8 +17,8 @@
   "scripts": {
     "lint:js": "prettier --check '**/*.js' && eslint '**/*.js'",
     "lint:js:fix": "prettier --write '**/*.js' && eslint --fix '**/*.js'",
-    "lint:sol": "prettier --check 'packages/**/*.sol' && solhint 'packages/**/*.sol'",
-    "lint:sol:fix": "prettier --write 'packages/**/*.sol' && solhint --fix 'packages/**/*.sol'",
+    "lint:sol": "prettier --check 'packages/**/contracts/*.sol' && solhint 'packages/**/*.sol'",
```
leomassazza, line 20: Not needed. Added the exclusion in .solhintignore

packages/deployer/test/sample-project/contracts/MainProxy.sol
```diff
@@ -0,0 +1,37 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "./storage/ProxyNamespace.sol";
+
+contract MainProxy is ProxyNamespace {
+    function _fallback(address implementation) internal virtual {
```
leomassazza, line 7: MainProxy defines the proxy on it, not extending any Proxy

packages/deployer/test/sample-project/contracts/MainProxy.sol
```diff
@@ -0,0 +1,37 @@
...
+                return(0, returndatasize())
+            }
+        }
+    }
+
+    fallback() external payable virtual {
+        _fallback(_getImplementation());
+    }
+
+    receive() external payable virtual {
```
leomassazza, line 30: See https://docs.soliditylang.org/en/v0.8.0/contracts.html#receive-ether-function

packages/deployer/test/sample-project/contracts/modules/AnotherModule.sol
```diff
@@ -0,0 +1,22 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../storage/GlobalNamespace.sol";
+import "../mixins/ModulesMixin.sol";
+import "./SomeModule.sol";
+
+contract AnotherModule is ModulesMixin {
```
leomassazza, line 8: AnotherModule will communicate with SomeModule and set the values from here.

packages/deployer/test/sample-project/contracts/modules/SettingsModule.sol
```diff
@@ -0,0 +1,19 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../storage/SettingsNamespace.sol";
+import "../mixins/OwnerMixin.sol";
+
+contract SettingsModule is SettingsNamespace, OwnerMixin {
```
leomassazza, line 7: Settings Module kept to show an independent Namespace that will require ownership to update

packages/deployer/test/sample-project/contracts/storage/ProxyNamespace.sol
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+
+    function _setImplementation(address newImplementation) internal {
+        assembly {
+            sstore(_IMPLEMENTATION_SLOT, newImplementation)
+        }
+    }
+
+    struct ProxyStorage {
```
leomassazza, line 20: Do we need that? I think not at the moment.

packages/deployer/test/sample-project/contracts/MainProxy.sol
ajsantander: A bit new on the "virtual" keyword here too. Question: do we need it here, since I doubt we will be inheriting MainProxy anywhere.

packages/deployer/test/sample-project/contracts/MainProxy.sol
ajsantander: From those docs:
"If no such function exists, but a payable fallback function exists, the fallback function will be called on a plain Ether transfer."

If so, then shouldn't having a payable fallback function be enough? I.e. no need for receive and for making _fallback an external function.

packages/deployer/test/sample-project/contracts/MainProxy.sol
```diff
@@ -0,0 +1,37 @@
...
+                revert(0, returndatasize())
+            }
+            default {
+                return(0, returndatasize())
+            }
+        }
+    }
+
+    fallback() external payable virtual {
+        _fallback(_getImplementation());
```
ajsantander, line 27: See comment in ProxyNamespace.sol, I think this getter should be `_proxyStorage().implementation`

packages/deployer/test/sample-project/contracts/mixins/ModulesMixin.sol
```diff
@@ -0,0 +1,12 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../storage/ProxyNamespace.sol";
+
+contract ModulesMixin is ProxyNamespace {
```
ajsantander, line 6: This mixin is not needed. Please delete.

packages/deployer/test/sample-project/contracts/modules/UpgradeModule.sol
```diff
@@ -0,0 +1,41 @@
...
+        require(_isContract(newImplementation), "Invalid: not a contract");
+        // WARNING: This contract is brickable.
+        // In production you should check the newImplementation is upgradeable too
+
+        _setImplementation(newImplementation);
+    }
+
+    /* VIEW FUNCTIONS */
+
+    function isUpgradeable() public pure returns (bool) {
```
ajsantander, line 34: This function can be removed.

ajsantander (approved):
Awesome!

PR #18, opened 2021-08-31
Add deployer prepare files subtask

mjlescano:
This PR adds all the necessary task for preparing all the necessary files on the `deployments` folder.

## Usage
```bash
npx hardhat deploy --debug --alias sirius --instance official
```

## Deploy Steps Implemented
1. Try to create a deployments file for the given instance, e.g.: `./deployments/mainnet/official/2021-08-31-00-sirius.json`
2. If a deployment file exists but with `properties.completed = false`, it will ask the user to continue the deployment using that file
3. Opens/Create the deployment file and all the changes generated during the deploy are saved on the `.json` file
4. Creates an autosave data object on `hre.deployer.data` ready to be used by other subtasks


packages/deployer/index.js
```diff
@@ -1,7 +1,40 @@
 const fs = require('fs');
 const path = require('path');
+const { extendConfig, extendEnvironment } = require('hardhat/config');
 
-['tasks', 'subtasks', 'extensions'].forEach((folder) =>
+extendConfig((config, userConfig) => {
```
leomassazza, line 5: We can have extendConfig and extendEnvironment as independent files in extensions to have this index super slim and without any specific logic

packages/deployer/tasks/deploy.js
```diff
@@ -1,13 +1,37 @@
...
-} = require('../task-names');
+const { TASK_DEPLOY, SUBTASK_PREPARE_DEPLOYMENT } = require('../task-names');
+const logger = require('../util/logger');
+const prompter = require('../util/prompter');
 
 task(TASK_DEPLOY, 'Deploys all system modules')
-	.addOptionalParam('instance', 'The name of the target instance for deployment', 'official')
+  .addFlag('noConfirm', 'Skip all confirmation prompts', false)
+  .addFlag('debug', 'Display debug logs', false)
+  .addFlag('clear', 'Clear all previous deployment data for the selected network', false)
```
leomassazza, line 11: clear and alias are not used (maybe used later since it still a draft)

leomassazza: Just saw it on the subtask

packages/deployer/index.js
mjlescano: Yeah, I removed the extensions because I prefered to have all the global env defined on a single place (like [this](https://github.com/nomiclabs/hardhat-ts-plugin-boilerplate/blob/master/src/index.ts)), but as its probably going to get larger I will put it back as before.

packages/deployer/extensions/config.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  config.deployer = {
+    proxyName: '',
+    paths: {
+      modules: 'contracts/modules',
+      deployments: 'deployments',
+      ...(userConfig.deployer.paths || {}),
+    },
+    ...userConfig.deployer,
```
ajsantander, line 21: This should be before paths

packages/deployer/extensions/environment.js
```diff
@@ -0,0 +1,11 @@
+const { extendEnvironment } = require('hardhat/config');
+
+extendEnvironment((hre) => {
+  if (hre.deployer) {
+    throw new Error('Deployer plugin already loaded.');
+  }
+
+  hre.deployer = {
+    paths: {},
```
ajsantander, line 9: Should add data object?

ajsantander:
```suggestion
    paths: {},
    data: null,
```

leomassazza (approved):
LGTM

PR #31, opened 2021-09-02
Add deploy modules subtask

mjlescano:
This PR adds the subtasks necessary for deploying all the contracts on the `contracts/modules` folder.

Before deploying, searches for all the modules files at `contracts/modules/**/*.sol` and checks the following strategy:
1. If the current network is `hardhat`, it will deploy all the modules again.
2. Checks on the blockchain if the same module was deployed on another Deployment, and if weren't any changes on the bytecode, it skips the deployment of that module.
3. If the bytecode is different, it notifies the user that is going to update that specific module.
4. If the module weren't previously deployed, notifies about its creation (deployment for the first time).

### DIFF
--- a/packages/deployer/subtasks/deploy-contracts.js
+++ b/packages/deployer/subtasks/deploy-contracts.js
@@ -100,7 +100,7 @@ async function _processContracts(contracts) {
 }
 
 /**
- * Deploy the give contract using ethers.js
+ * Deploy the given contract using ethers.js
  */
 async function _createAndDeployContract(contractName) {
   logger.success(`Deploying ${contractName}`);
--- a/packages/deployer/subtasks/deploy-contracts.js
+++ b/packages/deployer/subtasks/deploy-contracts.js
@@ -10,7 +10,7 @@ const { SUBTASK_DEPLOY_CONTRACTS } = require('../task-names');
 
 subtask(
   SUBTASK_DEPLOY_CONTRACTS,
-  'Deploys a list of contracts, avoiding contracts that do not need to be compiled, and prompting the user for confirmation.'
+  'Deploys a list of contracts, avoiding contracts that do not need to be deployed, and prompting the user for confirmation.'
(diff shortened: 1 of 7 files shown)


packages/deployer/subtasks/deploy-contracts.js
```diff
@@ -0,0 +1,130 @@
...
+  getContractNameFromPath,
+  getContractBytecodeHash,
+  getAddressBytecodeHash,
+} = require('../utils/contracts');
+const { subtask } = require('hardhat/config');
+const { SUBTASK_DEPLOY_CONTRACTS } = require('../task-names');
+
+subtask(
+  SUBTASK_DEPLOY_CONTRACTS,
+  'Deploys a list of contracts, avoiding contracts that do not need to be compiled, and prompting the user for confirmation.'
```
ajsantander, line 13: "be deployed"?

packages/deployer/subtasks/deploy-contracts.js
```diff
@@ -0,0 +1,130 @@
...
+    throw new Error(`Error deploying ${contractName}`);
+  }
+
+  return { contract, transaction: contract.deployTransaction };
+}
+
+/**
+ * Given a transaction, wait for it to finish and return the state with the gas used.
+ */
+async function _waitForTransaction(transaction) {
```
ajsantander, line 121: We might want to move this out to a util later, when some other sub-task needs it. Eventually, it might even move out to a core-js util outside of the package. Or maybe not. We'll see. It's good here for now.

ajsantander (approved):
Looks good! Only nits

packages/deployer/subtasks/deploy-contracts.js
mjlescano: For sure! Would be handy to have a single function (or prototype) that does `_deployContract` + `_waitForTransaction`

PR #32, opened 2021-09-03
Add router generation & deployment subtasks

mjlescano:
This PR generates the updated Router using the modules defined on the deployments file and deploys it if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,3 @@
 node_modules
 cache
 artifacts
-deployments/hardhat*
-deployments/local*
-contracts/Router_local.sol
--- a/packages/deployer/subtasks/deploy-router.js
+++ b/packages/deployer/subtasks/deploy-router.js
@@ -1,0 +1,15 @@
+const logger = require('../utils/logger');
+const { subtask } = require('hardhat/config');
+const { TASK_COMPILE } = require('hardhat/builtin-tasks/task-names');
+
+const { SUBTASK_DEPLOY_CONTRACTS, SUBTASK_DEPLOY_ROUTER } = require('../task-names');
+
+subtask(SUBTASK_DEPLOY_ROUTER).setAction(async (_, hre) => {
+  logger.subtitle('Deploying router');
+
+  await hre.run(TASK_COMPILE, { force: false, quiet: true });
+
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -4,6 +4,3 @@
 node_modules
 cache
 artifacts
-deployments/hardhat*
-deployments/local*
-contracts/Router_local.sol
```
mjlescano, line 9: This is already defined on `packages/deployer/.gitignore`

packages/deployer/test/sample-project/package.json
```diff
@@ -2,6 +2,10 @@
   "name": "sample-project",
   "version": "1.0.0",
   "description": "",
+  "repository": {
```
leomassazza, line 5: Why Synthetix repo here? and not Synthetix-v3

packages/deployer/subtasks/deploy-router.js
```diff
@@ -0,0 +1,15 @@
+const logger = require('../utils/logger');
+const { subtask } = require('hardhat/config');
+const { TASK_COMPILE } = require('hardhat/builtin-tasks/task-names');
+
+const { SUBTASK_DEPLOY_CONTRACTS, SUBTASK_DEPLOY_ROUTER } = require('../task-names');
+
+subtask(SUBTASK_DEPLOY_ROUTER).setAction(async (_, hre) => {
+  logger.subtitle('Deploying router');
+
+  await hre.run(TASK_COMPILE, { force: false, quiet: true });
```
leomassazza, line 10: Shouldn't we compile ONLY the router here? (hh will check all the sources again to look for changes)

mjlescano: That would be nice, but looks like the original task doesn't allow us to define which contracts to compile: https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/builtin-tasks/compile.ts#L1257-L1259

And FYI, I just did a `console.time` on this line and its already taking `700ms`, which I think is a lot keeping in mind that we only have 6 modules here.

leomassazza (approved):
LGTM

ajsantander (approved):

PR #43, opened 2021-09-06
Add instance to the generated Router name

mjlescano:
This PR adds the `--instance` parameter to the Router.

And, as a plus, changes the naming convention of the generated router to use `CapitalCase` & replaces the templating to use Mustache.js.


packages/deployer/tasks/deploy.js
```diff
@@ -15,6 +15,7 @@ const {
 const logger = require('../utils/logger');
 const prompter = require('../utils/prompter');
 const relativePath = require('../utils/relative-path');
+const { capitalize } = require('../utils/string');
 
 const isWord = (str) => /^[\w\d]+$/.test(str);
```
leomassazza, line 20: np (or future improvement). Could we move those definitions (and other regex) to the string utils?

packages/deployer/utils/render-template.js
```diff
@@ -0,0 +1,7 @@
+const fs = require('fs');
+const Mustache = require('mustache');
+
+module.exports = function renderTemplate(filepath, data = {}) {
+  const template = fs.readFileSync(filepath).toString();
+  return Mustache.render(template, data);
```
leomassazza, line 6: love it! super simple 👍

packages/deployer/utils/string.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  /**
+   * Capitalize the first letter of the given string
+   * @param {string} str
+   * @returns {string}
+   */
+  capitalize(str) {
+    return str.charAt(0).toUpperCase() + str.substring(1);
```
leomassazza, line 8: return undefined or '' for undefined input

leomassazza (approved):
Approved with some comments

PR #50, opened 2021-09-06
Bootstraps sample-project tests

### LINKED ISSUE #26: Bootstrap sample-project tests

ajsantander:
Fix #26 

The gist of this PR is the addition of the new file:
`deployer/test/sample-project/test/SettingsModule.test.js`

This tests works on an existing deployment on local, and is run with
`npx hardhat test`

The tricky part was getting the address of the main proxy of the latest deployment. A lot of logic could have been added to the test for that, but all tests would need that logic. So, perhaps it could be a helper? But then projects using the plugin would need these helpers, and the logic probably already exists in the deployer itself. So, I tried to start exposing these "common operations" from the deployer:
https://github.com/Synthetixio/synthetix-v3/pull/50/files#diff-d27aebaf2b1cb5b872cfa531db45487b4f7376085356e4886d1024107b55d5deR16

Indeed, it does make tests really simple, but I'm left with the feeling that coding the plugin with hre data injection makes these exposed tools weird. Notice how I moved some stuff out of `prepare-deployments` because it involved the logic needed by the test.

My question is: if there was no hre runtime injection, wouldn't all the tools be immediately usable by outsiders? Consider that outsiders don't have this runtime environment that gets populated during deployment.

### DIFF
--- a/packages/deployer/subtasks/prepare-deployment.js
+++ b/packages/deployer/subtasks/prepare-deployment.js
@@ -1,19 +1,15 @@
 const fs = require('fs');
 const path = require('path');
-const glob = require('glob');
 const mkdirp = require('mkdirp');
 const rimraf = require('rimraf');
-const naturalCompare = require('string-natural-compare');
 const { subtask } = require('hardhat/config');
 
 const logger = require('../utils/logger');
 const prompter = require('../utils/prompter');
 const relativePath = require('../utils/relative-path');
 const autosaveObject = require('../utils/autosave-object');
+const { getDeploymentFiles } = require('../utils/deployments');
 const { SUBTASK_PREPARE_DEPLOYMENT } = require('../task-names');
-
-// Regex for deployment file formats, e.g.: 2021-08-31-00-sirius.json
-const DEPLOYMENT_FILE_FORMAT = /^[0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{2,}(?:-[a-z0-9]+)?\.json$/;
 
 const DEPLOYMENT_SCHEMA = {
   properties: {
@@ -38,10 +34,7 @@ subtask(
 
   mkdirp.sync(hre.deployer.paths.instance);
 
-  const { previousFile, currentFile } = await _determineDeploymentFiles(
-    hre.deployer.paths.instance,
-    alias
-  );
+  const { previousFile, currentFile } = await _determineDeploymentFiles(alias);
 
   hre.deployer.file = currentFile;
   hre.deployer.data = autosaveObject(hre.deployer.file, DEPLOYMENT_SCHEMA);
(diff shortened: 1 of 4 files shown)


packages/deployer/utils/deployments.js
```diff
@@ -0,0 +1,69 @@
...
+  paths.routerTemplate = path.resolve(__dirname, '../templates/GenRouter.sol.mustache');
+  paths.routerPath = relativePath(
+    path.join(hre.config.paths.sources, `${hre.deployer.routerModule}.sol`)
+  );
+  paths.proxyPath = relativePath(
+    path.join(hre.config.paths.sources, `${hre.config.deployer.proxyName}.sol`)
+  );
+
+  return paths;
+}
```
mjlescano, line 61: As a response to the PR's description question, Yes, it would be great if we do these utils to not use the `hre` global variable on context. We could plan a refactor to change all the interfaces of these function to take everything they need as parameters, so they will be easily reusable from the tests.

E.g., this is an ugly example, which could be improved, but is 100% reusable without the global context:

```javascript
function getDeploymentPaths({
  instance,
  routerModule,
  proxyName,
  network,
  root,
  sources,
  deployments,
  modules,
}) {
  const paths = {};

  paths.deployments = path.resolve(root, deployments);
  paths.modules = path.resolve(root, modules);
  paths.network = path.join(paths.deployments, network);
  paths.instance = path.join(paths.network, instance);
  paths.extended = path.join(paths.instance, 'extended');
  paths.routerTemplate = path.resolve(__dirname, '../templates/GenRouter.sol.mustache');
  paths.routerPath = relativePath(
    path.join(sources, `${routerModule}.sol`)
  );
  paths.proxyPath = relativePath(
    path.join(sources, `${proxyName}.sol`)
  );

  return paths;
}
```

Then, call it once at the beginning of the task:
```javascript
hre.deployer.paths = getDeploymentPaths({
  instance: 'official',
  routerModule: 'GenRouterLocalOfficial',
  proxyName: hre.config.deployer.proxyName,
  network: hre.network.name,
  root: hre.config.paths.root,
  sources: hre.config.paths.sources,
  deployments: hre.config.deployer.paths.deployments,
  modules: hre.config.deployer.paths.modules,
})

```

packages/deployer/subtasks/prepare-deployment.js
```diff
@@ -38,10 +34,7 @@ subtask(
 
   mkdirp.sync(hre.deployer.paths.instance);
 
-  const { previousFile, currentFile } = await _determineDeploymentFiles(
-    hre.deployer.paths.instance,
-    alias
-  );
+  const { previousFile, currentFile } = await _determineDeploymentFiles(alias);
```
mjlescano, line 37: The `_determineDeploymentFiles` function takes 2 params instead of 1: `(instance, alias)`

packages/deployer/utils/deployments.js
mjlescano: This would make the code a little harder to track all the variables we send around, but more explicit, it would not require the reader to know about global context, and easier to reuse on tests, etc

packages/deployer/subtasks/prepare-deployment.js
ajsantander: Fixed

mjlescano (approved):

leomassazza (approved):
LGTM

PR #52, opened 2021-09-07
Generate-Intermodule-Communication-Mixing-Module

### LINKED ISSUE #42: Dynamically generate a Mixin that exposes hardcoded module addresses via internal calls
Multiple experiments were carried out to see if we can have modules interact with each other with a lower gas cost:
* https://github.com/Synthetixio/synthetix-v3-labs/tree/master/proxy-architecture-poc-5#communication-between-modules
* https://github.com/Synthetixio/synthetix-v3-labs/tree/master/intermodule-communication-poc1

The conclusion of these experiments was to try to not only generate the Solidity code of the router, but also generate a mixin that any module can inherit, exposing the addresses of all the modules via constants.

This poses many challenges:

1) The mixin needs to be generated before the modules are deployed, and since the module addresses are not known before they are deployed, it becomes a chicken vs egg problem. However, we can predict what module addresses will be, since they are deterministically calculated from the address of the deployer EOA (or contract?) and its nonce.

2) Since a project may target multiple networks/instances, it may have multiple routers such as RouterLocal.sol, RouterKovan.sol, etc. This duplication should also apply to this mixin. The duplication in the router code is ugly but tolerable, but the duplication in the mixin would be too weird. I.e.

```
contract AModule is RegistryMixinLocal { ... }
or
contract AModule is RegistryMixinKovan { ... }
```

A solution to this issue could be to have a single mixin:
```contract AModule is RegistryMixin```
and allow it to change depending on the last deployment. This means that this Solidity file will be constantly changing, depending on what the last targeted network was. Which is weird too, but tolerable.

In fact, if this works, the same principle could be applied to the router, effectively removing the need to keep multiple routers around in the codebase. Since the deployer always has all the necessary data to generate a new router, then why keep duplicates around, or even keep any around in the source code?

-------
### Basic (excessive gas usage on deployment)

- Generate a fake AddressMixin with random addresses
- Compile all modules to find out the ones that changed due to explicit code change and by using the mixin
- Identify all contracts to deploy (and the order that will be used)
- Calculate and collect all modules new addresses 
- Fill the missing modules addresses (the ones that weren’t re-deployed) using latest deployment info on the instance
- Generate a new AddressMixin with the right addresses
- Deploy all system (**it will always deploy ALL MODULES that depends on AddressMixin**)
- Validate that the AddressMixin has the right addresses  

This process will always try to generate a new AddressMixin even if no modules were updated on the first time

### Improved Process (only deploy when a module was updated)

- First identify if any module was updated
- If any module was updated, continue with the Basic process. If not, the current AddressMixin is still valid

**it will deploy ALL MODULES that depends on AddressMixin when any module was updated**

### Improved Process 2 (tag Addressable modules)

This improvement also adds a tag to the modules that are addressable (the ones that are potentially usable from another module, and hence has the address exported in AddressMixin) 
i.e. `// address-mixin isAddressable`
- Set a tag on the Modules source (this is done only once while coding and the tag can be added later if there's a need to use a module that wasn't set originally as addressable)
- Identify Modules that are addressable and check if any was updated
- If any addressable module was updated, continue with the Basic process. If not, the current AddressMixin is still valid

**it will deploy ALL MODULES that depends on AddressMixin when any addressable module was updated**

### PR DESCRIPTION
**Pull request by leomassazza**

### CODE REVIEW

File: packages/deployer/subtasks/sync-sources.js [hunk 1]
```diff
@@ -53,6 +53,7 @@ async function _removeDeletedSources({ sources, previousData }) {
 async function _addNewSources({ sources, data, previousData }) {
   const toAdd = sources.filter((source) => {
     const previousModule = previousData?.contracts.modules[source];
+    // if data.contracts.modules[source] is undefined at execution time will use the previousModule or the default one.
```
leomassazza, line 56: Add comment to simplify line readability

packages/deployer/task-names.js
```diff
@@ -3,11 +3,15 @@ module.exports = {
   SUBTASK_DEPLOY_CONTRACTS: 'deploy-contracts',
   SUBTASK_DEPLOY_MODULES: 'generate-deploy-modules',
   SUBTASK_DEPLOY_ROUTER: 'generate-deploy-router',
+  SUBTASK_FINALIZE_DEPLOYMENT: 'finalize-deployment',
```
leomassazza, line 6: missing subtask name

packages/deployer/task-names.js
```diff
@@ -3,11 +3,15 @@ module.exports = {
...
+  SUBTASK_GENERATE_IMC_SOURCE: 'generate-imc-source',
   SUBTASK_GENERATE_ROUTER_SOURCE: 'generate-router-source',
   SUBTASK_PREPARE_DEPLOYMENT: 'prepare-deployment',
   SUBTASK_PRINT_INFO: 'print-info',
   SUBTASK_SYNC_SOURCES: 'sync-sources',
   SUBTASK_UPGRADE_PROXY: 'upgrade-proxy',
+  SUBTASK_VALIDATE_IMC: 'validate-imc',
   SUBTASK_VALIDATE_ROUTER: 'validate-router',
   TASK_DEPLOY: 'deploy',
+  TASK_DUMMY_IMC: 'create-dummy-imc',
```
leomassazza, line 16: naming suggestions are welcome here :)

packages/deployer/tasks/create-dummy-imc.js
```diff
@@ -0,0 +1,28 @@
+const { task } = require('hardhat/config');
```
leomassazza, line 1: This task creates a fake GenIMCMixin.sol contract with random addresses for all the modules available to allow the developer to compile solidity source without errors

packages/deployer/tasks/deploy.js
```diff
@@ -37,14 +39,17 @@ task(TASK_DEPLOY, 'Deploys all system modules')
     prompter.noConfirm = noConfirm;
 
     hre.deployer.routerModule = ['GenRouter', hre.network.name, instance].map(capitalize).join('');
+    hre.deployer.imcMixinModule = 'GenIMCMixin';
 
     hre.deployer.paths = getDeploymentPaths(instance);
 
     await hre.run(SUBTASK_PREPARE_DEPLOYMENT, taskArguments);
     await hre.run(SUBTASK_PRINT_INFO, taskArguments);
-    await hre.run(TASK_COMPILE, { force: true, quiet: true });
```
leomassazza, line 45: reorder to compile after the IMC is created

packages/deployer/test/sample-project/contracts/modules/AModule.sol
```diff
@@ -0,0 +1,14 @@
+//SPDX-License-Identifier: Unlicense
```
leomassazza, line 1: Sample contract using the new IMC

packages/deployer/utils/contracts.js
```diff
@@ -50,10 +50,36 @@ async function alreadyDeployed(contractPath, contractData) {
   return sourceBytecodeHash === remoteBytecodeHash;
 }
 
+/**
+ * Sort contracts by the ones that doesn't need deployment, the ones that are going
+ * to be re-deployed with updated code, and the ones that are going to be deployed
+ * for the first time.
+ */
+async function processContracts(contracts) {
```
leomassazza, line 58: Moved from deploy-contracts subtask to be used also by imc creation subtask

packages/deployer/utils/deployments.js
```diff
@@ -56,6 +56,9 @@ function getDeploymentPaths(instance = 'official') {
   paths.proxyPath = relativePath(
     path.join(hre.config.paths.sources, `${hre.config.deployer.proxyName}.sol`)
   );
+  paths.mixins = path.resolve(hre.config.paths.root, hre.config.deployer.paths.mixins);
```
leomassazza, line 59: mixin and IMC paths

packages/deployer/utils/logger.js
```diff
@@ -59,6 +59,10 @@ module.exports = {
     this.log(chalk.green(`✅ ${msg}`));
   },
 
+  fail(msg) {
```
leomassazza, line 62: Added a failure message to the logger

packages/deployer/utils/provider.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+async function evaluateNextDeployedContractAddress(nonce, _signer) {
+  const from = _signer
+    ? await _signer.getAddress()
+    : await hre.ethers.provider.getSigner().getAddress();
+  return hre.ethers.utils.getContractAddress({ from, nonce });
+}
+
+module.exports = {
```
leomassazza, line 18: utils to get the new contract address

packages/deployer/subtasks/generate-imc.js
```diff
@@ -0,0 +1,83 @@
+const fs = require('fs');
```
ajsantander, line 1: Can we rename this file to something more intuitive? First time you see "imc" its hard to tell what it is.
I'd rather something verbose, but self explanatory.

packages/deployer/subtasks/generate-imc.js
```diff
@@ -0,0 +1,83 @@
...
+const { readPackageJson } = require('../utils/package');
+const { getContractNameFromPath, processContracts } = require('../utils/contracts');
+const { getNonce, nextNonce, evaluateNextDeployedContractAddress } = require('../utils/provider');
+const renderTemplate = require('../utils/render-template');
+
+const TAB = '    ';
+
+subtask(
+  SUBTASK_GENERATE_IMC_SOURCE,
+  'Pre-calculates the modules addresses and generates the source for a new IMCMixin contract.'
```
ajsantander, line 15: Naming could apply here as well.
Perhaps a `ModulesMixin` or a `RegistryMixin`?

packages/deployer/test/sample-project/contracts/modules/AModule.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity ^0.8.0;
+
+import "../mixins/GenIMCMixin.sol";
+import "./SomeModule.sol";
+
+contract AModule is GenIMCMixin {
```
ajsantander, line 7: Shouldn't there be a GenIMCMixin per network? If so, how does AModule know which one to extend?

leomassazza:
Closing it to start from scratch and bring the changes. There are more conflicts than proposed changes.

PR #54, opened 2021-09-07
Add custom alphanumeric type for hardhat argument

mjlescano:
This PRs adds a custom hardhat argument type for sake of celemines on the main deploy task. This validation is necessary to make sure that we generate correct deployment files formats.


leomassazza (approved):
LGTM

PR #69, opened 2021-09-08
Stop using hre.deployer.paths

### LINKED ISSUE #68: Stop using hre.deployer.paths
Even though its practical to calculate paths initially, and use the results in all subtasks, it goes against the principle of making subtasks as self-sufficient as possible.

ajsantander:
Fix #68

### DIFF
--- a/packages/deployer/subtasks/deploy-router.js
+++ b/packages/deployer/subtasks/deploy-router.js
@@ -1,15 +1,22 @@
+const path = require('path');
 const logger = require('../utils/logger');
 const { subtask } = require('hardhat/config');
 const { TASK_COMPILE } = require('hardhat/builtin-tasks/task-names');
+const { getRouterName } = require('../utils/router');
+const relativePath = require('../utils/relative-path');
 
 const { SUBTASK_DEPLOY_CONTRACT, SUBTASK_DEPLOY_ROUTER } = require('../task-names');
 
-subtask(SUBTASK_DEPLOY_ROUTER).setAction(async (_, hre) => {
+subtask(SUBTASK_DEPLOY_ROUTER).setAction(async ({ instance }, hre) => {
   logger.subtitle('Deploying router');
 
   await hre.run(TASK_COMPILE, { force: false, quiet: true });
 
-  const contractPath = hre.deployer.paths.router;
+  const contractPath = path.join(
+    relativePath(hre.config.paths.sources),
+    getRouterName({ network: hre.network.name, instance })
+  );
+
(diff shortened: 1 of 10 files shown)


packages/deployer/subtasks/prepare-deployment.js
```diff
@@ -28,7 +28,7 @@ subtask(
 ).setAction(async (taskArguments, hre) => {
   const { clear, alias } = taskArguments;
 
-  const deploymentsFolder = hre.deployer.paths.deployments;
+  const deploymentsFolder = hre.config.deployer.paths.deployments;
```
mjlescano, line 29: this is going to break because this variable doesn't include the network and instance subpaths

packages/deployer/subtasks/deploy-router.js
```diff
@@ -1,15 +1,22 @@
...
 
-subtask(SUBTASK_DEPLOY_ROUTER).setAction(async (_, hre) => {
+subtask(SUBTASK_DEPLOY_ROUTER).setAction(async ({ instance }, hre) => {
   logger.subtitle('Deploying router');
 
   await hre.run(TASK_COMPILE, { force: false, quiet: true });
 
-  const contractPath = hre.deployer.paths.router;
+  const contractPath = path.join(
+    relativePath(hre.config.paths.sources),
```
mjlescano, line 16: the relativePath should receive hre.config.paths.root as second parameter, because is defaulted to process.cwd but that's not always the case on hardhat projects (because the hardhats project root can be configured to any other path)

packages/deployer/subtasks/deploy-router.js
```diff
@@ -1,15 +1,22 @@
...
+    getRouterName({ network: hre.network.name, instance })
```
mjlescano, line 17: maybe you could create an util named getRouterPath, and recieves the instance, network & hre.config.paths.sources as params. And this util could be used in other places that need the router path

packages/deployer/subtasks/prepare-deployment.js
ajsantander: Fixed

packages/deployer/subtasks/deploy-router.js
ajsantander: Done!

packages/deployer/subtasks/deploy-router.js
ajsantander: I thought about this, but concluded that it wouldn't be worth the trouble. Let's leave it for now.

leomassazza (approved):
👍

PR #70, opened 2021-09-08
move some utils to core-js

### LINKED ISSUE #72: Move non-deployer specific utilities out to the core-js package

leomassazza:
Fix #72

Small (in content, but not in number of files) PR to move some utils to core-js.

Moved utils are:
- logger
- prompter
- git
- package
- relative-path
- string


packages/core-js/package.json
```diff
@@ -9,5 +9,10 @@
   "keywords": [],
   "author": "",
   "license": "MIT",
-  "devDependencies": {}
+  "devDependencies": {},
+  "dependencies": {
+    "chalk": "4.1.2",
+    "inquirer": "8.1.2",
+    "mustache": "4.2.0"
```
ajsantander, line 16: Isn't rendering a template a very deployer plugin specific task? Where do you see anyone using templating from our utils for something else?

ajsantander (changes requested):
Looks good! My only request would be to consider not moving template rendering utils to core-js. See comment.

packages/core-js/package.json
leomassazza: yep, it's super deployer specific... I'll pull it back

ajsantander (approved):

PR #90, opened 2021-09-09
Split utils

### LINKED ISSUE #86: Split deployer/utils into deployer/internal and deployer/utils - only utils are intended for external use and require simple primitives as input

ajsantander:
Fix #86


packages/core-js/utils/contracts.js
```diff
@@ -0,0 +1,43 @@
...
+
+  return sourceBytecodeHash === remoteBytecodeHash;
+}
+
+function getBytecodeHash(bytecode) {
+  return ethers.utils.sha256(bytecode);
+}
+
+async function getRemoteBytecode(address) {
+  return await ethers.provider.getCode(address);
```
ajsantander, line 17: Would be better not to use ethers default provider

packages/deployer/subtasks/prepare-deployment.js
```diff
@@ -52,7 +52,7 @@ subtask(
  * @returns {{ currentFile: string, previousFile: string }}
  */
 async function _determineDeploymentFiles(deploymentsFolder, alias) {
-  const deployments = getDeploymentFiles(deploymentsFolder);
+  const deployments = getAllDeploymentsFilesForInstance(deploymentsFolder);
   const previousFile = deployments.length > 0 ? deployments[deployments.length - 1] : null;
```
ajsantander, line 56: Rename to latest?

packages/core-js/utils/contracts.js
ajsantander: Fixed

packages/deployer/subtasks/prepare-deployment.js
ajsantander: Fixed

leomassazza (approved):
👍

PR #96, opened 2021-09-09
Fix modules constants names on router

### LINKED ISSUE #94: Consider using camelcase in module constants in the router

mjlescano:
Closes #94

This also fixes the __*very unlikely*__ case were the modules `BearableModule` & `BearAbleModule` would both generate the `_BEARABLEMODULE` constant and collide, instead of `_BEARABLE_MODULE` & `_BEAR_ABLE_MODULE`

### DIFF
--- a/packages/deployer/subtasks/generate-router.js
+++ b/packages/deployer/subtasks/generate-router.js
@@ -85,7 +85,7 @@ function _renderSelectors({ binaryData }) {
       for (const s of node.selectors) {
         selectorsStr += `\n${TAB.repeat(4 + indent)}case ${
           s.selector
-        } { result := _${s.contractName.toUpperCase()} } // ${s.contractName}.${s.name}()`;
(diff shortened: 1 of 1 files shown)


packages/deployer/subtasks/generate-router.js
```diff
@@ -99,22 +99,33 @@ function _renderSelectors({ binaryData }) {
...
 }
 
+/**
+ * Converts the contracts name to private _CONSTANT_CASE format.
+ * E.g.:
+ *   'BearableModule' => '_BEARABLE_MODULE'
+ *   'Proxy' => '_PROXY'
+ *   'ERC20Token' => '_ERC20_TOKEN'
+ */
+function _contractConstantName(name) {
```
ajsantander, line 125: Maybe a more intuitive name like _toConstantCase(…)?

ajsantander (approved):
Minor nit

PR #106, opened 2021-09-15
Checking for msg.sender in intermodular comms

### LINKED ISSUE #107: Helper for events in tests

ajsantander:
This PR shows that there is a problem with type 1 inter-modular communications.
(It also adds a handy util for events in core-js so Fix #107 )

When a module casts itself as another module, and performs a call through the proxy and then the router as so:

```
// in AModule.sol
BModule(address(this)).setSomeValue(42);
```

`msg.sender` is no longer the caller, but the proxy itself. The correct proxy storage is still used, but the `call` changes `msg.sender` from the user to the proxy.

This can be demonstrated with two new tests in `AnotherModule.test.js`. Whenever a value is changed in `SomeModule.sol`, it emits an event with `msg.sender` information. It shows that when called through `AnotherModule.sol` via casting, `msg.sender` is the proxy, which is not acceptable. When using router, `msg.sender` is the user, so no problem in that case.

This means that `delegatecall` will always be needed for inter-modular comms. Type 1 communications is no longer an option.


packages/core-js/utils/events.js
```diff
@@ -0,0 +1,45 @@
...
+ * will parse them if a suggested contract interface is provided.
+ * @param {Object} receipt An ethers transaction receipt
+ * @param {string} eventName The name of the event to find, e.g. "Transfer"
+ * @param {contract} contract The contract to use for identifying unparsed logs
+ * @returns {array} The array of parsed events
+ */
+function findEvent({ receipt, eventName, contract = undefined }) {
+  const eventsAreParsed = !receipt.events.some((e) => e.event === undefined);
+
+  let events = receipt.events;
```
mjlescano, line 12: Just a suggestion: 
```suggestion
  let { events } = receipt;
```

mjlescano (approved):

leomassazza (approved):
LGTM

PR #112, opened 2021-09-15
Remove comms type 1

### LINKED ISSUE #111: Fix intermodular comms simple way
Atm tests are failing because msg.sender is lost when using the casting method (comms type 1). This method should be removed for the time being.

ajsantander:
Fix #111


packages/deployer/test/sample-project/test/helpers/initializer.js
```diff
@@ -22,19 +22,7 @@ function bootstrap() {
 
 async function deploySystemIfNeeded() {
   if (!proxyAddress) {
-    const deployment = getDeployment();
```
ajsantander, line 25: Had to remove this:

The idea of this was to avoid local deployments every time tests are ran. Worked nicely when just changing test files, but it doesn't detect when contract changes and re-deploy the system.

Without this, every time tests are run, the system will be re-deployed. It's fast enough for now anyway.

mjlescano (approved):

i-stam (approved):

PR #115, opened 2021-09-15
Validate Slot address in storage contracts

### LINKED ISSUE #48: Namespace storage collision detection

leomassazza:
fix #48

This PR uses the AST already found in hardhat dbg artifacts. Then the ast is parsed to find the assignation nodes to identify store.slot occurrences to find duplicates.
It also set some placeholders for the next validations that needs to be done regarding Storage.


packages/deployer/internal/ast-helper.js
```diff
@@ -0,0 +1,142 @@
+const path = require('path');
+const fs = require('fs');
+
+function getContractDataFromDebugFile(debugFileName) {
+  const filename = path.parse(debugFileName).base;
```
mjlescano, line 5:
```suggestion
  const filename = path.basename(debugFileName);
```

packages/deployer/internal/ast-helper.js
```diff
@@ -0,0 +1,142 @@
...
+  const contractName = filename ? filename.substring(0, filename.indexOf('.dbg.json')) : null;
+  const contractPath = path.parse(debugFileName).dir;
```
mjlescano, line 7:
```suggestion
  const { base: filename, dir: contractPath } = path.parse(debugFileName)
  const contractName = filename ? path.basename(filename, '.dbg.json') : null;
```

packages/deployer/subtasks/validate-namespaces.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  logger.checked('Namespaces are valid');
+});
+
+async function _getDebugFiles(hre) {
+  const artifacts = await hre.artifacts.getArtifactPaths();
+
+  return artifacts
+    .filter((filename) => filename.indexOf('contracts/storage/') > 0)
```
mjlescano, line 47: We should use the config here:
```suggestion
  const storageFolder = path.join(hre.config.paths.sources, 'storage')
  return artifacts
    .filter((filename) => filename.startsWith(storageFolder))
```

mjlescano: Hmm, probably we should add the `storage` folder as configuration, as the same we do with `modules`, and then directly use `hre.config.deployer.paths.storage`:

```suggestion
  return artifacts
    .filter((filename) => filename.startsWith(hre.config.deployer.paths.storage))
```

packages/deployer/test/sample-project/contracts/storage/GlobalNamespace.sol
```diff
@@ -10,7 +10,8 @@ contract GlobalNamespace {
     function _globalStorage() internal pure returns (GlobalStorage storage store) {
         assembly {
             // bytes32(uint(keccak256("io.synthetix.global")) - 1)
-            store.slot := 0x8f203f5ee9f9a1d361b4a0f56abfdac49cdd246db58538b151edf87309e955b9
+            store.slot := 0x8f203f5ee9f9a1d361b4a0f56abfdac49cdd246db58538b151edf87309e955b9 // Correct Slot Addres
+            // store.slot := 0x1f33674ed9c09f309c0798b8fcbe9c48911f48b2defee8aecb930c5ef6f80e37 // Duplicate Slot Address
```
ajsantander, line 14: Pls remove

packages/deployer/internal/ast-helper.js
```diff
@@ -0,0 +1,142 @@
+const path = require('path');
```
ajsantander, line 1: I think that this is a good exercise to understand AST's produced by the Solidity compiler, but I think that we should be using a generic library instead of this. This is fragile, and too specific to the case at hand.

packages/deployer/subtasks/validate-namespaces.js
```diff
@@ -0,0 +1,49 @@
...
+
+const {
+  getContractDataFromDebugFile,
+  getDebugDataFile,
+  getSlotAddress,
+  findDuplicateSlots,
+} = require('../internal/ast-helper');
+const { SUBTASK_VALIDATE_NAMESPACES } = require('../task-names');
+
+subtask(SUBTASK_VALIDATE_NAMESPACES).setAction(async (_, hre) => {
```
ajsantander, line 14: IMO this subtask should be renamed to "VALIDATE_STORAGE", and it should  very concisely call functions whose names are very declarative as of what is being validated, e.g:
`_ensureNoRegularStorageSlots(...)`
`_ensureNoDuplicateStorageNamespaces(...)`
`_ensureAppendOnlyMutationsOnNamespaces(...)`
etc...

Right now, if you look at this code, it is not clear until you understand the whole thing, exactly what is being validated.

packages/deployer/subtasks/validate-namespaces.js
```diff
@@ -0,0 +1,49 @@
...
+  logger.subtitle('Validating namespaces');
+
+  const debugFiles = await _getDebugFiles(hre);
```
ajsantander, line 17: Wouldn't this be better named as "namespaceFiles"?

packages/deployer/subtasks/validate-namespaces.js
```diff
@@ -0,0 +1,49 @@
...
+    logger.error(`Duplicate Namespaces found!\n${details.join('')}`);
+    logger.info(`Deleting ${hre.deployer.deployment.file}`);
+    rimraf.sync(path.resolve(hre.config.paths.root, hre.deployer.deployment.file));
+    process.exit(0);
+  }
+
+  logger.checked('Namespaces are valid');
+});
+
+async function _getDebugFiles(hre) {
```
ajsantander, line 43: I'm not sure if just looking at the expected folder where namespaces are declared is enough. Even if, as Mati says, we designate a "storage" folder in the plugin, developers may declare storage namespaces elsewhere, and they would be completely missed by this validation.

ajsantander:
Several non-minor comments. My recommendation is that you use this as the entry point sub-task for a series of AST validations, clearly enumerated in this entry point. And that you don't write the AST traverser yourself.

packages/deployer/subtasks/validate-namespaces.js
ajsantander: Also, not just the location, but the time: We should probably not look at AST compiler data in hardhat, but in deployments. We should store AST trees, so that we can not only look at storage structure, but how it changes over time. This will be needed for determining if the struct within a namespace has been illegally reordered for example.

packages/deployer/subtasks/validate-namespaces.js
leomassazza: it's hardhat name for that file

packages/deployer/internal/ast-helper.js
```diff
@@ -0,0 +1,53 @@
...
+
+function getSlotAddresses(contractName, ast) {
+  const contractNode = findContractNode(ast, contractName);
+  if (!contractNode) {
+    return null;
+  }
+
+  const slots = [];
+  for (const assignment of findAll('YulAssignment', contractNode)) {
+    if (assignment.variableNames[0].name === 'store.slot') {
```
leomassazza, line 11: Note: I didn't find a better way to identify a namespace declaration than looking for the `store.slot := 0x...`

packages/deployer/subtasks/validate-storage.js
```diff
@@ -0,0 +1,75 @@
...
+async function _getSourcesAST(hre) {
+  const fqns = await hre.artifacts.getAllFullyQualifiedNames();
+
+  const contracts = {};
+  for (const fqn of fqns) {
+    const bi = await hre.artifacts.getBuildInfo(fqn);
+    const split = fqn.split(':');
+    const solPath = split[0];
+    const contractName = split[1];
+    // Hardhat will include the contract name when it builds the json only if any source file has more than one contract.
```
ajsantander, line 33: nit: can we make this comment shorter (in chars per line) so that it wraps more nicely?

packages/deployer/subtasks/validate-storage.js
```diff
@@ -0,0 +1,75 @@
...
+  errorsFound = _findInvalidMutationsOnNamespaces() || errorsFound;
+
+  if (errorsFound) {
+    logger.error('Storate usage is not valid');
+    return await hre.run(SUBTASK_CANCEL_DEPLOYMENT);
+  }
+  logger.checked('Namespaces are valid');
+});
+
+async function _getSourcesAST(hre) {
```
ajsantander, line 24: Wondering if this would have a better home in `internal/ast-helper`

packages/deployer/subtasks/validate-storage.js
```diff
@@ -0,0 +1,75 @@
...
+    // "sources": { "contracts/Proxy.sol": { "Proxy": { "ast":... , "id":...} } } (if any source has more than one contract defined)
+    // "sources": { "contracts/Proxy.sol": { "ast":... , "id":...} } (if not any source has more than one contract defined)
+    contracts[contractName] = bi.output.sources[solPath][contractName]
+      ? bi.output.sources[solPath][contractName].ast
+      : bi.output.sources[solPath].ast;
+  }
+  return { contracts };
+}
+
+function _findDuplicateStorageNamespaces(contracts) {
```
ajsantander, line 43: This could also live in `internal/`?

Perhaps we could have a subfolder `internal/ast/` and start putting these as smaller, dedicated files.

packages/deployer/internal/ast-helper.js
ajsantander: Yeah, it's tricky. I'm sure we'll figure out a better way. I'd say at least we should add a ticket in the backlog to improve this.

ajsantander (approved):
Only a few nits/comments

packages/deployer/subtasks/validate-storage.js
leomassazza: like it!

packages/deployer/internal/ast-helper.js
leomassazza: Done

packages/deployer/subtasks/validate-storage.js
leomassazza: done

packages/deployer/subtasks/validate-storage.js
leomassazza: :) done

packages/deployer/package.json
```diff
@@ -20,6 +20,8 @@
     "glob": "7.1.7",
     "mkdirp": "1.0.4",
     "mustache": "4.2.0",
+    "rimraf": "3.0.2",
```
mjlescano, line 23: We already have the `del` dependency for deleting stuff

mjlescano (approved):

PR #116, opened 2021-09-15
Add artifacts files on extended folder

### LINKED ISSUE #39: Artifact extensions: deployment artifacts should keep track of binaries, sources and ABIs
New environment could have the following structure:
```javascript
deployer.deployment = {
  name: '2021-12-12-00-sirius',
  properties: autosaveObject({ ... }), // current deployment file
  abis: autosaveObject({ ... }),
  txs: autosaveObject({ ... }),
  sources: autosaveObject({ ... }),
}
```

mjlescano:
Closes #39

This PR adds the files `extended/${name}.sources.json` & `extended/${name}.abi.json` with the corresponding metadata for each deploy.

## 
* It also includes some refactors on `contractData` passing and processing when deploying contracts.
* When possible, is not manually calculating the path to a contract (e.g.: `contracts/modules/SomeModule.sol`), but instead we're now taking it from the artifact data, to be more consistent with hardhat

### DIFF
--- a/packages/deployer/subtasks/deploy-modules.js
+++ b/packages/deployer/subtasks/deploy-modules.js
@@ -9,7 +9,7 @@ const { SUBTASK_DEPLOY_CONTRACTS, SUBTASK_DEPLOY_MODULES } = require('../task-na
 subtask(SUBTASK_DEPLOY_MODULES).setAction(async (_, hre) => {
   logger.subtitle('Deploying modules');
 
-  const modules = filterValues(hre.deployer.deployment.contracts, (c) => c.isModule);
+  const modules = filterValues(hre.deployer.deployment.data.contracts, (c) => c.isModule);
 
   const deployedSomething = await hre.run(SUBTASK_DEPLOY_CONTRACTS, {
     contracts: modules,
--- a/packages/deployer/subtasks/deploy-modules.js
+++ b/packages/deployer/subtasks/deploy-modules.js
@@ -1,10 +1,12 @@
-const rimraf = require('rimraf');
 const { subtask } = require('hardhat/config');
 
 const logger = require('@synthetixio/core-js/utils/logger');
-const relativePath = require('@synthetixio/core-js/utils/relative-path');
 const filterValues = require('filter-values');
-const { SUBTASK_DEPLOY_CONTRACTS, SUBTASK_DEPLOY_MODULES } = require('../task-names');
+const {
+  SUBTASK_DEPLOY_CONTRACTS,
+  SUBTASK_DEPLOY_MODULES,
+  SUBTASK_CANCEL_DEPLOYMENT,
+} = require('../task-names');
 
 subtask(SUBTASK_DEPLOY_MODULES).setAction(async (_, hre) => {
   logger.subtitle('Deploying modules');
@@ -16,9 +18,7 @@ subtask(SUBTASK_DEPLOY_MODULES).setAction(async (_, hre) => {
   });
 
   if (!deployedSomething) {
-    logger.info('No modules need to be deployed, clearing generated files and exiting...');
-    logger.info(`Deleting ${relativePath(hre.deployer.paths.deployment)}`);
-    rimraf.sync(hre.deployer.paths.deployment);
-    process.exit(0);
+    logger.info('No modules need to be deployed, cancelling execution...');
+    return await hre.run(SUBTASK_CANCEL_DEPLOYMENT);
   }
 });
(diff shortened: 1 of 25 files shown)


packages/deployer/utils/deployments.js
```diff
@@ -97,16 +104,16 @@ function getAllDeploymentFiles(info) {
...
 function getDeploymentFolder(info) {
-  info = _populateDefaults(info);
-
-  return path.join(info.folder, info.network, info.instance);
+  const { folder, network, instance } = _populateDefaults(info);
+  return path.resolve(folder, network, instance);
 }
 
 function _populateDefaults(info) {
-  return { ...info, ...DeploymentInfo };
```
mjlescano, line 106: This is a fix, on the old code the default values were overriding the given ones.

packages/deployer/subtasks/deploy-modules.js
```diff
@@ -1,24 +1,24 @@
...
 
   const deployedSomething = await hre.run(SUBTASK_DEPLOY_CONTRACTS, {
     contracts: modules,
   });
 
   if (!deployedSomething) {
     logger.info('No modules need to be deployed, clearing generated files and exiting...');
-    logger.info(`Deleting ${hre.deployer.deployment.file}`);
-    rimraf.sync(path.resolve(hre.config.paths.root, hre.deployer.deployment.file));
+    logger.info(`Deleting ${relativePath(hre.deployer.paths.deployment)}`);
```
ajsantander, line 20: FYI: I created a new issue for this. We should probably stop doing this (duplicating early exit code) and start routing to the finalize deployment task instead.

packages/deployer/subtasks/prepare-deployment.js
```diff
@@ -25,33 +29,55 @@ subtask(
...
+    alias,
+  });
+  const { sources, abis } = getDeploymentExtendedFiles(currentFile);
+
+  hre.deployer.paths.deployment = currentFile;
+  hre.deployer.paths.sources = sources;
+  hre.deployer.paths.abis = abis;
+
+  hre.deployer.deployment = {
+    data: autosaveObject(currentFile, DEPLOYMENT_SCHEMA),
```
ajsantander, line 57: Rename this to "info" or "general"? Or maybe not... "data" sounds so generic.

packages/deployer/subtasks/prepare-deployment.js
```diff
@@ -25,33 +29,55 @@ subtask(
...
-  hre.deployer.deployment = autosaveObject(currentFile, DEPLOYMENT_SCHEMA);
-  hre.deployer.deployment.file = relativePath(currentFile);
+  const { currentFile, previousFile } = await _determineDeploymentFiles({
+    deployments,
+    deploymentsFolder,
+    alias,
+  });
+  const { sources, abis } = getDeploymentExtendedFiles(currentFile);
+
+  hre.deployer.paths.deployment = currentFile;
```
ajsantander, line 52: This shouldn't be possible since `hre.deployer` is semi-frozen?

I like it tho. Perhaps you should see why the freeze is not working, and be declarative in `extensions.js` with this info.

Also, given that you're doing this, wouldn't it be redundant to also store `hre.deployer.deployment.file`? I'd suggest that we remove the latter if we use the paths defined above.

packages/deployer/tasks/deploy.js
```diff
@@ -49,7 +48,6 @@ task(TASK_DEPLOY, 'Deploys all system modules')
 
     await hre.run(SUBTASK_PREPARE_DEPLOYMENT, taskArguments);
     await hre.run(SUBTASK_PRINT_INFO, taskArguments);
-    await hre.run(TASK_COMPILE, { force: true, quiet: true });
```
ajsantander, line 52: Why is the first compilation moving to `sync-sources.js`?

ajsantander (changes requested):
Looking good. A few minor doubts and change requests.

packages/deployer/tasks/deploy.js
leomassazza: +1 I'd prefer this task to be visible in the deployment process here

leomassazza (approved):
LGTM

packages/deployer/subtasks/prepare-deployment.js
mjlescano: Yup, `hre.deployer.deployment.file` is not there anymore. And the prevent extension is not working because is not deep on its effect. So, the `hre.deployer` object is blocked, but `hre.deployer.paths` not.

mjlescano: Done, fixed it just adding `Object.preventExtensions(hre.deployer.paths);`  :)

packages/deployer/tasks/deploy.js
mjlescano: Did it because the deploy-router was also doing it on the subtask. Moving both compiles to the root :)

packages/deployer/subtasks/prepare-deployment.js
ajsantander: Awesome

ajsantander (approved):

PR #158, opened 2021-09-23
Get Contracts struct map 

leomassazza:
This PR adds a small util to be used later to fix Issue 129 (Storage - check for invalid mutations on structs) // Not referenced to not close it automatically //


leomassazza:
closed in favor of the new one with all the functionality

PR #160, opened 2021-09-23
Do not exit pipeline, plus clean up messaging

### LINKED ISSUE #148: Instead of exiting early when possible, continue, but with no-ops
For example, when the deploy modules subtask finds out that there are no modules to deploy, the deploy task exits.

BUT, it might be nice to have the router generated anyway, validated, etc. Then its deployment would also be a no-op because its compilation would produce the same bytecode.

People may want to run the deployer to validate a router for example.

No-ops should be gray.

ajsantander:
Fix #148

Avoids exiting anywhere in the pipeline when it becomes evident that the system will not change. The pipeline just continues with no-ops so that the user can see deployment info, validations, etc.

Also reviews messaging so that it is more consistent.

### DIFF
--- a/packages/deployer/internal/ast/router-ast-validator.js
+++ b/packages/deployer/internal/ast/router-ast-validator.js
@@ -31,12 +31,14 @@ async function findMissingSelectorsInAST(contracts) {
 }
 
 async function findUnreachableSelectorsInAST() {
-  logger.info('  Unreachable selectors not checked yet in compiled router');
+  logger.info('Unreachable selectors not checked yet in compiled router');
+
   return [];
 }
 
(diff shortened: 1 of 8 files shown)


packages/deployer/internal/ast/router-ast-validator.js
```diff
@@ -31,12 +31,14 @@ async function findMissingSelectorsInAST(contracts) {
 }
 
 async function findUnreachableSelectorsInAST() {
-  logger.info('  Unreachable selectors not checked yet in compiled router');
```
leomassazza, line 34: Please, revert those changes that are fixed on my PR so we don't have a collision here.

leomassazza (approved):

PR #163, opened 2021-09-24
Remove wrong modules source check in router

### LINKED ISSUE #162: Bug: router source validations fail after the linter runs on the router

ajsantander:
Fix #162

This check assumes that the comment  with the module name next to a selector will be in the same line:
```
case 0x1098c085 { result := _SETTINGS_MODULE } // SettingsModule.getASettingValue()
```

However, a linter such as prettier may break the assumption:
```
case 0x1098c085 {
  result := _SETTINGS_MODULE
} // SettingsModule.getASettingValue()
```

Which makes the test fail with a false negative.

Since this kind of thing is now being tested in AST checks anyway, my suggestion is that we simply remove this check, and do not dictate anything related to code formatting.

### DIFF
--- a/packages/deployer/internal/router-source-validator.js
+++ b/packages/deployer/internal/router-source-validator.js
@@ -55,32 +55,7 @@ async function findMissingSelectorsInSource() {
   return errors;
 }
 
-async function findWrongSelectorsInSource() {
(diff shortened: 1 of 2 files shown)


leomassazza (approved):
💯

PR #165, opened 2021-09-24
AST code cleanup

### LINKED ISSUE #157: AST code cleanup before continuing with harder AST tasks

ajsantander:
Fix #157

* Moves ast-helper to core-js.
* Semantic improvements on function names and input parameters with everything regarding AST.
* Other minor fixes.


packages/core-js/utils/ast.js
```diff
@@ -0,0 +1,97 @@
+const { findAll } = require('solidity-ast/utils');
```
ajsantander, line 1: * Moved this file to core-js.
* Only generic AST functions remain.
* Tried to clarify input parameters: when its expecting one ast or many

packages/core-js/utils/hardhat.js
```diff
@@ -0,0 +1,30 @@
+async function getContractAST({ sourceName, contractName }) {
```
ajsantander, line 1: Also created this util for hardhat related operations.

For now it only contains utils to retrieve AST data.

packages/deployer/internal/router-ast-validator.js
```diff
@@ -0,0 +1,100 @@
+const { findYulCaseValues, findFunctionSelectors } = require('@synthetixio/core-js/utils/ast');
+const { getModulesSelectors } = require('./contract-helper');
+const { toPrivateConstantCase } = require('./router-helper');
+const filterValues = require('filter-values');
+
+class RouterASTValidator {
```
ajsantander, line 6: Making these classes makes their usage cleaner.

packages/deployer/internal/router-helper.js
```diff
@@ -9,6 +13,17 @@ function toPrivateConstantCase(name) {
   return name.replace(/(?<![A-Z])[A-Z]/g, '_$&').toUpperCase();
 }
 
+function getRouterSource() {
```
ajsantander, line 16: This use to live in router-source-validator. Figured it could be used elsewhere.

packages/deployer/subtasks/validate-router.js
```diff
@@ -20,36 +11,50 @@ subtask(
 ).setAction(async (_, hre) => {
   logger.subtitle('Validating router');
 
-  const { contracts } = await getSourcesAST(hre);
-  const modules = filterValues(hre.deployer.deployment.general.contracts, (c) => c.isModule);
+  const sourceErrorsFound = await _runSourceValidations();
```
ajsantander, line 14: Who needs comments when a function name explains what a contained piece of code does?

packages/core-js/utils/ast.js
```diff
@@ -0,0 +1,99 @@
...
+  let dependencyContractNodes = [];
+
+  contractNode.linearizedBaseContracts.forEach((baseContractId) => {
+    const dependency = findContractNodeWithId(baseContractId, asts);
+    if (dependency) {
+      dependencyContractNodes.push(dependency);
+    }
+  });
+
+  return dependencyContractNodes;
```
leomassazza, line 39: np: use same return format for all array related on all utils. 
i.e. 
```
let dependencyContractNodes = [];
...
return dependencyContractNodes;

```
or 
```
const slots = [];
...
return slots ? slots : null;

```
(I prefer returning always the array, empty if no results found.)

packages/deployer/internal/storage-ast-validator.js
```diff
@@ -0,0 +1,103 @@
...
+    return ocurrences.length > 0 ? ocurrences : null;
+  }
+
+  findNamespaceCollisions() {
+    const namespaces = [];
+
+    for (var [contractName, ast] of Object.entries(this.asts)) {
+      const slot = findYulStorageSlotAssignments(contractName, ast);
+
+      if (slot) {
```
leomassazza, line 40: if we always return the array on the library this `if` is not needed

packages/deployer/internal/router-ast-validator.js
```diff
@@ -0,0 +1,103 @@
...
+  constructor(asts) {
+    this.asts = asts;
+  }
+
+  async findMissingModuleSelectors() {
+    const moduleSelectors = await getModulesSelectors();
+    const routerSelectors = findYulCaseValues('Router', this.asts['Router']);
+
+    const errors = [];
+    moduleSelectors.forEach((contractSelector) => {
```
leomassazza, line 16: This can fail with the current implementation of findYulCaseValues (that returns the array or null)

leomassazza (changes requested):

packages/core-js/utils/ast.js
ajsantander: Done! All functions now return the array and not null.

packages/deployer/internal/storage-ast-validator.js
ajsantander: Removed

packages/deployer/internal/router-ast-validator.js
ajsantander: No longer a problem cause we return arrays always now

leomassazza (approved):

PR #173, opened 2021-09-27
Add tests to deployer package

### LINKED ISSUE #121: Introduce basic deployer tests (beyond sample-project tests)

mjlescano:
Closes #121 

This PR adds base Unit & Integration tests to use as reference for future tests.

The Deploy script integration tests are disabled for now, because we have to first implement #143 & #193


packages/deployer/.gitignore
```diff
@@ -1,2 +1,4 @@
 test/sample-project/deployments
 test/sample-project/contracts/Router.sol
+test/fixture-projects/*/deployments
+test/fixture-projects/*/contracts/Router.sol
```
ajsantander, line 4: Not sure that we should ignore these:
* deployments - we may wanna keep mainnet deployments around, so perhaps only ignore local ones?
* Router - we may want to keep these around as well. Keep in mind that we have an issue to use deterministic same addressees for all networks, so the Router should be the same no matter the network it targets.

packages/deployer/test/fixture-projects/complete-run/contracts/Proxy.sol
```diff
@@ -0,0 +1,30 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "./storage/ProxyNamespace.sol";
+
+contract Proxy is ProxyNamespace {
```
ajsantander, line 6: Ass discussed in call, let's not duplicate this code and try to re-use sample-project.

It's ok to lose sample-project deployment data.

ajsantander (changes requested):
This is awesome @mjlescano 🙏 

But as discussed, we'll need a few changes as commented.

packages/deployer/.gitignore
mjlescano: Updated it to only ignore `deployments` on `local` & `hardhat`.

But, the `Router` we have to ignore it on the `sample-project` because it includes the branch and commit, and it would be cumbersome for the devs to have to git reset that change every time they do a test on that project and  generate a new Router with a custom branch/commit. Nevertheless, this shouldn't be a problem as the router generation should print always the same.

packages/deployer/test/fixture-projects/complete-run/contracts/Proxy.sol
mjlescano: done!

ajsantander (approved):

leomassazza (approved):
LGTM

PR #176, opened 2021-09-27
Include all source ASTs in extended deployment artifacts

### LINKED ISSUE #156: Include all source ASTs in extended deployment artifacts
We should add the ASTs from all artifacts, not just from compiled contracts.

mjlescano:
Closes #156


packages/core-js/utils/hardhat.js
```diff
@@ -7,7 +7,7 @@ async function getContractAST({ sourceName, contractName }) {
     : output.sources[sourceName].ast;
 }
 
-async function getAllContractASTs(hre) {
+async function getAllContractASTs() {
```
mjlescano, line 10: Removed the `hre` passed by params top be consistent with other internal tools (which use global `hre` definition)

packages/deployer/internal/process-contracts.js
```diff
@@ -1,46 +1,76 @@
...
     bytecode,
     deployedBytecode,
     sourceCode,
     ast,
   };
+
+  // Also init all the sources from the inherited contracts
+  for (const contractName of findInheritedContractNames(ast)) {
+    await initContractSource(contractName);
+  }
```
mjlescano, line 70: This is the big potato to be able to understand the Fix. It saves the sources from the Inherited contracts also. 

My initial idea was to use `glob('contracts/*')` and save all the created sources, but with this we only save the used/deployed ones, buut, including mixins, storage, etc

packages/deployer/subtasks/deploy-contract.js
```diff
@@ -19,15 +18,11 @@ subtask(
...
 
-  const contractArtifacts = await hre.artifacts.readArtifact(contractName);
-  const sourceBytecodeHash = getBytecodeHash(contractArtifacts.deployedBytecode);
-
   // Create contract & start the transaction on the network
   const { contract, transaction } = await _createAndDeployContract(contractName, constructorArgs);
 
   contractData.deployedAddress = contract.address;
   contractData.deployTransaction = transaction.hash;
-  contractData.bytecodeHash = sourceBytecodeHash;
```
mjlescano, line 30: letting the `initContractData()` do this calculation, we save 1 call to `readArtifacts` fn

packages/core-js/utils/hardhat.js
ajsantander: Disagree here.

Internal utils in the deployer can expect a globally injected hre, but generic tools in core-js cannot. This is why I specified it as a parameter.

packages/deployer/internal/contract-helper.js
```diff
@@ -58,9 +59,20 @@ function findDuplicateSelectors(selectors) {
   return ocurrences.length > 0 ? ocurrences : null;
 }
 
+/**
+ * Check if the given contract path is inside the modules folder.
+ * @param {string} contractSource contract path to file, e.g.: contracts/modules/SomeModule.sol
+ * @returns {boolean}
+ */
+function contractIsModule(contractSource) {
```
ajsantander, line 67: nit: parameter could be clearer if named "contractSourcePath"

packages/core-js/utils/hardhat.js
leomassazza: Agree with @ajsantander

packages/deployer/internal/process-contracts.js
leomassazza: Per my understanding this only gets first level. 
Quick question here: We can save all the contracts AST or identify all the candidates and pull AST for the ones that are referenced in any way. Is there any reason to do the extra work and not save all?

leomassazza (approved):

packages/deployer/internal/process-contracts.js
mjlescano: Also , this code allows us to not have to add a special case for saving mixins/storage on another subtask, so, in some way it was easier to implement.

i-stam (approved):

packages/core-js/utils/hardhat.js
mjlescano: fixed!

PR #183, opened 2021-09-28
Implements basic proxy in core-contracts

### LINKED ISSUE #179: Implement base Proxy contract

ajsantander:
Fix #179

### DIFF
--- a/packages/core-contracts/contracts/mocks/proxy/ForwardingProxyMock.sol
+++ b/packages/core-contracts/contracts/mocks/proxy/ForwardingProxyMock.sol
@@ -1,0 +1,38 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/ForwardingProxy.sol";
+
+contract ForwardingProxyMock is ForwardingProxy {
+    bytes32 __SLOT_0;
+    bytes32 __SLOT_1;
+    bytes32 __SLOT_2;
+    bytes32 __SLOT_3;
+    bytes32 __SLOT_4;
+    bytes32 __SLOT_5;
+    bytes32 __SLOT_6;
+    bytes32 __SLOT_7;
+    bytes32 __SLOT_8;
+    bytes32 __SLOT_9;
+
+    address _implementation;
+
+    constructor(address firstImplementation) ForwardingProxy(firstImplementation) {}
+
+    function setImplementation(address newImplementation) external {
+        _setImplementation(newImplementation);
+    }
+
+    function getImplementation() external view returns (address) {
+        return _getImplementation();
+    }
+
+    function _setImplementation(address newImplementation) internal override {
+        _implementation = newImplementation;
+    }
+
+    function _getImplementation() internal override view returns (address) {
+        return _implementation;
+    }
+}
+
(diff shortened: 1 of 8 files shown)


packages/deployer/internal/contract-helper.js
```diff
@@ -69,10 +70,33 @@ function contractIsModule(contractSourcePath) {
   return source.startsWith(`${hre.config.deployer.paths.modules}${path.sep}`);
 }
 
+/**
+ * Get contract source path absolute location. Will try to look it up on the local
+ * hardat folders, or on node_modules.
+ * @param {string} contractSourcePath contract path to file, e.g.: contracts/modules/SomeModule.sol
+ * @returns {string}
+ */
+function getContractFilePath(contractSourcePath) {
```
ajsantander, line 79: This function was added by @mjlescano so that the deployer can see contracts defined in dependencies, in this case `core-contracts`.

packages/core-contracts/contracts/mocks/proxy/ForwardingProxyMock.sol
```diff
@@ -0,0 +1,38 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/ForwardingProxy.sol";
+
+contract ForwardingProxyMock is ForwardingProxy {
+    bytes32 private _slot0;
```
ajsantander, line 7: This mock stores the implementation in a very primitive form: just use the nth slot.

The `ForwardingProxy` just forwards calls, it does not determine how it gets or upgrades its implementation, only that its somewhere in storage.

packages/deployer/internal/contract-helper.js
```diff
@@ -69,10 +70,33 @@ function contractIsModule(contractSourcePath) {
...
+  // First, try to check if its a contract import from the local project
+  const localSource = path.resolve(hre.config.paths.root, contractSourcePath);
+  if (localSource.startsWith(`${hre.config.paths.sources}${path.sep}`)) {
+    if (!fs.existsSync(localSource)) {
+      throw new Error(`Contract file for ${localSource} not found.`);
+    }
+
+    return path.resolve(hre.config.paths.root, contractSourcePath);
```
leomassazza, line 87: this is `localSource` already defined a few lines before

mjlescano (approved):

packages/core-contracts/test/proxy/ForwardingProxy.test.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  describe('when setting ImplementationMockB as the implementation', () => {
+    before('set up proxy and implementation', async () => {
+      let factory;
+
+      factory = await ethers.getContractFactory('ImplementationMockB');
+      Implementation = await factory.deploy();
+
+      factory = await ethers.getContractFactory('ForwardingProxyMock');
```
leomassazza, line 66: if I understand correctly, here we don't want to create a new ForwardingProxy but use the one already created and just change the implementation.
ForwardingProxy.setImplementation(Implementation.address)

leomassazza (approved):
Approved but with one comment on the test that might require a change

packages/core-contracts/test/proxy/ForwardingProxy.test.js
ajsantander: No, `ForwardingProxy` is not intended to be upgradeable. The mock has that function, but the actual contract does and _should_ not.

packages/deployer/internal/contract-helper.js
ajsantander: Fixed

PR #187, opened 2021-09-29
Ownable creation

thanpolas:


packages/core-contracts/contracts/mocks/utils/OwnableParent.sol
```diff
@@ -0,0 +1,12 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../utils/Ownable.sol";
+
+contract OwnableParent is Ownable {
+    constructor(address ownerCtor) Ownable(ownerCtor) {}
+
+    function sum(uint256 a, uint256 b) public view onlyOwner returns (uint256) {
```
ajsantander, line 9: Function should be pure?

packages/core-contracts/contracts/mocks/utils/OwnableParent.sol
```diff
@@ -0,0 +1,12 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../utils/Ownable.sol";
+
+contract OwnableParent is Ownable {
```
ajsantander, line 6: Could you rename this to `OwnableMock` pls?
We're following this pattern with all mocks.

packages/core-contracts/contracts/utils/Ownable.sol
```diff
@@ -0,0 +1,37 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+abstract contract Ownable {
+    address public owner;
+    address public nominatedOwner;
```
ajsantander, line 6: This is a very interesting dilemma that we first bump into with this PR:

This Ownable implementation is perfect for an architecture like Synthetix v2x, but incompatible with the v3 architecture.

I.e. we wouldn't be able to use this in a module in the v3 architecture. The deployer's AST parser would point out the invalid use of storage. See `packages/deployer/test/sample-project/contracts/modules/OwnerModule.sol`.

So, we need to make a decision. Do we make our core-contracts library compatible with the architecture? If so, should we abstract away anything that writes a value to storage?

packages/core-contracts/test/utils/Ownable.test.js
```diff
@@ -0,0 +1,196 @@
+const { ethers } = hre;
+const assert = require('assert/strict');
+
+const { ZERO_ADDRESS } = require('../test-libs/test-lib');
+
+describe('Ownable', () => {
+  describe('Constructor / Deployment', () => {
```
ajsantander, line 7: Pls follow test pattern from other tests.
See `test/ContractUtil.test.js`

We use a before for identifying signers, another for deploying / connecting to contracts, etc.

packages/core-contracts/test/utils/Ownable.test.js
```diff
@@ -0,0 +1,196 @@
...
+    it('Should revert when no argument passed to Ctor', async () => {
+      const OwnableParent = await ethers.getContractFactory('OwnableParent');
+
+      await assert.rejects(
```
ajsantander, line 11: Pls use `assertRevert`.
See `ForwardingProxy.test.js` in the main branch.

packages/core-contracts/test/utils/Ownable.test.js
```diff
@@ -0,0 +1,196 @@
+const { ethers } = hre;
+const assert = require('assert/strict');
+
+const { ZERO_ADDRESS } = require('../test-libs/test-lib');
+
+describe('Ownable', () => {
+  describe('Constructor / Deployment', () => {
+    it('Should revert when no argument passed to Ctor', async () => {
```
ajsantander, line 8: No need to check for incorrect ABI usage. Just for invalid constructor parameters.

packages/core-contracts/test/utils/Ownable.test.js
```diff
@@ -0,0 +1,196 @@
...
+        {
+          name: 'Error',
+          message:
+            `VM Exception while processing transaction: reverted with` +
+            ` reason string 'Owner address cannot be 0x0'`,
+        }
+      );
+    });
+
+    it('Should set owner address on deployment', async () => {
```
ajsantander, line 40: As described above, the pattern we are trying to follow is to deploy only once and operate on that instance, vs deploying on each test. Our `it` cases do only one thing. The whole test spreads out like a tree and traverses all possible branches.

packages/core-contracts/test/utils/Ownable.test.js
```diff
@@ -0,0 +1,196 @@
...
+    });
+
+    it('Should trigger OwnerChanged event on deployment', async () => {
+      const [owner] = await ethers.getSigners();
+      const OwnableParent = await ethers.getContractFactory('OwnableParent');
+      const ownableParent = await OwnableParent.deploy(owner.address);
+      await ownableParent.deployed();
+
+      await new Promise((resolve) => {
+        ownableParent.once('OwnerChanged', (oldOwner, newOwner) => {
```
ajsantander, line 58: We shouldn't check for emitted events like this.
See `packages/deployer/test/sample-project/test/OwnerModule.test.js` L58

ajsantander (changes requested):
Looking good!

Only pointed out some things that need to be changed to follow patterns that we are trying to enforce in Synthetix v3.

packages/core-contracts/test/test-libs/test-lib.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @fileoverview Utilities and helpers for testing.
+ */
+
+const lib = (module.exports = {});
+
+lib.ZERO_ADDRESS = '0x' + '0'.repeat(40);
```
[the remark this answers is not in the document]
thanpolas: @ajsantander can you clarify what is expected to happen here? I understood I need to type 40 zeros in a string literal and that is quite laborious and prone to errors. Please share the reasoning.

packages/core-contracts/test/utils/Ownable.test.js
thanpolas: @ajsantander please confirm and explain the reasoning. Retaining the same state across test cases is an anti-pattern that will produce flaky tests and forbid parallelizations of test running.

Also mocha has a context (`this`) and it's encouraged to use it when setting up test cases vs having global variables like in the `test/ContractUtil.test.js` test module.

packages/core-contracts/test/utils/Ownable.test.js
thanpolas: I'll mark this as resolved as it is a duplicate of the comment above that you mention: https://github.com/Synthetixio/synthetix-v3/pull/187#discussion_r718766820

packages/core-contracts/contracts/mocks/utils/OwnableParent.sol
thanpolas: Will do, but I have a couple notes I'd like to be considered:

1. This particular module is not a mock. It inherits `Ownable` so it can be tested.
2. Could we move all test related modules under the corresponding test/ folders vs being inside the production folders?

ajsantander: 1. I understand the subtlety but IMO we should not refine the naming conventions for now. Atm, anything that needs to be modified to be testable is being added the word "Mock".
2. No, because you wouldn't be able to compile them.

packages/core-contracts/test/test-libs/test-lib.js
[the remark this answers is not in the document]
ajsantander: The reasoning is explained above. It is easier to input the value than to import it from a library. We discussed this as a team and agreed on it. We could change it in the future, but the current agreement is that.

packages/core-contracts/test/utils/Ownable.test.js
ajsantander: Again, this is the pattern we are following. I am familiar with your reasoning, I am just asking you to adapt to our current practices.

ajsantander (approved):

packages/core-contracts/contracts/mocks/utils/OwnableMock.sol
```diff
@@ -0,0 +1,37 @@
...
+    address public override owner;
+    address public override nominatedOwner;
+
+    constructor(address firstOwner) {
+        require(firstOwner != address(0), "Owner cannot be 0x0");
+        owner = firstOwner;
+        emit OwnerChanged(address(0), firstOwner);
+    }
+
+    function nominateNewOwner(address ownerNominated) external override onlyOwner {
```
ajsantander, line 16: These functions shouldn't be implemented in the mock, but in Ownable.sol.

The only abstract __internal methods in Ownable.sol are `_setOwner`, _getOwner`, `_setNominatedOwner`, and _getNominatedOwner`.

So Ownable defines the functionality but does not specify where things are stored. The mock overrides the functions above to store these in regular storage.

packages/core-contracts/contracts/utils/Ownable.sol
```diff
@@ -0,0 +1,23 @@
...
+
+import "../interfaces/IOwnable.sol";
+
+abstract contract Ownable is IOwnable {
+    modifier onlyOwner() {
+        _onlyOwner();
+        _;
+    }
+
+    function acceptOwnership() external virtual override {}
```
ajsantander, line 12: All these are implemented in this contract.

When owner is set for example, they call `_setOwner(newOwner)`

ajsantander (changes requested):
Commented some nits and an important required change.

Also, could you modify the sample project to use Ownable from core-js?
(feel free to remove redundant tests from sample-projects that are tested in core-js)

packages/deployer/subtasks/validate-router.js
```diff
@@ -51,7 +51,7 @@ async function _runASTValidations() {
 
   logger.debug('Validating Router compiled code');
   errorsFound.push(...(await validator.findMissingModuleSelectors()));
-  errorsFound.push(...(await validator.findUnreachableModuleSelectors()));
+  // errorsFound.push(...(await validator.findUnreachableModuleSelectors()));
```
ajsantander, line 54: @leomassazza we had to disable this check because the ast verification fails with the new inheritance tree

PR #197, opened 2021-09-30
Erc20 contract

### LINKED ISSUE #169: Design basic ERC20 token

i-stam:
Fix #169 
ERC20 core-contract implementation:

 Please take a look at the following and comment if needed:
- Increase/Decrease-allowance to be added in a future PR if needed
- transfer to 0x0 allowed.
- functions revert on over/under-flow. Require statements can be added (the OZ way) to get a more relevant revert reason, e.g. `transfer: not enough balance`
- No infinite approval i.e. if set to max value then `allowance` is never decreased (gas optimization) unless it is reset to another value again.


packages/core-contracts/contracts/mocks/utils/ERC20Mock.sol
```diff
@@ -0,0 +1,20 @@
+//SPDX-License-Identifier: Unlicense
```
ajsantander, line 1: ultra nit: but why have different versions of "Unlicense"/"UNLICENSED"?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
+//SPDX-License-Identifier: UNLICENSED
+
```
ajsantander, line 2: nit: Not using spaces here in other contracts

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
...
+abstract contract ERC20 is IERC20 {
+    string public override name;
+
+    string public override symbol;
+
+    uint8 public immutable override decimals;
+
+    mapping(address => uint256) public override balanceOf;
+
+    mapping(address => mapping(address => uint256)) public override allowance;
```
ajsantander, line 16: Should we stick to the pattern of always using private variables and exposing them via our own getters?
I.e. rename this to `private _allowances`

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
...
+        return true;
+    }
+
+    function transferFrom(
+        address from,
+        address to,
+        uint amount
+    ) external override returns (bool) {
+        // it reverts if the sender does not have sufficient allowance
+        allowance[from][msg.sender] -= amount;
```
ajsantander, line 50: What if `from = msg.sender`? Is it ERC20 compatible to revert in those cases?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
...
+import "../interfaces/IERC20.sol";
+
+abstract contract ERC20 is IERC20 {
+    string public override name;
+
+    string public override symbol;
+
+    uint8 public immutable override decimals;
+
+    mapping(address => uint256) public override balanceOf;
```
ajsantander, line 14: Doesn't this collide with the `balanceOf` function?

packages/core-contracts/contracts/utils/ERC20.sol
ajsantander: Also, this would produce an error that is not very explicit.

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
...
+        return true;
+    }
+
+    function _transfer(
+        address from,
+        address to,
+        uint256 amount
+    ) private {
+        // it reverts if the sender does not have sufficient balance
+        balanceOf[from] -= amount;
```
ajsantander, line 63: Again, this would produce a not very explicit error. Is it worth the gas savings?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
+//SPDX-License-Identifier: UNLICENSED
+
+pragma solidity ^0.8.0;
+
+import "../interfaces/IERC20.sol";
+
+abstract contract ERC20 is IERC20 {
+    string public override name;
+
```
leomassazza, line 9: nit: same for the space between variables. On other contracts it was used just to separate blocks if adds readability (otherwise it reduce de lines seen in a single screen)

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,93 @@
...
+        address from,
+        address to,
+        uint256 amount
+    ) private {
+        // it reverts if the sender does not have sufficient balance
+        balanceOf[from] -= amount;
+        // we are now sure that we can perform this operation safely since it didn't revert in the previous step
+        // i.e. the total supply cannot exceed the maximum value of uint256
+        unchecked {
+            balanceOf[to] += amount;
```
ajsantander, line 67: Theoretically, there could be an overflow here right?

packages/core-contracts/contracts/mocks/utils/ERC20Mock.sol
```diff
@@ -0,0 +1,20 @@
...
+pragma solidity ^0.8.0;
+
+import "../../utils/ERC20.sol";
+
+contract ERC20Mock is ERC20 {
+    constructor(
```
leomassazza, line 7:
```suggestion
    // solhint-disable-next-line no-empty-blocks  
    constructor(
```

leomassazza (approved):
Approved with some NP

packages/core-contracts/contracts/mocks/utils/ERC20Mock.sol
i-stam: I am trying to fix it. The compiler suggests `UNLICENSED` in capital letters so I am replacing all occurrences with it.

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: fixed

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: So this si something I thought a bit about. I went for code efficiency. If we decide to go with this approach we should follow it in all contracts and we will end up bloating the contract with getters.

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: It is erc20 compatible. The other approach will be to have explicit requirements e.g.  `require(allowance >= amount)`
and then do the unchecked operation. But in my opinion an underflow error there is quite informative and it is whatyou get in most cases when using safe math.

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: nope, solc 0.7 I think allows public vars to override functions. The getter will be produced no matter what.

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: this is actually enforced by the linter!

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: nope: the implication is that since we have subtracted above, it means that everything is fine. The total supply would have overflowed before that as when we increase the balance then the supply is also increased.

packages/core-contracts/contracts/mocks/utils/ERC20Mock.sol
i-stam: again when i save the file the linter remove the blank line!

ajsantander (changes requested):
My only request would be to emit more explicit error messages

ajsantander (approved):

packages/core-contracts/contracts/interfaces/IERC20.sol
```diff
@@ -0,0 +1,31 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+interface IERC20 {
+    // Events
```
ajsantander, line 5: Remove this comment?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,101 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "../interfaces/IERC20.sol";
+
+abstract contract ERC20 is IERC20 {
```
ajsantander, line 6: Should this contract live in a `token` folder instead of `utils`?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,101 @@
...
+    string public override name;
+
+    string public override symbol;
+
+    uint8 public immutable override decimals;
```
ajsantander, line 11: Isn't it weird to make this immutable and not the others because Solidity does not allow it?

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,101 @@
+//SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "../interfaces/IERC20.sol";
+
+abstract contract ERC20 is IERC20 {
+    string public override name;
```
ajsantander, line 7: I don't looooooove to have to override a variable that was never _directly_ declared in the interface... but well, fair I guess.

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,101 @@
...
+        return true;
+    }
+
+    function transferFrom(
+        address from,
+        address to,
+        uint amount
+    ) external override returns (bool) {
+        uint256 currentAllowance = allowance[from][msg.sender];
+        // it reverts if the sender does not have sufficient allowance
```
ajsantander, line 49: This comment is redundant now that the custom revert reason is in place.

packages/core-contracts/contracts/utils/ERC20.sol
```diff
@@ -0,0 +1,101 @@
...
+        require(currentAllowance >= amount, "Amount exceeds allowance");
+        unchecked {
+            allowance[from][msg.sender] -= amount;
```
ajsantander, line 52: Shouldn't any changes in approval emit Approval events?
(not sure what the spec has to say about this)

packages/core-contracts/contracts/utils/ERC20.sol
i-stam: yeap, I was wondering about the naming, utils was the easy choice.

PR #211, opened 2021-10-04
find selectors in dependencies

### LINKED ISSUE #210: Router verifier AST wont find selectors in dependencies

leomassazza:
Fix #210 
This small PR fixes a bug that prevented the Router AST validator to correctly find selectors in dependencies.

### DIFF
--- a/packages/core-js/utils/ast/finders.js
+++ b/packages/core-js/utils/ast/finders.js
@@ -1,13 +1,13 @@
 const { findAll } = require('solidity-ast/utils');
 
-function findContractNodeWithId(contractId, asts) {
-  return Array.from(findAll('ContractDefinition', asts)).find(
+function findContractNodeWithId(contractId, ast) {
+  return Array.from(findAll('ContractDefinition', ast)).find(
     (contractDefiniton) => contractDefiniton.id === contractId
   );
 }
 
-function findContractNodeWithName(contractName, asts) {
-  return Array.from(findAll('ContractDefinition', asts)).find(
+function findContractNodeWithName(contractName, ast) {
+  return Array.from(findAll('ContractDefinition', ast)).find(
     (contractDefiniton) => contractDefiniton.name === contractName
   );
 }
@@ -34,9 +34,11 @@ function findContractDependencies(contractName, asts) {
   let dependencyContractNodes = [];
 
   contractNode.linearizedBaseContracts.forEach((baseContractId) => {
-    const dependency = findContractNodeWithId(baseContractId, asts);
-    if (dependency) {
-      dependencyContractNodes.push(dependency);
+    for (var [, ast] of Object.entries(asts)) {
+      const dependency = findContractNodeWithId(baseContractId, ast);
+      if (dependency) {
+        dependencyContractNodes.push(dependency);
+      }
     }
   });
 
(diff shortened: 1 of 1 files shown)


packages/core-js/utils/ast/finders.js
```diff
@@ -1,13 +1,13 @@
 const { findAll } = require('solidity-ast/utils');
 
-function findContractNodeWithId(contractId, asts) {
```
leomassazza, line 3: Fix naming. It checks on a single AST not on the asts object containing { 'contractNameX': {}, 'contractNameY': {}... }

packages/core-js/utils/ast/finders.js
```diff
@@ -1,13 +1,13 @@
...
-  return Array.from(findAll('ContractDefinition', asts)).find(
+function findContractNodeWithId(contractId, ast) {
+  return Array.from(findAll('ContractDefinition', ast)).find(
     (contractDefiniton) => contractDefiniton.id === contractId
   );
 }
 
-function findContractNodeWithName(contractName, asts) {
```
leomassazza, line 9: same naming fix here

packages/core-js/utils/ast/finders.js
```diff
@@ -82,19 +84,30 @@ function findYulCaseValues(contractName, ast) {
...
+      selectors.push({ selector: '0x' + functionDefinition.functionSelector });
+    }
+  }
+
+  return selectors;
+}
+
+function findFunctionSelectors(contractName, asts) {
+  const selectors = [];
+  for (const contractNode of findContractDependencies(contractName, asts)) {
```
leomassazza, line 101: checks all dependencies to find the selectors

packages/core-js/utils/ast/finders.js
```diff
@@ -82,19 +84,30 @@ function findYulCaseValues(contractName, ast) {
...
+function _findFunctionSelectors(contractNode) {
   const selectors = [];
-  const contractNode = asts[contractName];
-  if (!contractNode) {
-    return selectors;
-  }
 
   for (const functionDefinition of findAll('FunctionDefinition', contractNode)) {
-    selectors.push({ selector: '0x' + functionDefinition.functionSelector });
+    if (functionDefinition.functionSelector) {
```
leomassazza, line 91: this if is to exclude FunctionDefinitions that doesn't have a selector (i.e. internal functions). Previously we were adding lots of '0xundefined' to the array.

packages/core-js/utils/ast/finders.js
```diff
@@ -34,9 +34,11 @@ function findContractDependencies(contractName, asts) {
   let dependencyContractNodes = [];
 
   contractNode.linearizedBaseContracts.forEach((baseContractId) => {
-    const dependency = findContractNodeWithId(baseContractId, asts);
-    if (dependency) {
-      dependencyContractNodes.push(dependency);
+    for (var [, ast] of Object.entries(asts)) {
```
leomassazza, line 37: it didn't work before ¯\_(ツ)_/¯

packages/core-js/utils/ast/finders.js
ajsantander: I believe this is not true. Pls look at like 37 in this file.

packages/core-js/utils/ast/finders.js
ajsantander: This one looks correct tho.

packages/core-js/utils/ast/finders.js
ajsantander: Why not? I had tested this manually.
I remember adding a state declaration to a namespace, which was inherited from a mixin, which was inherited from a module, and the ast validation was detected.

packages/core-js/utils/ast/finders.js
```diff
@@ -82,19 +84,30 @@ function findYulCaseValues(contractName, ast) {
   return items;
 }
 
-function findFunctionSelectors(contractName, asts) {
+function _findFunctionSelectors(contractNode) {
```
ajsantander, line 87: Instead of making the single contract case private, could we also expose it and use a name for each that reflects the difference?

ajsantander (approved):
Approved. Just consider my comments above.

packages/core-js/utils/ast/finders.js
leomassazza: looks like it worked when we were based on HH build-info but not using the object as we build it (object with several entries like 'contract_name': {the ast}.

packages/core-js/utils/ast/finders.js
leomassazza: used to work... :( that's why we need test coverage :)

packages/core-js/test/utils/finders.spec.js
```diff
@@ -0,0 +1,30 @@
+const { equal } = require('assert/strict');
+const { findContractDependencies, findFunctionSelectors } = require('../../utils/ast/finders');
+const asts = require('../fixtures/asts.json');
+
+describe('utils/ast/comparator.js compareStorageStructs', function () {
```
leomassazza, line 5: added some basic tests. We can populate it later with more complex conditions and all the functions exported. The task is already in our backlog

mjlescano (approved):

PR #215, opened 2021-10-05
Implement universal proxy contract

### LINKED ISSUE #166: Implement UniversalProxy contract

leomassazza:
Fix #166 

This PR adds the implementation of the Universal Proxy to core-contracts.

**WARNING: TEMPORARILY DISABLING ANTI-BRICKING PROTECTION.**

I disabled anti-bricking protection since as it is is not working and it wasn't tested properly on deployer tests. Issue 214 created and added to this sprint to bring it back.


packages/core-contracts/contracts/mocks/proxy/UniversalProxyImplementationMockB.sol
```diff
@@ -0,0 +1,48 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/UniversalProxyImplementation.sol";
+
+contract UniversalProxyImplementationMockB is UniversalProxyImplementation {
+    uint private _a;
+    bytes32 private _b;
+
+    bytes32 private _slot2;
```
leomassazza, line 10: skipping slots 0 and 1 to preserve the implementation location.

packages/core-contracts/contracts/mocks/proxy/UniversalProxyImplementationMockA.sol
```diff
@@ -0,0 +1,40 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/UniversalProxyImplementation.sol";
+
+contract UniversalProxyImplementationMockA is UniversalProxyImplementation {
+    uint private _a;
+
+    bytes32 private _slot1;
```
leomassazza, line 9: skipping slot 0  to preserve the implementation location.

packages/core-contracts/contracts/mocks/proxy/UniversalProxyImplementationMockB.sol
```diff
@@ -0,0 +1,48 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/UniversalProxyImplementation.sol";
+
+contract UniversalProxyImplementationMockB is UniversalProxyImplementation {
+    uint private _a;
+    bytes32 private _b;
```
leomassazza, line 8: set to bytes32 instead of structured string to have more control over the memory size of the slot.

packages/deployer/test/fixture-projects/sample-project/test/UpgradeModule.test.js
```diff
@@ -55,7 +55,7 @@ describe('UpgradeModule', () => {
     });
   });
 
-  describe('when the owner attempts to upgrade to a sterile implementation', () => {
+  describe.skip('when the owner attempts to upgrade to a sterile implementation', () => {
```
leomassazza, line 58: skipped since the anti-bricking protection is disabled. Will be set back (and moved to core-contracts) with issue 214

packages/deployer/test/fixture-projects/sample-project/test/UpgradeModule.test.js
```diff
@@ -86,7 +86,7 @@ describe('UpgradeModule', () => {
     });
   });
 
-  describe('when attempting to destroy the implementation with a malicious contract', () => {
```
leomassazza, line 89: skipped since the anti-bricking protection is disabled. Will be set back (and moved to core-contracts) with issue 214

packages/core-contracts/contracts/mocks/proxy/UniversalProxyImplementationMockA.sol
```diff
@@ -0,0 +1,40 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/UniversalProxyImplementation.sol";
+
+contract UniversalProxyImplementationMockA is UniversalProxyImplementation {
```
ajsantander, line 6: Do you think that both `UniversalProxyImplementationMockA` and `UniversalProxyImplementationMockB` could inherit from a `UniversalProxyImplementationMock` that sets the implementation in slot 0 and implements the implementation setter/getter? That way, mock A and B don't need to do this weird slot shifting; they just don't use the first slot.

packages/core-contracts/test/proxy/UniversalProxy.test.js
```diff
@@ -0,0 +1,142 @@
+const { ethers } = hre;
+const assert = require('assert');
+const { assertRevert } = require('@synthetixio/core-js/utils/assertions');
+const { findEvent } = require('@synthetixio/core-js/utils/events');
+
+describe('UniversalProxy', () => {
+  let ForwardingProxy, Instance, Implementation;
```
ajsantander, line 7: Could we rename `ForwardingProxy` both in this test and in the other test? Not the contract name, but how we call them in the test. I think it will make the tests more readable.

We don't really care what kind of proxy this is, just that it is a proxy.

packages/core-contracts/test/proxy/UniversalProxy.test.js
```diff
@@ -0,0 +1,142 @@
...
+        });
+
+        it('reverts', async () => {
+          await assertRevert(BadInstance.getB(), 'function selector was not recognized');
+        });
+      });
+    });
+  });
+
+  describe('when attempts to upgrade to an EOA', () => {
```
ajsantander, line 67: I think these blocks need to be nested in the previous describe, since they assume that the previous things have happened right?

packages/core-contracts/test/proxy/UniversalProxy.test.js
```diff
@@ -0,0 +1,142 @@
...
+
+      await assertRevert(Instance.upgradeTo(user.address), 'Implementation not a contract');
+    });
+  });
+
+  describe('when upgrading to a non-sterile implementation', () => {
+    let receipt, UpgradedImplementation, UpgradedInstance;
+
+    before('upgrade', async () => {
+      let factory;
```
ajsantander, line 79: No need to declare factory like this since it will only be used once.

packages/core-contracts/contracts/mocks/Destroyer.sol
```diff
@@ -0,0 +1,23 @@
+//SPDX-License-Identifier: Unlicense
```
ajsantander, line 1: Should we get rid of this one until we need it back?

I.e. if its not used in the current state, it should be deleted IMO.

packages/deployer/test/fixture-projects/sample-project/test/UpgradeModule.test.js
ajsantander: What do you think about making the sample-project tests "dumber"? I.e. not check for the full functionality, but rather just check that it can upgrade.

IMO the full tests coverage should live in core-contracts and this should just serve as an example.

ajsantander (approved):

packages/core-contracts/contracts/mocks/proxy/UniversalProxyImplementationMockA.sol
leomassazza: in that case I prefer to keep it visible that there's a need to set "manually" the implementation location

packages/deployer/test/fixture-projects/sample-project/test/UpgradeModule.test.js
leomassazza: is what I said in the comment. It will be moved to core-contracts together with the fix for issue 214

ajsantander (approved):

i-stam (approved):

PR #223, opened 2021-10-06
Bring back Anti-bricking mechanism to UniversalProxyImplementation

leomassazza:


packages/core-contracts/contracts/mocks/proxy/Bricker.sol
```diff
@@ -0,0 +1,43 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "../../proxy/UniversalProxyImplementation.sol";
+
+contract Bricker is UniversalProxyImplementation {
```
leomassazza, line 6: This contracts test one case that I found not covered originally that is a contract that adheres to the paradigm but was poorly (or intentionality bad) developed.

packages/core-contracts/contracts/proxy/UniversalProxyImplementation.sol
```diff
@@ -2,24 +2,73 @@
...
         require(_isContract(newImplementation), "Implementation not a contract");
+        require(_canUpgradeInTheFuture(newImplementation), "Implementation is sterile");
 
         _setImplementation(newImplementation);
 
         emit Upgraded(newImplementation);
     }
+
+    function _canUpgradeInTheFuture(address newImplementation) private returns (bool) {
+        if (_getIsUpgrading()) {
```
leomassazza, line 31: one of the issues of the original implementation was that in a real upgrade it will enter into a recursive loop swapping old and new implementation address and will revert at the end because of hitting the limit of function calls.

This happened since simulateUpgrade will do the swap here:
```
        _setImplementation(newImplementation);
        (bool rollbackSuccessful, ) = newImplementation.delegatecall(
            abi.encodeWithSignature("upgradeTo(address)", oldImplementation)
        );
```
and the original verification, the if condition will never result in true
```
       if (newImplementation == getImplementation()) {
            return true;
        }
```

Issue #224, opened 2021-10-06
When deleting a module after anew deployment, the following one fails


mjlescano:
Steps to  repro:
1. Do a `--clean` deployment
2. Delete any Module from `contracts/modules`
3. Try to deploy again, without `--clean`, it will fail.

**Important!**
After fixing, handle TODOs in `deploy.spec.js`
https://github.com/Synthetixio/synthetix-v3/pull/225/files#diff-aac39963cd93065b128b68f10fa781d224942fc7fbf0e1c907d6cd3e06361cc4R71

PR #225, opened 2021-10-06
Add reusable test initializers on sample-project

### LINKED ISSUE #193: Also use test initializers (or similar) in deployer tests
Atm if you run a clean local deployment, and then another one without `--clear` upgrading the main proxy does not work. This is because OwnerModule's owner is not set, and thus, no one can call upgradeTo.

We've previously discussed this problem, and decided that the deployer should not do any initializations at all. This should be done by utilities during testing, and migrations in production.

Right now the deployer's sample-project tests do this via `helpers/initializer.js`, but the deployer tests don't. We should do something similar to solve the initialization problem in these, and perhaps even the same file could be reused.

mjlescano:
Closes #193 

Bug discovered: https://github.com/Synthetixio/synthetix-v3/issues/224
Almost closes: https://github.com/Synthetixio/synthetix-v3/issues/224

This PR refactors (and abstracts) sample-project test initializer on a single file. And from the deployer tests, we assume that the given sample project has a `test/helpers/initializer` file that implements the `deploy` & `init` methods.

It also fixes several renaming bugs on `OwnerModule` that caused the tests at `sample-project` to fail.

Also, thanks to this fix, we bump `deployer` tests coverage from like 5% to 87% 📈

### DIFF
--- a/packages/deployer/test/helpers.js
+++ b/packages/deployer/test/helpers.js
@@ -4,16 +4,6 @@ const { request } = require('http');
 const { resetHardhatContext } = require('hardhat/plugins-testing');
 const { takeSnapshot, restoreSnapshot } = require('@synthetixio/core-js/utils/rpc');
 
-function _getEnvironmentPath(fixtureProjectName) {
-  const pathname = path.join(__dirname, 'fixture-projects', fixtureProjectName);
-
-  if (!fs.existsSync(pathname)) {
-    throw new Error(`Invalid fixture project ${fixtureProjectName}`);
-  }
-
-  return pathname;
-}
-
 function useEnvironment(fixtureProjectName) {
   // TODO: enable tests when the connection to the hardhat network works
   // More info: https://github.com/Synthetixio/synthetix-v3/issues/143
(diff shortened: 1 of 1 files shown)


packages/deployer/test/fixture-projects/sample-project/test/AnotherModule.test.js
```diff
@@ -1,28 +1,25 @@
...
-const { getProxyAddress } = require('../../../../utils/deployments');
+const { getProxyAddress } = require('@synthetixio/deployer/utils/deployments');
 const { printGasUsed } = require('@synthetixio/core-js/utils/tests');
-const { bootstrap, initializeSystem } = require('./helpers/initializer');
 const { findEvent } = require('@synthetixio/core-js/utils/events');
+const bootstrap = require('./helpers/bootstrap');
 
 describe('AnotherModule', () => {
-  bootstrap();
+  const { info, init } = bootstrap();
```
mjlescano, line 10: Now  we are using a custom instance name (`test`), so we have to give it to `deployments` util

i-stam (approved):

PR #231, opened 2021-10-08
Fix modules updates

### LINKED ISSUE #168: Add a black boxed tests for the Generated Router

### LINKED ISSUE #224: When deleting a module after anew deployment, the following one fails

mjlescano:
Fixes #224
Closes #168

Coverage after being able to enable updates test is on `91%~` 🙌 :
Captura de Pantalla 2021-10-08 a la(s) 11 44 17

### DIFF
--- a/packages/deployer/test/src/tasks/deploy.test.js
+++ b/packages/deployer/test/src/tasks/deploy.test.js
@@ -61,7 +61,6 @@ describe('tasks/deploy.js', function () {
 
       await this.deploySystem({
         alias: 'third',
-        quiet: false,
       });
     } finally {
       // Restore all the changes


packages/deployer/tasks/deploy.js
```diff
@@ -51,13 +51,13 @@ task(TASK_DEPLOY, 'Deploys all system modules')
 
       await hre.run(SUBTASK_PREPARE_DEPLOYMENT, taskArguments);
       await hre.run(SUBTASK_PRINT_INFO, taskArguments);
-      await hre.run(TASK_COMPILE, { force: false, quiet: true });
+      await hre.run(TASK_COMPILE, { force: true, quiet: true });
```
mjlescano, line 54: This is the real fix for the PR. Before hadhat wasn't refreshing build info when deleting a contract. This will cause the  build to take a little longer, but is necessary for our use case.

PR #267, opened 2021-10-15
Improve slot detection

### LINKED ISSUE #125: Improve namespace collision detection
Currently we look for yul assignment to store.slot := 0x...

leomassazza:
Fix #125

### DIFF
--- a/packages/deployer/internal/storage-ast-validator.js
+++ b/packages/deployer/internal/storage-ast-validator.js
@@ -15,16 +15,16 @@ class ModuleStorageASTValidator {
 
   findDuplicateNamespaces(namespaces) {
     const duplicates = namespaces
-      .map((s) => s.position)
+      .map((s) => s.slot)
       .filter((s, index, namespaces) => namespaces.indexOf(s) !== index);
 
     const ocurrences = [];
 
     if (duplicates.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/deployer/internal/storage-ast-validator.js
```diff
@@ -15,16 +15,16 @@ class ModuleStorageASTValidator {
 
   findDuplicateNamespaces(namespaces) {
     const duplicates = namespaces
-      .map((s) => s.position)
+      .map((s) => s.slot)
```
leomassazza, line 18: Rename position as slot since is the right name.

packages/deployer/internal/storage-ast-validator.js
```diff
@@ -39,25 +39,63 @@ class ModuleStorageASTValidator {
...
       errors.push({
-        msg: `Duplicate namespaces found!\n${details.join('')}`,
+        msg: `Duplicate namespaces slot found!\n${details.join('')}`,
       });
     }
 
     return errors;
   }
 
+  findNamespaceSlotChanges() {
```
leomassazza, line 61: This new function will check the new namespace slots against the previous one to detect any deletion or change in the slot address.

PR #279, opened 2021-10-27
Add deployer tests

### LINKED ISSUE #240: Increase deployer line test coverage to 95%

mjlescano:
Closes #240

### DIFF
--- a/packages/deployer/subtasks/validate-router.js
+++ b/packages/deployer/subtasks/validate-router.js
@@ -5,6 +5,8 @@ const { initContractData } = require('../internal/process-contracts');
 const RouterSourceValidator = require('../internal/router-source-validator');
 const RouterASTValidator = require('../internal/router-ast-validator');
 const { ContractValidationError } = require('../internal/errors');
+const { getModulesSelectors } = require('../internal/contract-helper');
+const { getRouterSource } = require('../internal/router-helper');
 const { SUBTASK_VALIDATE_ROUTER } = require('../task-names');
 
 subtask(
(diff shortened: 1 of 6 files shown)


packages/deployer/subtasks/deploy-contracts.js
```diff
@@ -62,7 +62,7 @@ async function _processContracts(contracts) {
   const toCreate = [];
 
   for (const [contractName, contractData] of Object.entries(contracts)) {
-    if (hre.network.name === 'hardhat' || !contractData.deployedAddress) {
+    if (!contractData.deployedAddress) {
```
mjlescano, line 65: There is no need to handle the special case of having the deployment files have something that is not on the blockchain (in this case for the hardhat network, that resets everything on every run). If we want to use the hardhat network outside of tests, we have to just use `--clear` instead.

packages/deployer/subtasks/print-info.js
```diff
@@ -11,12 +11,6 @@ const { readPackageJson } = require('@synthetixio/core-js/utils/npm');
 const { SUBTASK_PRINT_INFO } = require('../task-names');
 
 subtask(SUBTASK_PRINT_INFO, 'Prints info about a deployment.').setAction(async (taskArguments) => {
-  const { quiet } = taskArguments;
-
-  if (quiet) {
-    return;
-  }
-
```
mjlescano, line 19: Removing this we still execute the logger so is covered on coverage, but nevertheless, it doesn't prints nothing because is configured on silent mode at the beggining of the deploy task.

packages/deployer/subtasks/validate-router.js
```diff
@@ -28,7 +30,10 @@ subtask(
 async function _runSourceValidations() {
   const errorsFound = [];
 
-  const validator = new RouterSourceValidator();
+  const validator = new RouterSourceValidator({
+    getModulesSelectors,
+    getRouterSource,
+  });
```
mjlescano, line 36: I had to move this here so we were able to unit test the validator helper without having to mock an entire project with build and everything.

leomassazza (approved):
LGTM

PR #309, opened 2021-11-03
BeaconProxy storage

### LINKED ISSUE #310: Implement BeaconProxy storage

i-stam:
FIX #310  
This is needed for the SynthFactory module.
With this approach, the beacon can be updated i.e. some synth can use a diff implementation
If we want the beacon to be monolithic we can use an immutable variable. Check this:
https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2538


i-stam:
Outdated. Addressed in #383

PR #315, opened 2021-11-03
Add Proxy flavors 

leomassazza:


packages/core-contracts/contracts/proxy/OwnableUUPSImplementation.sol
```diff
@@ -0,0 +1,68 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./UniversalProxyImplementation.sol";
+import "../ownership/Ownable.sol";
+
+contract OwnableUUPSImplementation is UniversalProxyImplementation, Ownable {
```
ajsantander, line 7: Rename `UniversalProxyImplementation` to `UUPSImplementation`?

packages/core-contracts/contracts/proxy/OwnableUUPSImplementation.sol
```diff
@@ -0,0 +1,68 @@
...
+    struct UnstructuredProxyStorage {
```
ajsantander, line 8: You only need to destructure  a single variable in storage, the implementation address. That's it, no storage namespaces at all.

packages/core-contracts/contracts/proxy/UnstructuredProxy.sol
```diff
@@ -0,0 +1,33 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./ForwardingProxy.sol";
+
+contract UnstructuredProxy is ForwardingProxy {
+    struct UnstructuredProxyStorage {
```
ajsantander, line 7: No need to use a namespace.

packages/core-contracts/contracts/proxy/OwnableUUPSImplementation.sol
```diff
@@ -0,0 +1,68 @@
...
+        address implementation;
+        bool simulatingUpgrade;
+    }
+
+    struct OwnableStorage {
```
ajsantander, line 13: Ownable should not be implemented at this level. For example, SNXToken.sol inherits UUPSImplementation and Ownable.

leomassazza: Yeah, that was my question on the call. if we needed it here or not

packages/core-contracts/contracts/proxy/OwnableUUPSImplementation.sol
leomassazza: The UniversalProxyImplementation is storage agnostic. This Ownable (that doesn't need to be owneable according to the other thread) is the one that adds the unstructured storage

packages/core-contracts/contracts/proxy/UnstructuredProxy.sol
i-stam: how are you going to return an address variable  in storage without using a namespace?

PR #320, opened 2021-11-04
Add updater for storage slots

### LINKED ISSUE #268: Verify slot address is the same as the comment (if present) in Namespaces
It would be great if validate-storage subtask can check if the namespace slot address is the one shown in the comments above it. i.e. in the following code we would like to check that `0x8f203f5ee9f9a1d361b4a0f56abfdac49cdd246db58538b151edf87309e955b9` is actually  `bytes32(uint(keccak256("io.synthetix.global")) - 1)`. 
```
    function _globalStorage() internal pure returns (GlobalNamespace storage store) {
        assembly {
            // bytes32(uint(keccak256("io.synthetix.global")) - 1)
            store.slot := 0x8f203f5ee9f9a1d361b4a0f56abfdac49cdd246db58538b151edf87309e955b9
        }
    }
```

mjlescano:
Closes #268

This PR adds a simple script for searching all the storage files that define a hash slot and update it if necessary using the label on the comment.

Usage:
```javascript
npm run update-storage-slots
```

This script will find all the contracts that have something like this defined:
```
        assembly {
            // bytes32(uint(keccak256("io.synthetix.proxy")) - 1)
            store.slot := 0x0000000000000000000000000000000000000000000000000000000000000000
        }
```

And replace it with the correct calculated hash from the label:
```
        assembly {
            // bytes32(uint(keccak256("io.synthetix.proxy")) - 1)
            store.slot := 0x9dbde58b6f7305fccdc5abd7ea1096e84de3f9ee47d83d8c3efc3e5557ac9c74
        }
```

### DIFF
--- a/bin/update-storage-slots.js
+++ b/bin/update-storage-slots.js
@@ -1,5 +1,33 @@
 #!/usr/bin/env node
 
+/**
+
+Helper script for searching all the storage files that define a hash slot and update
+it if necessary using the label on the comment.
+
+Usage:
+  npm run update-storage-slots
+
+This script will find all the contracts that have something like this defined:
+
+```
+        assembly {
+            // bytes32(uint(keccak256("io.synthetix.proxy")) - 1)
+            store.slot := 0x0000000000000000000000000000000000000000000000000000000000000000
+        }
+```
+
+And replace it with the correct calculated hash from the label:
+
+```
+        assembly {
+            // bytes32(uint(keccak256("io.synthetix.proxy")) - 1)
+            store.slot := 0x9dbde58b6f7305fccdc5abd7ea1096e84de3f9ee47d83d8c3efc3e5557ac9c74
+        }
+```
+
+*/
+
 const path = require('path');
 const fs = require('fs/promises');
 const { promisify } = require('util');
--- a/bin/update-storage-slots.js
+++ b/bin/update-storage-slots.js
@@ -46,8 +46,10 @@ function toEip1967Hash(label) {
 
 async function main() {
   const contracts = await glob(
-    path.join(__dirname, '..', '**', 'contracts', 'storage', '**', '*.sol'),
-    { ignore: ['**/node_modules/**', '**/artifacts/**'] }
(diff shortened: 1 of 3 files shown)


bin/update-storage-slots.js
```diff
@@ -0,0 +1,82 @@
...
+
+    if (newHash !== oldHash) {
+      console.log(
+        `Updating label "${label}" on "${path.relative(
+          path.resolve(__dirname, '..'),
+          contractPath
+        )}"`
+      );
+
+      const result = source.replace(oldHash, newHash);
```
i-stam, line 73: My main question is: should it replace it, or just raise an error and let us correct it manually.
I see it more like a validation tool. Correcting the error without us even realising that there was an error may not be the best approach.

mjlescano: Oh, my idea of the flow for using the script is that it replaces everything, and you can check all the changes using git diff. I really like the idea of not having to manually copy paste stuff, which could be prone to errors (maybe you paste it on another storage file, which all kinda lookalike).

As an alternative, I could add a confirmation on the terminal, like `"Would you like to update the storage slot LABEL on FILE ?"`

mjlescano:
Just reconverted it to draft because it doesn't take into account the case were a single file has several slots definitions.

ajsantander (changes requested):

bin/update-storage-slots.js
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
mjlescano: Yes, I like it being on the deployer, but, by AST traversion, you mean to take the entire tree of local contracts starting from the `Modules` ?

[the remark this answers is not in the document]
ajsantander: I think we'd need to traverse ALL contracts, not just modules.

bin/update-storage-slots.js
i-stam: ah ok i got it, yes sounds good, even though if the file ios already modified you wouldn't understand that something was wrong. maybe a warning as you described above would be better then.

PR #340, opened 2021-11-05
TO DELETE - Use Initializer in ERC20

leomassazza:


packages/core-modules/contracts/modules/OwnerModule.sol
```diff
@@ -3,4 +3,7 @@ pragma solidity ^0.8.0;
 
 import "@synthetixio/core-contracts/contracts/ownership/Ownable.sol";
 
-contract OwnerModule is Ownable {}
```
leomassazza, line 6: remove linter warning

ajsantander (approved):

packages/core-contracts/test/proxy/Beacon.test.js
```diff
@@ -14,16 +14,20 @@ describe('Beacon', () => {
...
   describe('when a beacon is deployed', () => {
     before('deploy the beacon and set an initial implementation', async () => {
-      let factory;
+      let factory, tx;
 
       factory = await ethers.getContractFactory('ERC20');
-      ERC20Mock1 = await factory.deploy('Random Token 1', 'rnd1', 18);
-      ERC20Mock2 = await factory.deploy('Random Token 2', 'rnd2', 18);
+      ERC20Mock1 = await factory.deploy();
+      tx = await ERC20Mock1.initialize('Random Token 1', 'rnd1', 18);
```
i-stam, line 21: in theory you can even avoid initializing them! they are just used for testing

packages/core-contracts/contracts/token/ERC20.sol
```diff
@@ -2,18 +2,22 @@
...
 
-    constructor(
+    function initialize(
         string memory tokenName,
         string memory tokenSymbol,
         uint8 tokenDecimals
-    ) {
+    ) public virtual {
         ERC20Namespace storage store = _erc20Storage();
+        if (bytes(store.name).length > 0 || bytes(store.symbol).length > 0 || store.decimals > 0) {
```
i-stam, line 18: again in theory, this check can fail, even though it wouldn't be practical, my point is, should we use a logic like the initializer contract of OZ. Maybe for future implementations

i-stam (approved):
Just left a couple of comments. Not prohibiting for now, but we should keep a note.

PR #341, opened 2021-11-05
Tidy ups on Ownable stuff

ajsantander:
Fix #339

Deleted OwnerModule tests in core-modules because they are not really needed. They don't increase coverage, and the mock is no longer needed.

PR #343, opened 2021-11-05
Add codecov.yml config file to be able to set coverage precision

mjlescano:
This configuration will make codecov stop showing an error if the coverage drops less than `0.1%`, e.g.: https://github.com/Synthetixio/synthetix-v3/pull/341


i-stam (approved):
looks rare!

leomassazza (approved):

PR #345, opened 2021-11-05
Major tidy up on proxies

### LINKED ISSUE #344: Clean up core-contracts/proxy

ajsantander:
Fix #344


packages/core-contracts/contracts/proxy/AbstractProxy.sol
```diff
@@ -1,9 +1,7 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "./ProxyStorage.sol";
-
-contract ForwardingProxy is ProxyStorage {
+abstract contract AbstractProxy {
```
ajsantander, line 4: This contract is now only in charge of forwarding stuff. It does not specify where the implementation is retrieved from.

E.g. in UUPSProxy it gets it from local storage.
In BeaconProxy it gets it from the beacon contract.

packages/core-contracts/contracts/proxy/Beacon.sol
```diff
@@ -1,29 +1,32 @@
...
 import "../utils/ContractUtil.sol";
 
-contract Beacon is IBeacon, BeaconStorage, ContractUtil, CommonErrors {
-    event Upgraded(address indexed implementation);
+contract Beacon is ProxyStorage, ContractUtil, CommonErrors {
+    event Upgraded(address implementation);
 
-    function getImplementation() external view override returns (address) {
-        return _beaconStorage().implementation;
+    constructor(address firstImplementation) {
```
ajsantander, line 11: Constructors are ok (and I think necessary) in proxies. It is the only way to set the first implementation without adding a an upgradeability admin selector to the ABI, which we can't have because of the proxy transparency problem.

packages/core-contracts/contracts/proxy/BeaconProxy.sol
```diff
@@ -1,21 +1,38 @@
...
 import "../common/CommonErrors.sol";
-import "../interfaces/IBeacon.sol";
-import "../ownership/OwnableMixin.sol";
 import "../utils/ContractUtil.sol";
 
-abstract contract BeaconProxy is ForwardingProxy {
-    // solhint-disable-next-line no-empty-blocks
-    function _getBeacon() internal view virtual returns (address) {}
+contract BeaconProxy is AbstractProxy, BeaconStorage, CommonErrors, ContractUtil {
+    constructor(address firstBeacon) {
```
ajsantander, line 11: Constructors are ok (and I think necessary) in proxies. It is the only way to set the first implementation without adding a an upgradeability admin selector to the ABI, which we can't have because of the proxy transparency problem.

packages/core-contracts/contracts/proxy/BeaconStorage.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity ^0.8.0;
 
 contract BeaconStorage {
     struct BeaconNamespace {
```
ajsantander, line 5: This namespace will only be used by beacon proxies, e.g. SyntheticUSD, SyntheticETH, etc.

This is how they know which beacon to ask for their implementation.

packages/core-contracts/contracts/proxy/ProxyStorage.sol
```diff
@@ -4,7 +4,6 @@ pragma solidity ^0.8.0;
 contract ProxyStorage {
     struct ProxyNamespace {
         address implementation;
-        bool simulatingUpgrade;
```
ajsantander, line 7: Removed anti-bricking for now. We can re-introduce when complexity stabilizes. And yes, I really want to try again to get rid of this boolean.

packages/core-contracts/contracts/proxy/UUPSImplementation.sol
```diff
@@ -2,93 +2,29 @@
...
 import "../common/CommonErrors.sol";
 import "./ProxyStorage.sol";
 
-contract UUPSImplementation is ProxyStorage, ContractUtil, TransformUtil, CommonErrors {
+contract UUPSImplementation is ProxyStorage, ContractUtil, CommonErrors {
     error ImplementationIsSterile(address implementation);
 
     event Upgraded(address implementation);
 
-    function _setImplementation(address newImplementation) internal virtual {
```
ajsantander, line 14: All these wrappers don't add anything and just makes things more confusing...

packages/core-contracts/contracts/proxy/UUPSImplementation.sol
```diff
@@ -2,93 +2,29 @@
...
-            revert ImplementationIsSterile(newImplementation);
-        }
 
-        _setImplementation(newImplementation);
+        _proxyStorage().implementation = newImplementation;
 
         emit Upgraded(newImplementation);
     }
 
-    function _canUpgradeInTheFuture(address newImplementation) private returns (bool) {
```
ajsantander, line 50: We will meet again anti-bricking problem!

packages/core-contracts/contracts/proxy/UUPSProxy.sol
```diff
@@ -0,0 +1,19 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./AbstractProxy.sol";
+import "./ProxyStorage.sol";
+
+contract UUPSProxy is AbstractProxy, ProxyStorage {
+    constructor(address firstImplementation) {
```
ajsantander, line 8: An UUPS proxy is just a proxy that gets its implementation from local storage.

packages/core-contracts/contracts/test/proxy/ImplementationMockA.sol
```diff
@@ -1,25 +1,16 @@
...
 
-    function _implementationMockStorage() internal pure returns (ImplementationMockNamespace storage store) {
-        assembly {
-            // bytes32(uint(keccak256("io.synthetix.v3.core-contracts.proxyimplementation")) - 1)
-            store.slot := 0xd2adbc3c9c4f8d59de43697ca626553b287f0ffd89301738d2b2d6ae3172b7f7
-        }
-    }
-}
+contract ImplementationMockA is UUPSImplementation {
+    uint private _a;
```
ajsantander, line 7: I really think its ok to use regular storage in mocks.

packages/core-contracts/test/proxy/Beacon.test.js
```diff
@@ -4,66 +4,54 @@ const assertRevert = require('@synthetixio/core-js/utils/assert-revert');
 const { findEvent } = require('@synthetixio/core-js/utils/events');
 
 describe('Beacon', () => {
-  let Beacon, ERC20Mock1, ERC20Mock2;
+  let Beacon, Implementation;
```
ajsantander, line 7: All we test here is that we can set the beacon implementation.

packages/core-contracts/test/proxy/BeaconProxy.test.js
```diff
@@ -1,55 +1,105 @@
 const { ethers } = hre;
 const assert = require('assert');
+const assertRevert = require('@synthetixio/core-js/utils/assert-revert');
+const { findEvent } = require('@synthetixio/core-js/utils/events');
 
 describe('BeaconProxy', () => {
-  let Beacon, BeaconProxy, MockedBeaconProxy, Implementation;
+  let BeaconProxy, Instance, Beacon, Implementation;
```
ajsantander, line 7: This is almost identical to UUPSProxy.test.js, except that the implementation is set in the beacon.

packages/core-modules/contracts/test/modules/UpgradeModuleMock.sol
```diff
@@ -4,13 +4,7 @@ pragma solidity ^0.8.0;
 import "../../modules/UpgradeModule.sol";
 
 contract UpgradeModuleMock is UpgradeModule {
-    // solhint-disable-next-line private-vars-leading-underscore
-    function __setOwner(address newOwner) public {
```
ajsantander, line 8: Can we use the "mock" prefix instead of these underscores?

packages/core-modules/contracts/modules/UpgradeModule.sol
```diff
@@ -6,10 +6,6 @@ import "@synthetixio/core-contracts/contracts/ownership/OwnableMixin.sol";
 
 contract UpgradeModule is UUPSImplementation, OwnableMixin {
     function upgradeTo(address newImplementation) public override onlyOwner {
```
ajsantander, line 8: All that UpgradeModule does on top of UUPSProxy is add the owner protection.

packages/core-modules/contracts/test/modules/UpgradeModuleMock.sol
```diff
@@ -4,13 +4,7 @@ pragma solidity ^0.8.0;
 import "../../modules/UpgradeModule.sol";
 
 contract UpgradeModuleMock is UpgradeModule {
```
ajsantander, line 6: In an integrated system, we would set the first owner via the OwnerModule.

Since we're testing the UpgradeModule in isolation here, we need to mock a way to set the first owner.

packages/core-modules/test/modules/UpgradeModule.test.js
```diff
@@ -5,50 +5,44 @@ const assertRevert = require('@synthetixio/core-js/utils/assert-revert');
...
+    const tx = await UpgradeModule.mockFirstOwner(owner.address);
+    await tx.wait();
+  });
+
+  it('shows that the implementation is not set', async () => {
+    assert.equal(await UpgradeModule.getImplementation(), '0x0000000000000000000000000000000000000000');
   });
 
-  describe('when a regular user attempts to upgrade the system', () => {
+  describe('when attempting to set the implementation with a non owner signer', () => {
```
ajsantander, line 32: Since all UpgradeModule adds to UUPSProxy is the onlyOwner protection, that's all we test here.

We assume that the rest of the functionality is tested at a lower dependency level.

packages/synthetix-main/test/SNXModule.test.js
```diff
@@ -1,99 +1,99 @@
...
-
-          await SNXUpdated.setValueA(42);
-          valueA = await SNXUpdated.getValueA();
-          assert.equal(valueA, 42);
-        });
-      });
-    });
-  });
-});
+// const hre = require('hardhat');
```
ajsantander, line 1: Disabled for now, otherwise this PR becomes endless...

leomassazza:
Massive changes. Looks fine, but I'll need to do a proper coverage review while is merged and after/while we do the Fix-XXX tasks. 
I also added another ticket to include (if missing after all this refactor, as I think it happened) coverage of payables, covering the four possible cases: Calling with/without data and with/without sending ether in the call.

leomassazza (approved):
Approved, contingent upon the prioritization of the follow up tickets.

packages/core-contracts/contracts/proxy/Beacon.sol
ajsantander: Hmmm or maybe not. Maybe the final rule could be that constructors can only be allowed in proxies, for simplicity of decisions.

ajsantander: Ok, updated. Beacon.sol no longer has a constructor.

mjlescano (approved):

PR #363, opened 2021-11-08
No sensitive functions exposed by default

### LINKED ISSUE #362: No sensitive functions should be public and unprotected by default

ajsantander:
Fix #362

### DIFF
--- a/packages/core-contracts/contracts/proxy/UUPSImplementation.sol
+++ b/packages/core-contracts/contracts/proxy/UUPSImplementation.sol
@@ -8,7 +8,7 @@ import "./ProxyStorage.sol";
 contract UUPSImplementation is ProxyStorage, ContractUtil, CommonErrors {
     event Upgraded(address implementation);
 
-    function upgradeTo(address newImplementation) public virtual {
+    function _upgradeTo(address newImplementation) internal virtual {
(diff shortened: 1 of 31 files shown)


packages/core-contracts/contracts/proxy/UUPSImplementation.sol
```diff
@@ -8,7 +8,7 @@ import "./ProxyStorage.sol";
 contract UUPSImplementation is ProxyStorage, ContractUtil, CommonErrors {
     event Upgraded(address implementation);
 
-    function upgradeTo(address newImplementation) public virtual {
```
leomassazza, line 11: I see a problem here thinking ahead, upgradeTo should exist (at least as virtual or in an interface) to enable again the anti-bricking protection (since we need to simulate an upgrade from the interface)

leomassazza (approved):
Approved, with a comment to consider

packages/core-contracts/contracts/proxy/UUPSImplementation.sol
ajsantander: Yeah, until then its not needed. Following the practice of only adding things when needed.

i-stam (approved):
look great, thanks!

PR #383, opened 2021-11-10
Use constructor in Beacon.sol

### LINKED ISSUE #382: Add constructor to Beacon and reimagine beacon storage to be used by both Beacon and BeaconProxy

i-stam:
FIx #382

- Uses the same storage namespace for both `Beacon` and `BeaconProxy`.
- Adds testing functionality for `_setBeacon()`.
- Uses `IBeacon` interface


packages/core-contracts/contracts/proxy/BeaconProxy.sol
```diff
@@ -2,18 +2,22 @@
...
 
 import "./AbstractProxy.sol";
-import "./Beacon.sol";
 import "./BeaconStorage.sol";
 import "../common/CommonErrors.sol";
+import "../interfaces/IBeacon.sol";
 import "../utils/ContractUtil.sol";
 
 contract BeaconProxy is AbstractProxy, BeaconStorage, CommonErrors, ContractUtil {
+    error NoChange();
```
ajsantander, line 11: Would this be better in CommonErrors?

ajsantander (approved):

mjlescano (approved):

packages/core-contracts/contracts/proxy/BeaconProxy.sol
i-stam: yeap..good point. I was wondering about the same. I left it there till we decide if we want to use a perhaps better term

PR #389, opened 2021-11-11
Adds a CoreOwnerModule

### LINKED ISSUE #377: Add CoreOwnerModule even if it doesn't add any functionality beyond Ownable for the time being?

ajsantander:
Fix #377


leomassazza (approved):

packages/synthetix-main/contracts/modules/OwnerModule.sol
```diff
@@ -1,9 +1,9 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "@synthetixio/core-contracts/contracts/ownership/Ownable.sol";
+import "@synthetixio/core-modules/contracts/modules/CoreOwnerModule.sol";
 
 // solhint-disable-next-line no-empty-blocks
-contract OwnerModule is Ownable {
+contract OwnerModule is CoreOwnerModule {
```
i-stam, line 7: why do we need to define `CoreOwnerModule` instead of having `OwnerModule is Ownable`

ajsantander: That's the whole point of this PR.

CoreUpgradeModule inherits UUPSImplementation and adds owner protection.
CoreOwnerModule inherits Ownable, but does nothing atm.

So, in projects, we inherit CoreUpgradeModule, but instead of inheriting a CoreOwnerModule, we "build" one in situ by inheriting Ownable.

My reasoning is that it is simpler to follow the same pattern for both modules. AND, in the future, we may add functionality to OwnerModule, and we'd have to change it in a single place.

i-stam (approved):

PR #392, opened 2021-11-11
Make CommsMixin a core-module contract

### LINKED ISSUE #381: Make CommsMixin a core-module contract?

leomassazza:


packages/core-modules/contracts/test/RouterMock.sol
```diff
@@ -0,0 +1,50 @@
+//SPDX-License-Identifier: MIT
```
ajsantander, line 1: Shouldn't this be in the 'mocks' folder. I thought we'd renamed it.

leomassazza: Yes we did... I had a previous version of the folders :( Moved

packages/core-modules/test/contracts/modules/CoreCommsMixin.test.js
```diff
@@ -0,0 +1,86 @@
...
+const assert = require('assert');
+const assertRevert = require('@synthetixio/core-js/utils/assert-revert');
+const bn = require('@synthetixio/core-js/utils/assert-bignumber');
+
+const { ethers } = hre;
+
+describe('CoreCommsMixin', () => {
+  let SomeModuleMockImp, AnotherModuleMockImp, Router;
+  const WRONG_VALUE = 13;
+  // let owner, user;
```
ajsantander, line 11: Remove commented out code?

packages/core-modules/test/contracts/modules/CoreCommsMixin.test.js
```diff
@@ -0,0 +1,86 @@
...
+    factory = await ethers.getContractFactory('SomeModuleMock');
+    SomeModuleMockImp = await factory.deploy();
+    factory = await ethers.getContractFactory('AnotherModuleMock');
+    AnotherModuleMockImp = await factory.deploy();
+    factory = await ethers.getContractFactory('RouterMock');
+    Router = await factory.deploy(SomeModuleMockImp.address, AnotherModuleMockImp.address);
+  });
+
+  before('setup the mock router', async () => {
+    const factory = await ethers.getContractFactory('RouterMock');
```
ajsantander, line 28: Wasn't the Router deployed in L23-24?

ajsantander (changes requested):
@leomassazza I kind of see what you're doing here, but I think its too complex in an unnecessary way. This is very hard to reason about.

IMO a much simpler test should do:
* Mock CoreCommsMixin with a constructor that sets the proxy implementation, and a function that calls `_intermoduleCall(...)`
* The implementation is just a RouterMock that has some function we wanna test.
* The function in RouterMock simply records the received that and the msg.sender. All we do is test if both are the expected values.

The above test tests the code that is written. So it should be good enough in my opinion.

PR #394, opened 2021-11-11
Fix mocks coverage

### LINKED ISSUE #388: Remove mock contracts from coverage

mjlescano:
Closes #388 

This main objective of this PR is to add tests for all the used mocks, so we don't deteriorate our coverage just for some Mocked contracts. From now on, we should strive to have 100% coverage on Mock contracts also, regardless they are meant for production or just as test utils.


leomassazza (approved):

packages/core-contracts/test/contracts/mocks/proxy/ImplementationMockB.test.js
```diff
@@ -0,0 +1,23 @@
+const { ethers } = hre;
+const assert = require('assert');
+
+describe('ImplementationMockB', () => {
```
ajsantander, line 4: Hmmm, I don't think we should test them directly should we?

What I had in mind is that we hit every line of code in the mocks when we use them for testing other stuff. I.e. hit the setA method in ImplementationMockB in the UUPSProxy tests.

ajsantander (approved):

PR #397, opened 2021-11-12
Add tests for bricking the proxy

### LINKED ISSUE #396: Add tests (failing and thus skipped for now) for bricking the proxy, and destroying its implementation of a UUPS proxy

ajsantander:
Fix #396

### DIFF
--- a/packages/core-contracts/test/contracts/proxy/UUPSProxy.test.js
+++ b/packages/core-contracts/test/contracts/proxy/UUPSProxy.test.js
@@ -159,5 +159,32 @@ describe('UUPSProxy', () => {
         });
       });
     });
+
+    // TODO: Unskip tests here when the anti-bricking mechanism is put back in place.
+    describe('when trying to brick the proxy', () => {
+      describe('when trying to upgrade to a sterile implementation', () => {
+        before('deploy the implementation', async () => {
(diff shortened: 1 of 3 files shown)


packages/core-contracts/test/contracts/proxy/UUPSProxy.test.js
```diff
@@ -159,5 +159,32 @@ describe('UUPSProxy', () => {
...
+        it.skip('reverts', async () => {
+          await assertRevert(Instance.upgradeTo(Implementation.address), 'SterileImplementation');
+        });
+      });
+
+      describe('when trying to destroy the implementaion', () => {
+        let Destroyer;
+
+        before('deploy the malicious implementation', async () => {
+          const factory = await ethers.getContractFactory('ImplementationDestroyer');
```
leomassazza, line 180: np: use the same (any) convention for naming implementations (ImplementationDestroyer vs SterileImplementation)

ajsantander: Yeah, I think that the convention is to put Implementation at the end of anything that's supposed to be behind a proxy.

In this case, its not that this is not following that convention, its just not trying to. So IMO naming is good for now...

i-stam (approved):

PR #408, opened 2021-11-16
Add proxy config

### LINKED ISSUE #306: Require the user to define a Proxy contract on hardhat deployer config

mjlescano:
Closes #306


packages/deployer/internal/config-defaults.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
+  proxyContract: 'Proxy',
   paths: {
-    modules: 'contracts/modules',
     deployments: 'deployments',
+    modules: 'modules',
```
mjlescano, line 5: To follow common practices with other Hardhat plugins, now the `modules` folder is configured from the `sources` folder, not the root.

packages/deployer/extensions/config.js
```diff
@@ -3,21 +3,19 @@ const { extendConfig } = require('hardhat/config');
...
       ...configDefaults.paths,
-      modules: resolvePath('modules'),
-      deployments: resolvePath('deployments'),
+      ...(givenConfig?.paths || {}),
     },
   };
+
+  // Resolve the absolute path from the root of the configurable path
+  config.deployer.paths.deployments = path.resolve(root, config.deployer.paths.deployments);
+  config.deployer.paths.modules = path.resolve(sources, config.deployer.paths.modules);
```
mjlescano, line 20: Updated this absolute paths calculations to be more explicit, the old approach was somewhat difficult to understand.

packages/core-modules/.solcover.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  skipFiles: ['contracts/Proxy.sol', 'contracts/Router.sol'],
+  skipFiles: ['contracts/Router.sol'],
```
mjlescano, line 2: We don't need to skip the Proxy because its empty, so it doesn't add uncovered lines to the coverage report.

packages/deployer/test/fixture-projects/custom-proxy/test/helpers/initializer.js
```diff
@@ -0,0 +1,28 @@
+const { TASK_DEPLOY } = require('@synthetixio/deployer/task-names');
+
+/**
+ * Initializer helper for the sample-project, allows to deploy and initialize
+ * the project on the given hardhat environment
+ * @param {import('hardhat/types').HardhatRuntimeEnvironment}
+ */
+module.exports = function createInitializer(hre) {
```
ajsantander, line 8: I'm afraid you'll need to adapt this to the new way of initializing systems in tests that was just merged.
It's very easy tho. You can find some examples here: https://github.com/Synthetixio/synthetix-v3/pull/418/files

ajsantander (approved):
Looks good @mjlescano !
Left a comment about something that needs to be fixed, and also, do you think this PR could also include the custom proxy name for the synthetix-main project?

leomassazza (approved):

i-stam (approved):

PR #421, opened 2021-11-19
Unify naming on SNXToken and Synths Modules

### LINKED ISSUE #417: Use common naming in SNXTokenModule and SynthsModule
Both modules does almost the same but the naming is different: SynthsModule use deployXXX and XXXDeployed while SNXTokenModule uses createXXX and XXXCreated

leomassazza:
Fix #417


ajsantander (approved):

i-stam (approved):
Ah went for the create prefix! I am a fan of democracy so...approved!

PR #427, opened 2021-11-19
Refactor deployer tests initializers

### LINKED ISSUE #420: Test Review: Simplify deployer fixture projects tooling / setup

mjlescano:
Closes #420


packages/deployer/test/src/tasks/deploy.test.js
```diff
@@ -1,13 +1,18 @@
...
 const { ContractValidationError } = require('../../../internal/errors');
 const { useEnvironment } = require('../../helpers');
 
 describe('tasks/deploy.js', function () {
   describe('when deploying a correctly written project', function () {
     useEnvironment('sample-project');
 
+    beforeEach('initialize sample-project owner', async function () {
+      await sampleProjectInitializer(this.deploymentInfo, this.hre);
+    });
```
mjlescano, line 14: I removed all the Initializer logic from the helper, and let the specific test take care of this because we were only needing it here. And doing it this way we remove a lot of logic from the helper.

packages/deployer/test/helpers.js
```diff
@@ -67,22 +60,6 @@ function _getEnvironmentPath(fixtureProjectName) {
...
-    if (err.code === 'MODULE_NOT_FOUND') {
-      throw new Error(
-        `Sample project didn't define any tests environment initializer helper: ${initializerPath}`
-      );
-    }
-
-    throw err;
-  }
-}
-
```
mjlescano, line 85: I'm removing all the initialization logic from here, as we are only needing it once on a single test, so, there's no need to try to load it always for now.

packages/deployer/test/src/tasks/deploy.test.js
```diff
@@ -17,7 +18,7 @@ describe('tasks/deploy.js', function () {
         clear: true,
       });
 
-      await this.initSystem();
+      await sampleProjectInitializer(this.deploymentInfo, this.hre);
```
mjlescano, line 21: Now, we directly initialize `sample-project` after deploy here.

packages/deployer/test/helpers.js
leomassazza: It's only need once up to now, but this is going to be used on almost all integration tests for the end user projects since they most likely will require an initialized system (i.e. owned)

leomassazza (changes requested):
TBH I think removing the injected initializer is not a good idea. Might be only used in a single place right now, but won't be the case when we build the real products

packages/deployer/test/fixture-projects/sample-project/test/helpers/initializer.js
```diff
@@ -1,13 +1,13 @@
 const { getProxyAddress } = require('@synthetixio/deployer/utils/deployments');
 
-module.exports = async function initializer(deploymentInfo) {
+module.exports = async function initializer(deploymentInfo, _hre = hre) {
```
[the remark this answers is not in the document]
mjlescano: Yes, because on the deployer tests we use fixture projects, meaning that each fixture project has its own folder and on each test we have to re-load the HRE object on the given fixture project folder (we're doing that [here](https://github.com/Synthetixio/synthetix-v3/blob/refactor-fixtures-initializers/packages/deployer/test/helpers.js#L17:L17)). I used [this boilerplate](https://github.com/nomiclabs/hardhat-ts-plugin-boilerplate/) as reference.

So, because we are loading the HRE for each fixture project and not using `hardhat  test` to run them, we do not have the global object injected at the beginning, and also, after loading it for the first time, everytime we re-load another environment the global reference to `hre` changes because we reset it. So, maybe we are keeping a reference to an old one.

packages/deployer/test/src/tasks/deploy.test.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const { rejects } = require('assert/strict');
 const { copyFile, unlink, readFile, writeFile } = require('fs/promises');
+const sampleProjectInitializer = require('../../fixture-projects/sample-project/test/helpers/initializer');
```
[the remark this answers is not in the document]
mjlescano: Yes! I think this is the best solution between re-adding the old code (like @leomassazza commented) and leaving  it as it is now (having to manually call the initializer).

packages/deployer/test/fixture-projects/sample-project/test/helpers/initializer.js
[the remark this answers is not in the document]
ajsantander: Ok cool, makes sense!

ajsantander (approved):

PR #429, opened 2021-11-19
Fix assert strict usage

### LINKED ISSUE #426: Use `assert/strict` on all tests instead of `assert`
This is so we have more predictable assertions and better error diffs, as noted on the docs: https://nodejs.org/api/assert.html

mjlescano:
Closes #426


leomassazza (approved):
💯

ajsantander (approved):
GM!

Just note that some more tests need `assertBn` to pass.

PR #463, opened 2021-11-23
Fix quiet option when compiling contracts

### LINKED ISSUE #311: Update forcing of hardhat output only on required function calls
Right now, if the `quiet` option is enabled, we are completely disabling `console.log` ([here](https://github.com/Synthetixio/synthetix-v3/blob/main/packages/deployer/tasks/deploy.js#L78:L94)) on the entire process and making it unavailable for the entire process.

Instead of doing that, we could just disable it during compilation, e.g.:
```javascript
// DISABLE console.log HERE
await hre.run(TASK_COMPILE, { force: true, quiet: true });
// RE-ENABLE console.log HERE
```
 
 Btw, check if hardhat fixed their quiet parameter

mjlescano:
Closes #311

### DIFF
--- a/packages/deployer/tasks/deploy.js
+++ b/packages/deployer/tasks/deploy.js
@@ -72,7 +72,7 @@ task(TASK_DEPLOY, 'Deploys all system modules')
 /*
  * Note: Even though hardhat's compile task has a quiet option,
  * it stil prints some output. This is a hack to completely silence
- * output during deployment.
+ * output during compile task run.
  */
 async function _compile(hre, quiet) {
   let logCache;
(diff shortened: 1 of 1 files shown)


packages/deployer/tasks/deploy.js
```diff
@@ -70,25 +66,25 @@ task(TASK_DEPLOY, 'Deploys all system modules')
...
       throw err;
-    } finally {
-      _forceSilenceHardhat(false, quiet);
     }
   });
 
 /*
  * Note: Even though hardhat's compile task has a quiet option,
  * it stil prints some output. This is a hack to completely silence
  * output during deployment.
```
mjlescano, line 75:
```suggestion
 * output during compile task run.
```

leomassazza (approved):

ajsantander (approved):

PR #473, opened 2021-11-24
Fix proxy storage validation

### LINKED ISSUE #428: On second deployment, the Proxy's storage is marked as deleted only because the  Proxy is not re-deployed
This is because we use only deployments file to compare ASTs ([here](https://github.com/Synthetixio/synthetix-v3/blob/main/packages/deployer/internal/storage-ast-validator.js#L70:L80)), and if we don't modify or delete we don't add the new source to the new JSONs.

After fixing, uncomment test in `deploy.test.js`

mjlescano:
Closes #428

This PR fixes a bug where we were validating Storage usage on the Proxy (comparing it with the previous deployment) before compiling, which caused to throw an error because the deployer thought that it was deleted.

So, the solution implemented here breaks up the `upgrade-proxy` subtask into `sync-proxy`, and the latter is executed before running any validations, so we make sure that all the `Proxy.sol` data is loaded on the current deployments json files.

### DIFF
--- a/packages/deployer/subtasks/sync-proxy.js
+++ b/packages/deployer/subtasks/sync-proxy.js
@@ -1,0 +1,38 @@
+const { subtask } = require('hardhat/config');
+const logger = require('@synthetixio/core-js/utils/logger');
+const { initContractData } = require('../internal/process-contracts');
+const { ContractValidationError } = require('../internal/errors');
+const { SUBTASK_SYNC_PROXY } = require('../task-names');
+
+/**
+ * Compiles the Proxy's source, saves it to the deployments file, and checks
+ * that it does not have changes if it's already deployed.
+ */
+subtask(SUBTASK_SYNC_PROXY, 'Compile and sync the source from the Proxy.').setAction(
+  async (_, hre) => {
+    logger.subtitle('Syncing and compiling source from the Proxy');
+
+    const proxyName = hre.config.deployer.proxyContract;
+
+    await initContractData(proxyName);
+
+    const currentBytecode =
+      hre.deployer.deployment.general.contracts[proxyName].deployedBytecodeHash;
+    const previousBytecode =
+      hre.deployer.previousDeployment?.general.contracts[proxyName]?.deployedBytecodeHash;
+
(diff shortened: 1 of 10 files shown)


packages/deployer/subtasks/sync-proxy.js
```diff
@@ -0,0 +1,38 @@
...
+ * Compiles the Proxy's source, saves it to the deployments file, and checks
+ * that it does not have changes if it's already deployed.
+ */
+subtask(SUBTASK_SYNC_PROXY, 'Compile and sync the source from the Proxy.').setAction(
+  async (_, hre) => {
+    logger.subtitle('Syncing and compiling source from the Proxy');
+
+    const proxyName = hre.config.deployer.proxyContract;
+
+    await initContractData(proxyName);
```
mjlescano, line 17: This is the main part of the fix, which initializes all the Proxy's contract data before running any validation.

packages/deployer/subtasks/sync-proxy.js
```diff
@@ -0,0 +1,38 @@
...
+      throw new ContractValidationError(
+        'Proxy contract cannot be deleted or have its name changed'
+      );
+    }
+
+    if (previousBytecode && previousBytecode !== currentBytecode) {
+      throw new ContractValidationError(
+        `The ${proxyName} contract cannot be changed after first deployment`
+      );
+    }
```
mjlescano, line 34: Took the chance to add some simple validations to make sure that we are not deleting nor editing the already deployed proxy, which is the only Contract that we cannot re-deploy (because it would change its address).

packages/deployer/subtasks/upgrade-proxy.js
```diff
@@ -63,14 +62,6 @@ function _getDeployedAddress(contractName, hre) {
   return hre.deployer.deployment.general.contracts[contractName].deployedAddress;
 }
 
-async function _deployProxy({ proxyName, routerAddress, hre }) {
-  await initContractData(proxyName);
```
mjlescano, line 67: Move initialization to `sync-proxy` subtask.

leomassazza (approved):

ajsantander (approved):

PR #475, opened 2021-11-24
Bootstrap CLI package

### LINKED ISSUE #455: Bootstrap the CLI project as a hardhat plugin with a placeholder "interact" task.
Use both the v3 deployer package, and the v2x interact task as reference.

In the v2x task, we can see how we build the CLI UI, interact with the contracts etc. Keep in mind that this code is a simple hacky script, and we want things to be more robust in v3.

In the v3 deployer task, we can see how to subdivide a task in subtasks to keep code simple and robust, how to bootstrap the plugin, how to set up sample projects, etc.

In this issue, we only need to bootstrap the project with a dummy task.

Like in the deployer plugin, add a test/fixture-projects/sample-project folder with a sample router compatible project that will be used to test the plugin, both manually and programmatically through mocha.

The folder for now should just be a vanilla hardhat project that loads the deployer plugin, with a few contracts with dummy code to interact with.

No need to set up anything for tests just yet. We will be adding automated tests later. We can manually test the cli for now by importing it as a plugin in `synthetix-main`.

i-stam:
FIX #455


packages/synthetix-cli/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": ""
+  },
+  "author": "",
+  "license": "MIT",
+  "devDependencies": {
+    "@nomiclabs/hardhat-ethers": "^2.0.2",
+    "ethers": "^5.5.1",
+    "hardhat": "^2.6.8"
```
leomassazza, line 14: I'd suggest using 2.6.6 as we have on the other packages and maybe move all together in a separate PR

packages/synthetix-cli/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": ""
+  },
+  "author": "",
+  "license": "MIT",
+  "devDependencies": {
+    "@nomiclabs/hardhat-ethers": "^2.0.2",
+    "ethers": "^5.5.1",
```
leomassazza, line 13: ethers is included in @nomiclabs/hardhat-ethers

leomassazza (approved):
Approved with some np

packages/synthetix-cli/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
ajsantander, line 1: I think we can name this package _just_ cli, so that the fully qualified package name is @synthetixio/cli

ajsantander: Also, this git ignore might be redundant (see deployer package's .gitignore). The workspace ignore might be enough.

packages/synthetix-cli/package.json
ajsantander: I think it's alright, since he's following the pattern used in the deployer package. If you do apply Leo's suggestion, I think it should be on both plugins.

packages/synthetix-cli/tasks/interact.js
```diff
@@ -0,0 +1,18 @@
+const { task } = require('hardhat/config');
+
+const { TASK_INTERACT } = require('../task-names');
+
+task(TASK_INTERACT, 'Interacts with a given Synthetix deployment')
```
ajsantander, line 5: Description should read "Interacts with a given modular system deployment" (it's not just Synthetix now)

Also, I think it's safe to remove all params for now. Most of them don't apply anymore in the v3 architecture, and we should probably add them only as they are actually used.

ajsantander (commented):
Looks good, the only critical change I want to ask is to rename the package.

mjlescano (changes requested):
Please, add the `.npmrc` file as the other packages, and pin the dependencies versions on the `package.json` (e.g. change `"^1.0.1"` to `"1.0.1"`), so we are sure to not add any unwanted vulnerabilities on automatic upgrades.

packages/synthetix-cli/.gitignore
i-stam: fixed

ajsantander (approved):

mjlescano (approved):

PR #478, opened 2021-11-25
Fix nondeterministic router compile

### LINKED ISSUE #431: Deployer: (bug) re-ran on kovan without contract changes and it appears to have performed actions and generated a diff in artifact files
To reproduce:
* cd synthetix-main
* npx hardhat deploy --network kovan

The deployer deploys a new router, and it shouldn't, since there are no contract or module changes. It then asks to upgrade the main proxy with the new router implementation.

UPDATE: It looks like the deployer is getting different local and remote bytecode hashes for the router, even though the sources do not change, and thus interprets that it needs to be deployed.

ajsantander:
Fix #431

The bug was caused because of different commits in the Router's comment block. The assumption was that comments did not affect runtime bytecode, but this is not true in Solidity. Solidity appends the metadata hash (which is affected by comments) to the _runtime_ bytecode of a contract.


packages/core-modules/hardhat.config.js
```diff
@@ -3,7 +3,15 @@ require('@nomiclabs/hardhat-ethers');
 require('@synthetixio/deployer');
 
 module.exports = {
-  solidity: '0.8.4',
+  solidity: {
```
ajsantander, line 6: Making solc compiler settings more explicit to avoid bytecode differences in the future.

mjlescano (approved):

PR #484, opened 2021-11-29
add slither analysis github action

artdgn:
~~## This PR is blocked until https://github.com/crytic/slither/issues/919 is resolved (slither doesn't still support custom errors).~~ No longer blocked after https://github.com/crytic/slither/releases/tag/0.8.2

Tasks:
- [ ] Ensure `slither-local` command in `synthetix-main` works (should print analysis results in terminal).
- [ ] Ensure `slither-sarif` command in `synthetix-main` works (should create a `sarif` format file).
- [ ] Ensure the sarif file is picked up in PRs and pushes and the analysis is shown in PRs diffs and "checks" tabs.
- [ ] Add slither analysis for `core-contracts` & `core-modules` and not only `synthetix-main` sub packages.
- [ ] Look into the released github-action https://github.com/marketplace/actions/slither-action


artdgn:
Updated the PR with latest `main`.

It seems that slither is having difficulties:
- when running in `core-contracts` it errors out with an error similar to https://github.com/crytic/slither/issues/1001 and https://github.com/crytic/slither/issues/776. Probably related to IR conversion of a library method.
- when run in `synthetix-main` it can't find the contracts imported from `core-contracts` for some reason. 

I guess worth to wait for resolutions of the errors on slither side.

dbeal-eth:
I believe we have decided we don't want slither.

PR #496, opened 2021-11-29
Load deployment and print header in INTERACT task

### LINKED ISSUE #467: Subtask: print-header
We should define the parameter interface of the main task, and only add for now the parameters that we intend to support in the MVP.

For reference, these are v2's current parameters:
 --block-tag           Specify the block tag to interact at, per ethers.js specification
  --deployment-path     Specify the path to the deployment data directory
  --gas-limit           Max gas to use when signing transactions (default: 8000000)
  --private-key         Private key to use to sign txs
  --provider-url        The http provider to use for communicating with the blockchain
  --target-network      Target the instance deployed in this network (default: "mainnet")
  --use-fork            Use a local fork
  --use-ovm             Use an Optimism chain

For now just take the parameters, and print some of them in the welcome header, like in the v2x interact task.

Consider using the core-js highlight box features used in the deployer logger.

The location of this code should be similar to how its done in the deployer plugin. Use a subtask if needed.

ajsantander:
Closes #467 

Implements a header, similar to the deployer's header, in the cli's interact task
Screen Shot 2021-11-29 at 10 47 29

In fact, this implementation does not duplicate code. The header has been made part of the logger util in the core-js package.

UPDATE: After consideration, the figlet package was removed. If we do want the graffiti fonts in the future, we can just hardcode figlet output. There is no need to have such a big dependency (with fonts) in core-js.

Another non-trivial change made is that a new task has been added to the deployer: load-deployment.js. This plugin, as other may in the future, calls it to have deployment data available at runtime via hre.deployer.deployment.*


packages/cli/subtasks/load-deployment.js
```diff
@@ -0,0 +1,41 @@
+const fs = require('fs');
```
ajsantander, line 1: This could have been done in a separate PR. In fact, we should've made an issue for it. But, if it's not too chaotic, we can include it here.

Basically, what this does is fetch the target deployment data and inject it at runtime. We could make this something that the deployer plugin can do, for other plugins to use, and not have every plugin implement it.

packages/core-js/test/utils/hardhat/argument-types.test.js
```diff
@@ -1,7 +1,7 @@
 const { equal, notEqual, throws, doesNotThrow } = require('assert/strict');
-const { alphanumeric } = require('../../../internal/argument-types');
+const { alphanumeric } = require('../../../utils/hardhat/argument-types');
```
ajsantander, line 2: Both the cli and the deployer plugins use this, so I moved it to core-js

packages/core-js/utils/logger.js
```diff
@@ -37,6 +39,11 @@ module.exports = {
     this._log(`${this.prepend}${msg}${completeStr}${this.postpend}`);
   },
 
+  async title(msg, font = 'small') {
```
ajsantander, line 42: The logger now prints the figlet title.

I did this because I didn't want to duplicate code in the CLI and the DEPLOYER's titles.

packages/deployer/task-names.js
```diff
@@ -1,18 +1,18 @@
...
-  SUBTASK_GENERATE_ROUTER_SOURCE: 'generate-router-source',
-  SUBTASK_PREPARE_DEPLOYMENT: 'prepare-deployment',
-  SUBTASK_PRINT_INFO: 'print-info',
-  SUBTASK_SYNC_PROXY: 'sync-proxy',
-  SUBTASK_SYNC_SOURCES: 'sync-sources',
-  SUBTASK_UPGRADE_PROXY: 'upgrade-proxy',
-  SUBTASK_VALIDATE_MODULES: 'validate-modules',
-  SUBTASK_VALIDATE_STORAGE: 'validate-storage',
-  SUBTASK_VALIDATE_ROUTER: 'validate-router',
+  SUBTASK_CANCEL_DEPLOYMENT: 'deployer-cancel-deployment',
```
ajsantander, line 2: Had to prepend task names with the plugin, to avoid collisions. Looks like hardhat mixes all plugin subtasks in the same environment.

packages/cli/subtasks/load-deployment.js
ajsantander: Actually yeah, I moved this subtask to the deployer.

Even though the deployer does not use it, other plugins (campatible with the deployer) can use it.

packages/cli/tasks/interact.js
```diff
@@ -1,5 +1,23 @@
...
+const types = require('@synthetixio/core-js/utils/hardhat/argument-types');
+const logger = require('@synthetixio/core-js/utils/logger');
+
+task(TASK_INTERACT, 'Interacts with a given modular system deployment')
+  .addOptionalParam(
+    'instance',
+    'The name of the target instance for deployment',
+    'official',
+    types.alphanumeric
+  )
```
mjlescano, line 15:
```suggestion
  )
  .addFlag('debug', 'Display debug logs', false)
```

i-stam (approved):
this is useful! let's merge it after linting are fixed

packages/deployer/package.json
```diff
@@ -18,7 +18,6 @@
   "dependencies": {
     "chalk": "4.1.2",
     "del": "6.0.0",
-    "figlet": "1.5.2",
```
mjlescano, line 21: ❤️

packages/deployer/subtasks/load-deployment.js
```diff
@@ -0,0 +1,41 @@
...
+    hre.deployer.paths.sources = sources;
+    hre.deployer.paths.abis = abis;
+
+    hre.deployer.deployment = {
+      general: currentFile,
+      sources,
+      abis,
+    };
+  }
+);
```
mjlescano, line 41: Could we somehow use this same subtask on `prepare-deployment` to avoid some duplicate code?

packages/deployer/package.json
ajsantander: You win 😄

packages/deployer/subtasks/load-deployment.js
ajsantander: I thought about it, but my conclusion was that they are quite different. One is read only. Making the same code do both would make it a bit complicated unnecessarily?

PR #503, opened 2021-11-29
Cli pick-contract subtask

### LINKED ISSUE #468: Subtask: pick-contract
As soon as the CLI starts, and after it prints out the environment information, it reaches out to the deployment JSON files of the target instance and identifies all system modules.

It prints a list of the modules with:
ModuleName - Address

User can use the keyboard arrows to navigate the list.

As the user types a module name, the list is filtered.

(See how this is implemented in the v2x code).

This code should also reside in a subtask file.

i-stam:
FIX #468 
Run: 
cd packages/synthetix-main
npx hardhat interact --instance test


packages/cli/subtasks/pick-contract.js
```diff
@@ -0,0 +1,43 @@
...
+const inquirer = require('inquirer');
+const autocomplete = require('inquirer-list-search-prompt');
+const logger = require('@synthetixio/core-js/utils/logger');
+const { SUBTASK_PICK_CONTRACT } = require('../task-names');
+
+subtask(SUBTASK_PICK_CONTRACT, 'Pick contract to interact with').setAction(
+  async (taskArguments, hre) => {
+    logger.info('Pick a contract:');
+
+    const deploymentData = JSON.parse(fs.readFileSync(hre.deployer.paths.deployment));
```
ajsantander, line 12: No need to parse the data, it's already parsed at `hre.deployer.deployment.[general|abis|sources]`

packages/cli/subtasks/pick-contract.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs');
+const { subtask } = require('hardhat/config');
+const inquirer = require('inquirer');
+const autocomplete = require('inquirer-list-search-prompt');
+const logger = require('@synthetixio/core-js/utils/logger');
+const { SUBTASK_PICK_CONTRACT } = require('../task-names');
+
+subtask(SUBTASK_PICK_CONTRACT, 'Pick contract to interact with').setAction(
+  async (taskArguments, hre) => {
+    logger.info('Pick a contract:');
```
ajsantander, line 10: No need for this, since the step title is already printed in the prompt.

packages/cli/subtasks/pick-contract.js
```diff
@@ -0,0 +1,43 @@
...
+);
+
+function _prioritizeTarget(contracts, itemName) {
+  contracts.splice(contracts.indexOf(itemName), 1);
+  contracts.unshift(itemName);
+}
+
+async function _prompt(contracts) {
+  // Set up inquirer
+  inquirer.registerPrompt('autocomplete', autocomplete);
```
ajsantander, line 28: This is awesome!

packages/cli/subtasks/pick-contract.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+);
+
+function _prioritizeTarget(contracts, itemName) {
+  contracts.splice(contracts.indexOf(itemName), 1);
+  contracts.unshift(itemName);
+}
+
+async function _prompt(contracts) {
+  // Set up inquirer
```
ajsantander, line 27: Maybe make this comment more informative? I.e. "Set up inquirer with plugin to filter results as you type (autocomplete)"

ajsantander (approved):
Looks great!

I think we need one more thing before we can merge. Take a look at `packages/deployer/extensions/environment.js`. This is a little "deployer" object that we inject at runtime to the hre, that we use to share state between the deployer's subtasks.

Notice how extensions are loaded in index.js, and how the runtime object is frozen in environment.js lines 22 and 23.

I believe that the cli package could use the same technique for its state. The pick contract task could inject the chosen contract to the state, so that the next subtask (pick-function) can take over.

packages/cli/subtasks/pick-contract.js
```diff
@@ -0,0 +1,43 @@
...
+    await _prompt(contracts);
+  }
+);
+
+function _prioritizeTarget(contracts, itemName) {
+  contracts.splice(contracts.indexOf(itemName), 1);
+  contracts.unshift(itemName);
+}
+
+async function _prompt(contracts) {
```
leomassazza, line 26: This looks useful enough to be generalized and put in the js-core. Maybe a future PR.

packages/cli/task-names.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  SUBTASK_PICK_CONTRACT: 'cli-pick-contract',
```
leomassazza, line 2: super np. Put all subtask together (ordering)

leomassazza (approved):
Approved with some minor comments

packages/cli/subtasks/pick-contract.js
i-stam: yeap, i thought about this...makes sense

ajsantander: Yeah, we'll probably move it to an internal util in the next PR

PR #504, opened 2021-11-29
Make-NFT-Ownable

### LINKED ISSUE #457: Extend NFT to make it non-transferable by holder, only by owner

leomassazza:
Fix #457

### DIFF
--- a/packages/core-contracts/contracts/token/ERC721Owned.sol
+++ b/packages/core-contracts/contracts/token/ERC721Owned.sol
@@ -30,8 +30,8 @@ contract ERC721Owned is ERC721, Ownable {
         super.safeTransferFrom(from, to, tokenId, data);
     }
 
-    function _isApprovedOrOwner(address, uint256) internal view virtual override onlyOwner returns (bool) {
+    function _isApprovedOrOwner(address, uint256) internal view virtual override returns (bool) {
         // The owner (and only the owner) is authorized to transfer
-        return true;
+        return (this.owner() == msg.sender);
     }
 }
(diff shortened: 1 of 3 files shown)


packages/core-contracts/contracts/token/ERC721Owned.sol
```diff
@@ -0,0 +1,37 @@
...
+    function safeTransferFrom(
+        address from,
+        address to,
+        uint256 tokenId,
+        bytes memory data
+    ) public virtual override onlyOwner {
+        super.safeTransferFrom(from, to, tokenId, data);
+    }
+
+    function _isApprovedOrOwner(address, uint256) internal view virtual override onlyOwner returns (bool) {
```
ajsantander, line 33: Hmmm, not sure about this one. It's weird to have a view function with an access modifier. I'd prefer if it had no modifier and returned `address == owner`.

ajsantander (approved):
Awesome 😎 

Just a little comment on code style on one of the utility view functions.

packages/core-contracts/contracts/token/ERC721Owned.sol
i-stam: agreed, it can be used inside of a modifier instead fi we want this check

i-stam (approved):

packages/core-contracts/contracts/token/ERC721Owned.sol
leomassazza: Comment addressed. Please confirm this is better (@ajsantander ). Thanks

PR #505, opened 2021-11-30
Confirm on deploy subtask

### LINKED ISSUE #491: When not using --no-confirm, if the router changes, the user is not prompted to deploy it, it just deploys it.

ajsantander:
Fix #491

### DIFF
--- a/packages/deployer/subtasks/deploy-contracts.js
+++ b/packages/deployer/subtasks/deploy-contracts.js
@@ -21,7 +21,7 @@ subtask(
   for (const contractName of [...toUpdate, ...toCreate]) {
     await hre.run(SUBTASK_DEPLOY_CONTRACT, {
       contractName,
-      requireConfirmation: false,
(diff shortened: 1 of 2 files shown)


packages/deployer/subtasks/deploy-contracts.js
```diff
@@ -21,6 +21,7 @@ subtask(
   for (const contractName of [...toUpdate, ...toCreate]) {
     await hre.run(SUBTASK_DEPLOY_CONTRACT, {
       contractName,
+      requireConfirmation: true, // This could be set to false. For now, double confirmation seems kinda nice.
```
leomassazza, line 24: can we make it dependent on the quiet or some deployer parameter?

leomassazza: forget it... It's already checked :/

leomassazza (approved):

i-stam (approved):

PR #510, opened 2021-11-30
Adds interfaces for all public facing contracts

### LINKED ISSUE #486: Easy to use interface access (useful for subgraphs)

ajsantander:
Fix #486

This PR should cover all the interfaces needed so far in the repo.

Note that not all contracts have interfaces, but only those supposed to be outward facing, i.e. that we can expect other protocols to reach into, as well as subgraphs.


leomassazza (approved):

i-stam (commented):
It looks good, but are adding all the external functions there or just on a per demand basis. +I thought that we wanted the events/errors to be inside the contracts for convenience and readability.

mjlescano (approved):

i-stam (approved):

PR #511, opened 2021-11-30
Add election nominees

### LINKED ISSUE #451: Ability to self nominate in ElectionModule

mjlescano:
Closes #451


packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -1,6 +1,7 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
```
ajsantander, line 4: We want to use our own core-contracts library instead of OZ. You could argue that this case could be an exception, which would make sense, but my case is that that would make it hard for us to draw the line further down the road. So my suggestion is to only use OZ as reference.

ajsantander (changes requested):
See my comment about not importing OZ

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Agree.. let's bring the library that we need (and just what we use)

packages/core-contracts/contracts/utils/AddressSet.sol
```diff
@@ -0,0 +1,70 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+library AddressSet {
+    struct Values {
```
ajsantander, line 5: Can we have a better naming for the type?

i.e. what happens when we have a UintSet. The type can't also be called Values.

packages/core-contracts/contracts/utils/AddressSet.sol
```diff
@@ -0,0 +1,70 @@
...
+        address[] values;
+        mapping(address => uint256) indexes;
+    }
+
+    function add(Values storage set, address value) internal returns (bool) {
+        if (!contains(set, value)) {
+            set.values.push(value);
```
ajsantander, line 12: nit: Breather line?

packages/core-contracts/contracts/utils/AddressSet.sol
```diff
@@ -0,0 +1,70 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+library AddressSet {
```
ajsantander, line 4: Any idea on what will our strategy be when we need to support other types? Will we duplicate code?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -40,4 +41,16 @@ contract CoreElectionModule is ElectionStorage, OwnableMixin {
     function getMemberTokenAddress() public view returns (address) {
         return _electionStore().memberTokenAddress;
     }
+
+    function getNominees() external view returns (address[] memory) {
+        return AddressSet.values(_electionStore().nominees);
+    }
+
+    function selfNominate() external {
+        AddressSet.add(_electionStore().nominees, msg.sender);
```
ajsantander, line 50: Add zero address validation?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -40,4 +41,16 @@ contract CoreElectionModule is ElectionStorage, OwnableMixin {
...
+    }
+
+    function selfUnnominate() external {
+        AddressSet.remove(_electionStore().nominees, msg.sender);
```
ajsantander, line 54: Do we want to revert if msg.sender is not in the set?

Atm it fails silently.

packages/core-contracts/contracts/utils/AddressSet.sol
mjlescano: We could duplicate code, or just bring the entire util from OZ's code which is compatible with multiple types: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol

I didn't do that last option just because we were needing it at the moment :)

ajsantander: Yeah, A generic enumerable set might not be very gas efficient, haven't checked tho.

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: Yup, done!

packages/core-contracts/contracts/utils/AddressSet.sol
mjlescano: Not adding the Set for now

packages/core-contracts/contracts/utils/AddressSet.sol
mjlescano: For now, I'm going to remove the Set entirely and refactor it when needing it.

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: done!

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -3,11 +3,19 @@ pragma solidity ^0.8.0;
 
 import "@synthetixio/core-contracts/contracts/ownership/OwnableMixin.sol";
 import "@synthetixio/core-contracts/contracts/proxy/UUPSProxy.sol";
+import "@synthetixio/core-contracts/contracts/errors/AddressError.sol";
 import "../interfaces/IElectionModule.sol";
 import "../token/MemberToken.sol";
 import "../storage/ElectionStorage.sol";
 
 contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
+    error MemberTokenAlreadyCreated();
```
ajsantander, line 12: You should only declare errors and events in one place; either the interface or the implementation.

A very recent decision was made to do so in the implementation.

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -4,11 +4,20 @@ pragma solidity ^0.8.0;
...
 
     function upgradeMemberTokenImplementation(address newMemberTokenImplementation) external;
 
     function getMemberTokenAddress() external view returns (address);
+
+    function getNominees() external view returns (address[] memory);
+
+    function selfNominate() external;
+
+    function selfUnnominate() external;
```
ajsantander, line 17: Andy recently suggested a better name for these functions:
nominate()
withdrawNomination()

packages/core-modules/contracts/storage/ElectionStorage.sol
```diff
@@ -4,6 +4,8 @@ pragma solidity ^0.8.0;
 contract ElectionStorage {
     struct ElectionStore {
         address memberTokenAddress;
+        address[] nominees;
+        mapping(address => uint256) nomineesIndexes;
```
ajsantander, line 8: I think the proper grammar is `nomineeIndexes`

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -37,4 +43,58 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        }
+
+        store.nominees.push(msg.sender);
+        store.nomineesIndexes[msg.sender] = store.nominees.length;
+    }
+
+    function selfUnnominate() external override {
+        ElectionStore storage store = _electionStore();
+
+        uint256 valueIndex = store.nomineesIndexes[msg.sender];
```
ajsantander, line 69: Do we really need information about the index of the nominee?
Can't we use a boolean dictionary to just know if the nominee is in the array? That way all this removal logic becomes unnecessary.

ajsantander (changes requested):
Sorry Mati, still requesting changes 😬

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: Hmm, not sure how would that work without having to potentially iterate the entire array on removals, which could be a gas limited operation. Also, we need to keep the array up to date so we can return it on `getNominees()`.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -37,4 +45,58 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        store.nominees.push(msg.sender);
+        store.nomineeIndexes[msg.sender] = store.nominees.length;
+    }
+
+    function withdrawNomination() external override {
+        ElectionStore storage store = _electionStore();
+
+        uint256 valueIndex = store.nomineeIndexes[msg.sender];
+
+        if (msg.sender == address(0)) {
```
ajsantander, line 73: Move this above the index read?

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander: Shit yeah. I brainfarted. Please ignore!

ajsantander (approved):

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -37,4 +45,58 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
     function getMemberTokenAddress() public view override returns (address) {
         return _electionStore().memberTokenAddress;
     }
+
+    function getNominees() external view override returns (address[] memory) {
+        return _electionStore().nominees;
+    }
+
+    function nominate() external override {
```
leomassazza, line 53: @ajsantander is there any limit in the size of an address array that can be retrieved? 
line 49 will retrieve the whole array. Shall we limit the size here?

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: this condition can be removed (unreachable path)

packages/core-modules/test/contracts/modules/CoreElectionModule.test.js
```diff
@@ -96,4 +96,30 @@ describe('CoreElectionModule', () => {
       });
     });
   });
+
+  describe('when the user self nominates', () => {
+    it('sets the value & unsets the value', async () => {
```
leomassazza, line 101: np it can be splitted in different describes to identify where it fails if it fails

leomassazza (approved):
Approved. Left some comments for your consideration

packages/core-modules/test/contracts/modules/CoreElectionModule.test.js
mjlescano: Done!

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: Done!

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: @mjlescano confirmed. getNominees() will revert with out of gas if the address[] is too large.

PR #513, opened 2021-11-30
Convert utils to libraries

### LINKED ISSUE #509: Use libraries (without use syntax) for utils

ajsantander:
Fix #509

Before this PR, utilities like `isContract` were used by extending a contract and calling internal functions like: `_isContract(...)`.

This PR makes utils libraries, so you can do `AddressUtil.isContract(...)`. It also splits some utils.

Note that this PR intentionally does not implement Solidity's `use` keyword. I believe that it is more explicit and readable to use `AddressUtil.isContract(value)` instead of `value.isContract()`.


packages/deployer/test/unit/internal/process-transactions.test.js
```diff
@@ -69,9 +69,9 @@ describe('internal/process-transactions.js', function () {
       });
 
       it('registers a transaction in the deployment data', async function () {
-        const transaction = hre.deployer.deployment.general.transactions[transaction.hash];
```
ajsantander, line 72: Unrelated fix. Not sure how this got to main 😬

leomassazza (approved):

i-stam (approved):

PR #523, opened 2021-12-01
Cli pick-function subtask 

### LINKED ISSUE #469: Subtask: pick-function
With the same UI that we use to print the contract list, print the list of functions of the current contract that can be called.

This also supports filtering by typing the function name.

i-stam:
Fix #469


ajsantander (changes requested):
Can we try to find a way to achieve this without nesting functions?

packages/cli/subtasks/pick-function.js
```diff
@@ -0,0 +1,60 @@
...
+      }
+      return false;
+    });
+
+    if (query === '') {
+      abiMatches.splice(0, 0, escItem);
+    }
+    resolve(abiMatches);
+  });
+}
```
mjlescano, line 60: Looks like the promise/async interface is not necessary here:

```suggestion
function _searchAbi(abi, matches, query = '') {
  const abiMatches = abi.filter((item) => {
    if (item.name && item.type === 'function') {
      return item.name.toLowerCase().includes(query.toLowerCase());
    }

    return false;
  });

  if (query === '') {
    abiMatches.splice(0, 0, escItem);
  }

  return abiMatches;
}
```

(also,  replaced `let abiMatches` for `const abiMatches`)

packages/cli/subtasks/pick-function.js
```diff
@@ -0,0 +1,60 @@
...
+async function _prompt(abi) {
+  // Set up inquirer with autocomplete plugin i.e. as you type filtering
+  inquirer.registerPrompt('autocomplete', autocomplete);
+
+  const prompt = await inquirer.prompt([
+    {
+      type: 'autocomplete',
+      name: 'contractFunction',
+      message: 'Pick a FUNCTION:',
+      source: (matches, query) => _searchAbi(abi, matches, query),
```
mjlescano, line 33: Is this `source:` option valid? it is not documented here: https://www.npmjs.com/package/inquirer#prompt 🤔

mjlescano (approved):

packages/cli/extensions/environment.js
```diff
@@ -12,5 +12,4 @@ extendEnvironment((hre) => {
   // Prevent any properties being added to hre.cli
   // other than those defined above.
   Object.preventExtensions(hre.cli);
```
ajsantander, line 14: If we're not using injected objects, we should consider removing the whole extension thing.

packages/cli/subtasks/pick-contract.js
```diff
@@ -7,7 +7,7 @@ subtask(SUBTASK_PICK_CONTRACT, 'Pick contract to interact with').setAction(
   async (taskArguments, hre) => {
     const contracts = Object.keys(hre.deployer.deployment.general.contracts);
 
-    _prioritizeTarget(contracts, 'Synthetix');
+    _prioritizeTarget(contracts, 'SNXTokenModule');
```
ajsantander, line 10: Should probably remove this, since we want this CLI to be non-Synthetix specific.

packages/cli/subtasks/pick-function.js
```diff
@@ -0,0 +1,45 @@
...
+const { SUBTASK_PICK_FUNCTION } = require('../task-names');
+
+subtask(SUBTASK_PICK_FUNCTION, 'Pick a function from the given contract').setAction(
+  async (taskArguments, hre) => {
+    const abi = hre.deployer.deployment.abis[hre.cli.contract];
+    return await _prompt(abi);
+  }
+);
+
+async function _prompt(abi) {
```
ajsantander, line 14: Review if this can be abstracted and re-used between the `pick-contract` and `pick-function` subtasks.

packages/cli/tasks/interact.js
```diff
@@ -18,8 +23,15 @@ task(TASK_INTERACT, 'Interacts with a given modular system deployment')
     const { debug } = taskArguments;
 
     logger.debugging = debug;
+    const escItem = '↩ BACK';
```
ajsantander, line 26: If this is used by subtasks, we should probably abstract it to a constants file.

packages/cli/tasks/interact.js
```diff
@@ -18,8 +23,15 @@ task(TASK_INTERACT, 'Interacts with a given modular system deployment')
...
 
     await hre.run(SUBTASK_LOAD_DEPLOYMENT, { ...taskArguments, readOnly: true });
     await hre.run(SUBTASK_PRINT_INFO, taskArguments);
-    await hre.run(SUBTASK_PICK_CONTRACT, taskArguments);
+
+    let contractFunction = escItem;
```
ajsantander, line 31: Can we use null or undefined as the identifier here instead of escItem?

packages/cli/extensions/environment.js
ajsantander: Removed

packages/cli/subtasks/pick-contract.js
ajsantander: Removed

packages/cli/subtasks/pick-function.js
ajsantander: Implemented

packages/cli/subtasks/pick-function.js
```diff
@@ -0,0 +1,45 @@
...
+  // Set up inquirer with autocomplete plugin i.e. as you type filtering
+  inquirer.registerPrompt('autocomplete', autocomplete);
```
ajsantander, line 16: Can this be loaded only once at the task level?

packages/cli/tasks/interact.js
```diff
@@ -18,8 +23,15 @@ task(TASK_INTERACT, 'Interacts with a given modular system deployment')
...
+    while (contractFunction === escItem) {
```
ajsantander, line 32: This loop sort of works but it needs to be polished a little.

packages/cli/subtasks/pick-function.js
ajsantander: Done

packages/cli/subtasks/pick-function.js
ajsantander: Done

packages/cli/tasks/interact.js
ajsantander: Done

packages/cli/tasks/interact.js
ajsantander: Fixed

ajsantander (approved):

packages/cli/subtasks/pick-function.js
```diff
@@ -0,0 +1,32 @@
...
+const constants = require('../constants');
+const { SUBTASK_PICK_FUNCTION } = require('../task-names');
+const inquirer = require('inquirer');
+
+subtask(SUBTASK_PICK_FUNCTION, 'Pick a function from the given contract').setAction(
+  async (taskArguments, hre) => {
+    const abi = hre.deployer.deployment.abis[hre.cli.contractName];
+    const abiFunctions = abi.filter((abiItem) => abiItem.name && abiItem.type === 'function');
+
+    abiFunctions.splice(0, 0, { name: constants.BACK });
```
mjlescano, line 11: Does this `.splice` work correctly? Being the value to remove `{ name: constants.BACK }` I think that it will compare 2 different instances of the same object and not going to remove it.

May, you can do this directly:
```suggestion
    const abiFunctions = abi
      .filter((abiItem) => abiItem.name && abiItem.type === 'function');
      .filter((abiItem) => abiItem.name !== constants.BACK);
```

ajsantander: Delete count is zero, I'm actually trying to inject the element, not remove it.

PR #524, opened 2021-12-02
Move events and errors out of interfaces

### LINKED ISSUE #519: Consider not using errors and events in interfaces

ajsantander:
Fix #519

As recently stated by frob.eth in Twitter:

"The function definitions are the on-chain inputs and outputs. The events are the off-chain outputs.

Solidity interfaces are used for integrating smart contracts between themselves, and those smart contracts can't see the events."

According to this, not only errors don't have a place in interfaces, but events don't either.

This PR moves all errors and events out of Solidity interfaces.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -581,7 +581,6 @@
     },
     "node_modules/@ethersproject/abi": {
       "version": "5.5.0",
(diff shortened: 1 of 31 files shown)


i-stam (approved):
I prefer it this way!

leomassazza (approved):
That way looks better.

PR #527, opened 2021-12-02
[PoC] Add basic council elections

### LINKED ISSUE #459: Basic voting functionality

mjlescano:
Closes #459 

This PR adds the first base for implementing elections on top. At this point, this is not a working governance module.

### DIFF
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -8,6 +8,10 @@ interface IElectionModule {
 
     function getMemberTokenAddress() external view returns (address);
 
+    function setElectionTokenAddress(address addr) external;
+
+    function getElectionTokenAddress() external view returns (address);
+
     function getNominees() external view returns (address[] memory);
 
     function nominate() external;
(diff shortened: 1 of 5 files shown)


packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -13,4 +13,18 @@ interface IElectionModule {
     function nominate() external;
 
     function withdrawNomination() external;
+
+    function setSeatCount(uint seats) external;
+
+    function setEpochDuration(uint duration) external;
+
+    function setPeriodPercent(uint8 percent) external;
```
mjlescano, line 21: Maybe create a single method for configuring an epoch? So we don't have 3 separate methods removing some noise:

```solidity
function configureCurrentEpoch(uint seats, uint duration, uint nominatePerdiodPercent) external;
```

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -90,4 +93,44 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         // Delete the index for the deleted slot
         delete store.nomineeIndexes[msg.sender];
     }
+
+    function setSeatCount(uint seats) external override onlyOwner {
+        _electionStore().seatCount = seats;
+    }
```
mjlescano, line 99: Shouldn't we block to change this value after an epoch has started?

leomassazza (approved):

ajsantander (approved):

PR #529, opened 2021-12-03
Add pick parameters subtask for the cli

### LINKED ISSUE #470: Subtask: populate-params
Arguably the hardest subtask of the plugin, since it needs to be able to parse user input into types known by ethers/solidity.

As in the v2x task, it should prompt the user to populate one parameter at a time. Note that this is poorly achieved in the v2x implementation, since things start getting shady once complex parameters like arrays appear.

ajsantander:
Fix #470


leomassazza (approved):

packages/cli/subtasks/pick-parameters.js
```diff
@@ -0,0 +1,41 @@
...
+        {
+          type: 'input',
+          name: 'userInput',
+          message: `  ${parameter.name} (${parameter.type}):`,
+        },
+      ]);
+
+      let encodedParameter;
+      try {
+        // Encode and decode to parse user input.
```
ajsantander, line 29: Could use a better comment here

packages/cli/subtasks/pick-parameters.js
```diff
@@ -0,0 +1,42 @@
...
+        encodedParameter = abiCoder.decode([parameter.type], encodedParameter);
+
+        hre.cli.functionParameters.push(encodedParameter);
+
+        parameterIndex++;
+      } catch (error) {
+        logger.warn(error);
+      }
+    }
+  }
```
mjlescano, line 41:
```suggestion
    for (const parameter of functionAbi.inputs) {
      const { userInput } = await inquirer.prompt([
        {
          type: 'input',
          name: 'userInput',
          message: `  ${parameter.name} (${parameter.type}):`,
        },
      ]);

      let encodedParameter;
      try {
        // Encode and decode the user's input to parse the input
        // into types acceptable by ethers.
        encodedParameter = abiCoder.encode([parameter.type], [userInput]);
        encodedParameter = abiCoder.decode([parameter.type], encodedParameter);

        hre.cli.functionParameters.push(encodedParameter);
      } catch (error) {
        logger.warn(error);
      }
    }
  }
```

mjlescano (approved):

packages/cli/subtasks/pick-parameters.js
ajsantander: This does not retry parameters 😉

mjlescano: Ooh, gotcha :)

PR #530, opened 2021-12-03
Cli preview call

### LINKED ISSUE #528: Subtask: preview-call

ajsantander:
Fix #528 

Also prints addresses next to contract names list, and displays full signatures for functions.


packages/cli/internal/signatures.js
```diff
@@ -0,0 +1,25 @@
+function getSignatureWithParameterNamesAndValues(contractName, functionName, functionParameters) {
```
ajsantander, line 1: Using this utility to present a function with its signature (and optionally values) instead of just the function name.

packages/cli/subtasks/pick-contract.js
```diff
@@ -1,28 +1,39 @@
...
+            .filter((contractName) => {
+              if (query) {
+                return contractName.toLowerCase().includes(query.toLowerCase());
+              }
 
-            return contractName;
-          });
+              return true;
+            })
+            .map((contractName) => {
```
ajsantander, line 24: Show contract address next to contract name

packages/cli/internal/signatures.js
leomassazza: what a name! :)

leomassazza (approved):

packages/cli/internal/signatures.js
ajsantander: Lol, well, it is what the function does ^^

PR #538, opened 2021-12-03
Adds cli execute-call subtask

### LINKED ISSUE #471: Subtask: call-function
Once the execution state has information about the contract, function, and parameters of the interaction, execute the call. If just a read, parse and print the result. If its a write, then process the transaction showing tx hash etc.

Things to consider:
- Confirm gas limit / price on every call?

ajsantander:
Fix #471


packages/cli/extensions/environment.js
```diff
@@ -9,7 +9,6 @@ extendEnvironment((hre) => {
     contractName: null,
     functionName: null,
     functionParameters: null,
-    callConfirmed: null,
```
ajsantander, line 12: Getting rid of this runtime env variable since its not really needed.

packages/cli/subtasks/pick-parameters.js
```diff
@@ -31,7 +31,7 @@ subtask(SUBTASK_PICK_PARAMETERS, 'Populate the selected function parameters').se
         encodedParameter = abiCoder.encode([parameter.type], [userInput]);
         encodedParameter = abiCoder.decode([parameter.type], encodedParameter);
 
-        hre.cli.functionParameters.push(encodedParameter);
+        hre.cli.functionParameters.push(...encodedParameter);
```
ajsantander, line 34: Decode always returns an array, so we need to destrucutre it here.

packages/cli/subtasks/preview-call.js
```diff
@@ -1,25 +1,14 @@
 const { subtask } = require('hardhat/config');
 const { SUBTASK_PREVIEW_CALL } = require('../task-names');
 const logger = require('@synthetixio/core-js/utils/logger');
-const prompter = require('@synthetixio/core-js/utils/prompter');
```
ajsantander, line 4: Bunch of functionality moved to the `execute-call` subtask because it makes more sense there.

This subtask will be kinda skinny for now, but I think we will add stuff to it like gas price, gas limit, sender balance, etc.

packages/core-js/test/utils/logger.test.js
```diff
@@ -61,7 +61,7 @@ describe('utils/logger.js', () => {
           '┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓',
         ],
         [
-          '┃ ‣ hello........................................................................................... ┃',
```
ajsantander, line 64: Removed these lines in the info box (also affects the deployer) because it doesn't allow you to alt-click to select an address, which is annoying.

leomassazza (approved):

PR #541, opened 2021-12-03
Governance Epoch

### LINKED ISSUE #535: Governance: Implement calculation of current and next epoch times

leomassazza:
Fix #535

### DIFF
--- a/packages/core-js/utils/rpc.js
+++ b/packages/core-js/utils/rpc.js
@@ -12,8 +12,8 @@ async function restoreSnapshot(snapshotId, provider) {
   await mineBlock(provider);
 }
 
-async function timeWarp(ellapsedTime, provider) {
-  await provider.send('evm_increaseTime', [ellapsedTime]);
+async function fastForward(seconds, provider) {
+  await provider.send('evm_increaseTime', [seconds]);
 
   await mineBlock(provider);
 }
@@ -25,5 +25,5 @@ async function mineBlock(provider) {
 module.exports = {
   takeSnapshot,
   restoreSnapshot,
-  timeWarp,
+  fastForward,
 };
--- a/packages/core-js/utils/rpc.js
+++ b/packages/core-js/utils/rpc.js
@@ -12,6 +12,12 @@ async function restoreSnapshot(snapshotId, provider) {
   await mineBlock(provider);
 }
 
+async function timeWarp(ellapsedTime, provider) {
+  await provider.send('evm_increaseTime', [ellapsedTime]);
+
+  await mineBlock(provider);
+}
+
 async function mineBlock(provider) {
   await provider.send('evm_mine');
 }
@@ -19,4 +25,5 @@ async function mineBlock(provider) {
 module.exports = {
   takeSnapshot,
   restoreSnapshot,
+  timeWarp,
 };
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -20,8 +20,6 @@ interface IElectionModule {
 
     function setSeatCount(uint seats) external;
 
-    function setEpochDuration(uint64 duration) external;
-
     function setPeriodPercent(uint8 percent) external;
 
     function setNextSeatCount(uint seats) external;
@@ -32,11 +30,11 @@ interface IElectionModule {
 
     function elect(address[] memory candidates) external;
 
-    function epochFinised() external view returns (bool);
+    function isEpochFinished() external view returns (bool);
 
-    function isEpochNominationPeriod() external view returns (bool);
+    function isNomination() external view returns (bool);
 
-    function isEpochVotingPeriod() external view returns (bool);
+    function isVoting() external view returns (bool);
 
-    function initialize() external;
+    function setupFirstEpoch() external;
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -20,15 +20,23 @@ interface IElectionModule {
 
     function setSeatCount(uint seats) external;
 
-    function setEpochDuration(uint duration) external;
+    function setEpochDuration(uint64 duration) external;
 
     function setPeriodPercent(uint8 percent) external;
 
     function setNextSeatCount(uint seats) external;
 
-    function setNextEpochDuration(uint duration) external;
+    function setNextEpochDuration(uint64 duration) external;
 
     function setNextPeriodPercent(uint8 percent) external;
 
     function elect(address[] memory candidates) external;
+
+    function epochFinised() external view returns (bool);
+
+    function isEpochNominationPeriod() external view returns (bool);
+
+    function isEpochVotingPeriod() external view returns (bool);
+
+    function initialize() external;
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -32,7 +32,7 @@ interface IElectionModule {
 
     function isEpochFinished() external view returns (bool);
 
-    function isNomination() external view returns (bool);
+    function isNominating() external view returns (bool);
 
     function isVoting() external view returns (bool);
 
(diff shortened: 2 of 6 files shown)


packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -20,15 +20,23 @@ interface IElectionModule {
 
     function setSeatCount(uint seats) external;
 
-    function setEpochDuration(uint duration) external;
+    function setEpochDuration(uint64 duration) external;
```
leomassazza, line 23: changed to uint64 to prevent overflows when operating with it

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -20,15 +20,23 @@ interface IElectionModule {
...
 
     function elect(address[] memory candidates) external;
+
+    function epochFinised() external view returns (bool);
+
+    function isEpochNominationPeriod() external view returns (bool);
+
+    function isEpochVotingPeriod() external view returns (bool);
+
+    function initialize() external;
```
leomassazza, line 41: I don't know if will be needed. Potentially yes, but at least needed for testing and we need to cover the Epoch 0 case.

mjlescano (approved):

packages/core-js/utils/rpc.js
```diff
@@ -12,11 +12,18 @@ async function restoreSnapshot(snapshotId, provider) {
   await mineBlock(provider);
 }
 
+async function timeWarp(ellapsedTime, provider) {
```
ajsantander, line 15: For consistency with what people are used to, could we rename this to `fastForward(seconds, provider)`?

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: Ok, but if it is needed, this name won't do because of the router proxy's requirement of unique selectors. It would probably need to be named `initializeElectionModule` or something.

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: Does this mean that the duration of the current epoch can be changed on the run? Won't this complicate logic a lot and introduce fragility?

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -20,15 +20,23 @@ interface IElectionModule {
...
     function setSeatCount(uint seats) external;
 
-    function setEpochDuration(uint duration) external;
+    function setEpochDuration(uint64 duration) external;
 
     function setPeriodPercent(uint8 percent) external;
 
     function setNextSeatCount(uint seats) external;
 
-    function setNextEpochDuration(uint duration) external;
```
ajsantander, line 29: Nit: Stick to a single naming pattern? If you look at the functions in this interface, some adhere to one pattern and some to the other.
Either `setNext<name>` or `setNextEpoch<name>`.

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -20,15 +20,23 @@ interface IElectionModule {
...
+    function setNextEpochDuration(uint64 duration) external;
 
     function setNextPeriodPercent(uint8 percent) external;
 
     function elect(address[] memory candidates) external;
+
+    function epochFinised() external view returns (bool);
```
ajsantander, line 35: Multiple things:
* Typo "Finished", with h
* Should this be called `isEpochFinished`?

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -20,15 +20,23 @@ interface IElectionModule {
...
+
+    function isEpochNominationPeriod() external view returns (bool);
```
ajsantander, line 37: These two functions `isEpochNominationPeriod` and `isEpochVotingPeriod` could have better naming.
Do you mean `isOnEpochNominationPeriod` and `isOnEpochVotingPeriod`?

If so, I'd recommend `isNominating` and `isVoting`.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -155,15 +157,62 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         // TODO: Mark msg.sender as already voted on electionVotes
     }
 
-    function _isNextEpochNomination() internal view virtual returns (bool) {
-        return false;
+    function epochFinised() external view override returns (bool) {
+        if (_electionStore().epochStart == 0) {
```
ajsantander, line 161: I'd avoid reverting in view functions as much as possible.  (Also applies in other parts of the file)

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -155,15 +157,62 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            revert EpochNotStarted();
+        }
+
+        return _epochFinished();
```
ajsantander, line 165: Is it really necessary to wrap this in an internal function? (Also applies in other parts of the file)

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -155,15 +157,62 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        }
+
+        return _isEpochVotingPeriod();
+    }
+
+    function initialize() external override {
+        // TODO set epoch 0 seat to 1
+        // TODO set epoch 0 period
+        // TODO set token for owner
+        _changeEpoch();
```
ajsantander, line 188: OZ syndrome?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -155,15 +157,62 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+    }
+
+    function _changeEpoch() internal virtual {
```
ajsantander, line 191: Maybe `_flipEpoch` sounds cooler?

ajsantander (changes requested):
Looking good! Suggesting some changes, and clarifications.

packages/core-modules/contracts/interfaces/IElectionModule.sol
leomassazza: Yes, here I was respecting the original spec. But it makes sense to keep it fixed

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: originally (and still not sure) I thought the changeEpoch can be called from more than one place... but yes, up to now is only once, so agree it looks OZed

packages/core-modules/contracts/interfaces/IElectionModule.sol
leomassazza: It is consistent. We are setting/updating the NextEpochDuration. If we use here, i.e. setNextDuration, we can don't know which duration. Also this adhere with the naming on the specs. https://www.notion.so/GOV-v3-Spec-a045502acd93481182541441e4cf5ddd

packages/core-modules/contracts/interfaces/IElectionModule.sol
leomassazza: Changed for something that makes more sense in the naming, since is not initializing, but setting up the Epoch Zero (initial epoch)

ajsantander: Are we protecting overwrites in these new setup functions?

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: Gotcha!

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: I see you renamed one of the functions to `isNomination`. Is this a typo? My suggestion was `isNominating`.

ajsantander (approved):

packages/core-modules/contracts/interfaces/IElectionModule.sol
leomassazza: fixed

PR #543, opened 2021-12-06
Add elections history

### LINKED ISSUE #536: Governance: Wrap up logic for assigning votes on election

mjlescano:
Closes #536

### DIFF
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -30,7 +30,7 @@ interface IElectionModule {
 
     function elect(address[] memory candidates) external;
 
-    function getNomineeVotes(address candidate) external view returns (uint);
+    function getNomineeVotes(address nominee) external view returns (uint);
 
     function isEpochFinished() external view returns (bool);
 
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -30,6 +30,8 @@ interface IElectionModule {
 
     function elect(address[] memory candidates) external;
 
+    function getNomineeVotes(address candidate) external view returns (uint);
+
     function isEpochFinished() external view returns (bool);
 
     function isNominating() external view returns (bool);
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -183,8 +183,8 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         electionData.addressVoted[msg.sender] = true;
     }
 
-    function getNomineeVotes(address candidate) external view override returns (uint) {
-        return _electionStore().electionData.nomineeVotes[candidate];
+    function getNomineeVotes(address nominee) external view override returns (uint) {
+        return _electionStore().electionData.nomineeVotes[nominee];
     }
 
     function isEpochFinished() public view override returns (bool) {
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -16,6 +16,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
     error InvalidPeriodPercent();
     error InvalidCandidatesCount();
     error InvalidCandidatesRepeat();
+    error FirstEpochAlreadySetUp();
 
     error AlreadyVoted();
 
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -141,7 +141,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         ElectionStore storage store = _electionStore();
         ElectionData storage electionData = store.electionData;
 
-        if (candidates.length == 0) {
+        if (candidates.length == 0 || candidates.length > store.seatCount) {
             revert InvalidCandidatesCount();
         }
 
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -61,7 +61,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         return _electionStore().electionTokenAddress;
     }
 
-    // TODO add pagination for getting nominees list
+    // TODO: add pagination for getting nominees list
     function getNominees() external view override returns (address[] memory) {
         return _electionStore().electionData.nominees;
     }
@@ -150,13 +150,11 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
             revert AlreadyVoted();
         }
 
-        // The voting power is the amount of votes we are going to assign to the given candidates
         uint votePower = ERC20(store.electionTokenAddress).balanceOf(msg.sender);
 
         for (uint i = 0; i < candidates.length; i++) {
             address candidate = candidates[i];
 
-            // Check that the candidate is a nominee
             if (electionData.nomineeIndexes[candidate] == 0) {
                 revert InvalidCandidate(candidate);
             }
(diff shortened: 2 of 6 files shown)


packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -133,37 +139,46 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
-    function elect(address[] memory candidates) external view override {
-        uint seatCount = _electionStore().nextSeatCount;
+    function elect(address[] memory candidates) external override {
+        ElectionStore storage store = _electionStore();
+        ElectionData storage electionData = store.electionData;
+
+        uint seatCount = store.nextSeatCount;
 
-        if (candidates.length != seatCount) {
+        if (candidates.length == 0 || candidates.length > seatCount) {
```
leomassazza, line 148: We don't need to limit it to the seats count. We can include (if needed) a max voting length as another parameter, but we shouldn't limit it here

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -133,37 +139,46 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        // The voting power is the amount of votes we are going to assign to the given candidates
+        uint votePower = ERC20(store.electionTokenAddress).balanceOf(msg.sender);
 
-        // TODO: Recalculate top [seatsCount] nominees
-        //   _minimunIdx;
-        //   _minimumValue;
-        //   if votes > minimum
-        //     electionTopNominees[last][minumunNomineeTop] = msg.sender
-        //     for () // get new minimum (idx & value)
+        for (uint i = 0; i < candidates.length; i++) {
```
leomassazza, line 160: we should store (somewhere) the original list of candidates voted in the ballot

mjlescano: If instead of assigning the same amount of voting power to all the candidates we divide it equally, we could remove this check because the repeated candidates wouldn't affect the final result.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -133,37 +139,46 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            address candidate = candidates[i];
 
-        // TODO: Mark msg.sender as already voted on electionVotes
-    }
+            // Check that all the values on the candidates Array are unique
+            if (i < candidates.length - 1) {
```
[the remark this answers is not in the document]
leomassazza: we should avoid repetitions (that will also have a weird consequence if we use ordered ballots (single transferable votes)

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: But, if we remove the max amount of candidates to vote, I think we should implement also the distribution of the voting power to all the candidates (equally, at the begining).

e.g., if a whale votes the "first" 100 candidates, any address with less voting power than that will almost be left out, because it would get the minimum bar for affecting the votes higher.

mjlescano: As we just talk on DM, adding hard limit by amount of nominees.

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: I will store the list of candidates, vote author, and voting power assigned.

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: Didn't add a check for nominees count, just started checking that the candidate being voted is actually a nominee, and with that and the repeated addresses check, we have a hard limit.

leomassazza (approved):

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -30,6 +30,8 @@ interface IElectionModule {
 
     function elect(address[] memory candidates) external;
 
+    function getNomineeVotes(address candidate) external view returns (uint);
```
ajsantander, line 33: nit: rename parameter to `nominee`?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -57,50 +62,50 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
     }
 
     function getNominees() external view override returns (address[] memory) {
```
ajsantander, line 64: Add TODO for pagination?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -131,26 +136,55 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
 
-        // TODO: if msg.sender already voted, rollback previous votes.
+        if (electionData.addressVoted[msg.sender]) {
+            // TODO: if msg.sender already voted, rollback previous votes and allow to re-vote
+            revert AlreadyVoted();
+        }
 
-        // TODO: Assign votes to each address (same to all)
-        // uint votePower = ERC20(_electionStore().electionTokenAddress).balanceOf(msg.sender);
+        // The voting power is the amount of votes we are going to assign to the given candidates
```
ajsantander, line 152: Comment is not necessary IMO (applies to most/all comments in this function)

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -131,26 +136,55 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
     }
 
-    function elect(address[] memory candidates) external view override {
-        uint seatCount = _electionStore().nextSeatCount;
+    function elect(address[] memory candidates) external override {
+        ElectionStore storage store = _electionStore();
+        ElectionData storage electionData = store.electionData;
 
-        if (candidates.length != seatCount) {
+        if (candidates.length == 0) {
```
ajsantander, line 143: I think we should also enforce a max number of candidates that can be voted for. Not just for algorithmic reasons, but to ensure the loop further down does not run out of gas.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -131,26 +136,55 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        for (uint i = 0; i < candidates.length; i++) {
+            address candidate = candidates[i];
+
+            // Check that the candidate is a nominee
+            if (electionData.nomineeIndexes[candidate] == 0) {
+                revert InvalidCandidate(candidate);
+            }
+
+            // Check that all the values on the candidates Array are unique
+            if (i < candidates.length - 1) {
```
ajsantander, line 164: Loops are bad in Solidity. Nested loops are really bad!

I think that this check can be performed outside of this loop, probably with a utility function that sweeps the array and reverts if it counts more than one occurrence for any given entry. I.e. it doesn't need to know which two indices collide, just that they do.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -131,26 +136,55 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            }
+
+            // Assign votes to the given candidate
+            electionData.nomineeVotes[candidate] += votePower;
+        }
+
+        // Save vote data for future usage
+        VoteData memory voteData = VoteData({candidates: candidates, votePower: votePower});
+        electionData.votes.push(voteData);
+        electionData.votesIndexes[msg.sender] = electionData.votes.length - 1;
```
ajsantander, line 180: Arguable, but to me `voteIndices` sounds more natural.

ajsantander (approved):
Awesome progress 👯 
Left some comments!

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: We can limit the number of inputs, but we cannot use seatCount for that, we can limit to the number of candidates (that is going to be finite). But limiting to the number of seats could make the election unresolved.

The simple case we discusses in our call (counting with preferences) was for 2 seats and 4 candidates and was resolved using the 3rd option.

packages/core-modules/contracts/modules/CoreElectionModule.sol
mjlescano: Also arguable, but going to keep `index`, because is the same notation that is used on the [official docs](https://docs.soliditylang.org/en/develop/types.html?highlight=array#iterable-mappings).

PR #553, opened 2021-12-07
Update run tests to matrix file

mjlescano:
Closes #430

PR #564, opened 2021-12-08
Add hardhat plugin for contract sizes

### LINKED ISSUE #550: Write utility to print contract sizes, and add test in CI that fails if any contract is oversized

ajsantander:
Fix #550

The plugin [hardhat-contract-sizes](https://github.com/ItsNickBarry/hardhat-contract-sizer) was added to the packages:
* core-contracts
* core-modules
* synthetix-main

Which adds a new `npx hardhat size-contracts` hardhat task.

This PR also adds GA tests for checking if any contract is oversize in CI.

FYI, all contracts appear to be very small rn.

### DIFF
--- a/.github/workflows/contract-size.yml
+++ b/.github/workflows/contract-size.yml
@@ -3,7 +3,7 @@ name: Contract sizes
 on: [push]
 
 jobs:
-  test:
+  size:
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/contract-size.yml
+++ b/.github/workflows/contract-size.yml
@@ -9,10 +9,6 @@ jobs:
       matrix:
         package:
           [core-contracts, core-modules, synthetix-main]
-        include:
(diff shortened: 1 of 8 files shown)


.github/workflows/contract-size.yml
```diff
@@ -0,0 +1,27 @@
...
+  size:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+          [core-contracts, core-modules, synthetix-main]
+        include:
+          - package: core-contracts
+          - package: core-modules
+          - package: synthetix-main
```
mjlescano, line 15: This is not necessary on this case because it's not adding any custom variables per matrix item. (On [the other PR](https://github.com/Synthetixio/synthetix-v3/pull/553/files) we were adding  the var `coverage-file` pero item.)
```suggestion
```

mjlescano (approved):

PR #570, opened 2021-12-09
Add Count Votes using simple counting method

### LINKED ISSUE #558: Governance: Count votes

leomassazza:
Fix #558 

Count Votes using Simple counting and plurality to resolve the election. 

In this way to resolve the election, the Election Evaluation is done by counting votes for candidates no matter the preference order. The candidate(s) that receives the more votes wins. 

When a vote is casted, all the candidates in the ballot receive the voting power of the voter, if the voter changes the vote (vote again with another ballot) the previous one should be rolled back (votes decremented) and new ones will receive their vote power.

At the moment of counting votes, there's a list of candidates with the votes received. We need to loop over this list to find the ones most voted and that will give us the winners of the election.

This PR also includes some cleanup.

### DIFF
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -12,15 +12,29 @@ interface IElectionModule {
 
     function getElectionTokenAddress() external view returns (address);
 
-    function getNominees() external view returns (address[] memory);
-
     function nominate() external;
 
     function withdrawNomination() external;
 
-    function setSeatCount(uint seats) external;
+    function getNominees() external view returns (address[] memory);
 
-    function setPeriodPercent(uint8 percent) external;
+    function elect(address[] memory candidates) external;
+
+    function isElectionEvaluated() external view returns (bool);
+
+    function evaluateElectionBatch() external;
+
+    function resolveElection() external;
+
+    function isEpochFinished() external view returns (bool);
+
+    function isNominating() external view returns (bool);
+
+    function isVoting() external view returns (bool);
+
+    function getSeatCount() external view returns (uint);
+
+    function getPeriodPercent() external view returns (uint);
 
     function setNextSeatCount(uint seats) external;
 
@@ -28,15 +42,11 @@ interface IElectionModule {
 
     function setNextPeriodPercent(uint8 percent) external;
 
-    function elect(address[] memory candidates) external;
+    function getNextSeatCount() external view returns (uint);
 
-    function getNomineeVotes(address nominee) external view returns (uint);
+    function getNextEpochDuration() external view returns (uint);
 
-    function isEpochFinished() external view returns (bool);
-
-    function isNominating() external view returns (bool);
-
-    function isVoting() external view returns (bool);
+    function getNextPeriodPercent() external view returns (uint);
 
     function setupFirstEpoch() external;
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -49,4 +49,8 @@ interface IElectionModule {
     function getNextPeriodPercent() external view returns (uint);
 
     function setupFirstEpoch() external;
+
+    function setMaxProcessingBatchSize(uint size) external;
+
+    function getMaxProcessingBatchSize() external view returns (uint);
 }
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -236,20 +236,24 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
             uint currentNomineeVotes = electionData.nomineeVotes[currentNominee];
 
             if (currentNomineeVotes > 0) {
-                if (electionData.winner.length < store.seatCount) {
+                if (electionData.nextEpochRepresentatives.length < store.seatCount) {
                     // seats not filled, fill it with whoever received votes
-                    electionData.winner.push(currentNominee);
-                    electionData.winnerVotes.push(currentNomineeVotes);
+                    electionData.nextEpochRepresentatives.push(currentNominee);
+                    electionData.nextEpochRepresentativeVotes.push(currentNomineeVotes);
 
-                    if (electionData.winner.length == store.seatCount) {
-                        (lessVotedCandidateIdx, lessVotedCandidateVotes) = _findLessVoted(electionData.winnerVotes);
+                    if (electionData.nextEpochRepresentatives.length == store.seatCount) {
+                        (lessVotedCandidateIdx, lessVotedCandidateVotes) = _findLessVoted(
+                            electionData.nextEpochRepresentativeVotes
+                        );
                     }
                 } else if (currentNomineeVotes > lessVotedCandidateVotes) {
                     // replace minimun
-                    electionData.winner[lessVotedCandidateIdx] = currentNominee;
-                    electionData.winnerVotes[lessVotedCandidateIdx] = currentNomineeVotes;
+                    electionData.nextEpochRepresentatives[lessVotedCandidateIdx] = currentNominee;
+                    electionData.nextEpochRepresentativeVotes[lessVotedCandidateIdx] = currentNomineeVotes;
 
-                    (lessVotedCandidateIdx, lessVotedCandidateVotes) = _findLessVoted(electionData.winnerVotes);
+                    (lessVotedCandidateIdx, lessVotedCandidateVotes) = _findLessVoted(
+                        electionData.nextEpochRepresentativeVotes
+                    );
                 }
             }
         }
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -19,7 +19,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
     error InvalidPeriodPercent();
     error InvalidBatchSize();
 
-    error FirstEpochAlreadySetUp();
+    error FirstEpochAlreadySet();
 
     error AlreadyVoted();
 
@@ -194,7 +194,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         electionData.addressVoted[msg.sender] = true;
     }
 
-    function isElectionEvaluated() external view virtual override returns (bool) {
+    function isElectionEvaluated() public view virtual override returns (bool) {
         return _electionStore().electionData.isElectionEvaluated;
     }
 
@@ -203,7 +203,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
             revert EpochNotFinished();
         }
 
-        if (_electionStore().electionData.isElectionEvaluated) {
+        if (isElectionEvaluated()) {
             revert ElectionAlreadyEvaluated();
         }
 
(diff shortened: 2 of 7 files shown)


packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -12,31 +12,41 @@ interface IElectionModule {
...
     function nominate() external;
 
     function withdrawNomination() external;
 
-    function setSeatCount(uint seats) external;
-
-    function setPeriodPercent(uint8 percent) external;
-
-    function setNextSeatCount(uint seats) external;
+    function getNominees() external view returns (address[] memory);
```
leomassazza, line 19: Reordered

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -12,31 +12,41 @@ interface IElectionModule {
...
-    function getNomineeVotes(address nominee) external view returns (uint);
+    function resolveElection() external;
 
     function isEpochFinished() external view returns (bool);
 
     function isNominating() external view returns (bool);
 
     function isVoting() external view returns (bool);
 
+    function getSeatCount() external view returns (uint);
```
leomassazza, line 35: Added getters for the storage variables to be consumed by external operators and the UI

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -109,18 +113,6 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         delete electionData.nomineeIndexes[msg.sender];
     }
 
-    function setSeatCount(uint seats) external override onlyOwner {
```
leomassazza, line 112: Removed option to change current voting parameters

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +192,81 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         electionData.addressVoted[msg.sender] = true;
     }
 
-    function getNomineeVotes(address nominee) external view override returns (uint) {
```
leomassazza, line 183: getNomineeVotes is not needed and only applicable on this votes counting methodology. If needed we should expose a way to see all the ballots later

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +192,81 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+    function evaluateElectionBatch() external virtual override {
+        if (!isEpochFinished()) {
+            revert EpochNotFinished();
+        }
+
+        if (_electionStore().electionData.isElectionEvaluated) {
+            revert ElectionAlreadyEvaluated();
+        }
+
+        _evaluateElectionBatchBySimpleCounting();
```
leomassazza, line 208: This is where the evaluation (counting of votes) happens. I used an internal function to explicitly express in code which alternative or methodology is used.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +192,81 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+    }
+
+    function _evaluateElectionBatchBySimpleCounting() internal virtual {
+        ElectionStore storage store = _electionStore();
+        ElectionData storage electionData = store.electionData;
+
+        uint maxBatchSize = 100;
```
leomassazza, line 215: This function iterate over the candidates to identify the TOP N (where N is the seats count) voted ones. Usually the number of candidates will be limited and manageable, but in some corner case it could be big enough to prevent this function to run due to gas limitations. The **maxBatchSize** limits the quantity of candidates to be processed on every function call.

**maxBatchSize** could be a parameter or external constant. Set to 100 just using some best guest approach.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +192,81 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            if (votes[idx] < minVotes) {
+                minVotes = votes[idx];
+                minIdx = idx;
+            }
+        }
+        return (minIdx, minVotes);
+    }
+
+    // solhint-disable-next-line no-empty-blocks
+    function resolveElection() external virtual override {
```
leomassazza, line 263: to be done in another PR

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Update: maxBatchSize is now a parameter that can be configured by the owner

packages/core-modules/contracts/storage/ElectionStorage.sol
```diff
@@ -38,6 +38,22 @@ contract ElectionStorage {
...
          */
         mapping(address => uint256) nomineeIndexes;
+        /**
+         * @dev Flag to indicate if the election was evaluated (if batched, latest batch was processed).
+         */
+        bool isElectionEvaluated;
+        /**
+         * @dev Used to keep track of the next epoch's winners. Gets erased when an epoch starts and a new council takes effect.
+         */
+        address[] winners;
```
mjlescano, line 48: Arguable, but `winners` doesn't look right, maybe `nextEpochRepresentatives`?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        }
+
+        if (_electionStore().maxProcessingBatchSize == 0) {
+            revert BatchSizeNotSet();
+        }
+
+        _evaluateElectionBatchBySimpleCounting();
+    }
+
+    function _evaluateElectionBatchBySimpleCounting() internal virtual {
```
ajsantander, line 217: So, from what I can see, this computation is split in batches configured in `store.maxProcessingBatchSize` correct?

Then the idea would be to reduce this value if an out-of-gas revert during the computation is observed. My question is: can this be adjusted during a computation that's already started?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        ElectionStore storage store = _electionStore();
+        ElectionData storage electionData = store.electionData;
+
+        uint maxBatchSize = store.maxProcessingBatchSize;
+        uint offset = 0;
+        uint previousBatchIdx = electionData.processedBatchIdx;
+        uint lessVotedCandidateVotes;
+        uint lessVotedCandidateIdx;
+
+        while (offset < maxBatchSize && !electionData.isElectionEvaluated) {
```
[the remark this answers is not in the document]
ajsantander: +1

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            uint currentNomineeVotes = electionData.nomineeVotes[electionData.nominees[previousBatchIdx + offset]];
```
ajsantander, line 228: Break this line in two for better readability and reuse of `electionData.nominees[previousBatchIdx + offset]` in  multiple lines below?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         electionData.addressVoted[msg.sender] = true;
     }
 
-    function getNomineeVotes(address nominee) external view override returns (uint) {
-        return _electionStore().electionData.nomineeVotes[nominee];
+    function isElectionEvaluated() external view virtual override returns (bool) {
+        return _electionStore().electionData.isElectionEvaluated;
+    }
+
+    function evaluateElectionBatch() external virtual override {
```
ajsantander, line 201: Are we checking that voting finished before this computation starts? I fear that if votes continue to move around while this computation is ongoing, it may break.

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        uint lessVotedCandidateVotes;
+        uint lessVotedCandidateIdx;
+
+        while (offset < maxBatchSize && !electionData.isElectionEvaluated) {
+            uint currentNomineeVotes = electionData.nomineeVotes[electionData.nominees[previousBatchIdx + offset]];
+
+            if (currentNomineeVotes > 0) {
+                if (electionData.winners.length < store.seatCount) {
+                    // seats not filled, fill it with whoever received votes
+                    electionData.winners.push(electionData.nominees[previousBatchIdx + offset]);
```
ajsantander, line 233: Store `previousBatchIdx + offset` in a variable to avoid making the addition multiple times?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,85 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+    }
+
+    function _evaluateElectionBatchBySimpleCounting() internal virtual {
+        ElectionStore storage store = _electionStore();
+        ElectionData storage electionData = store.electionData;
+
+        uint maxBatchSize = store.maxProcessingBatchSize;
+        uint offset = 0;
+        uint previousBatchIdx = electionData.processedBatchIdx;
+        uint lessVotedCandidateVotes;
```
ajsantander, line 224: Also initialize these as zero? Or don't initialize offset, since everything is auto-initialized as zero in da blokchin 👯

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander, line 228: What pushes values into `nomineeVotes` and `winnersVotes`? I see they're only read from in this computation, and I can't find who writes to them.

packages/core-modules/contracts/storage/ElectionStorage.sol
```diff
@@ -38,6 +38,22 @@ contract ElectionStorage {
...
+        /**
+         * @dev Used to keep track of the next epoch's winners votes. Gets erased when an epoch starts and a new council takes effect.
+         */
+        uint256[] winnersVotes;
```
ajsantander, line 52: Pls review english grammar in some of the variables in ElectionStorage for this rule: You usually don't use two plurals in the same sentences. I.e. winnersVotes 👉 winnerVotes. Pointed out some of these in other PRs, and I'm not going to point out every case in this one (or check). Please check.

ajsantander (changes requested):
Starting to take shape 🚀 

Left a few comments, suggestions, and questions to help me grasp this better.

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: nomineeVotes is filled in elect (while voting), winnersVotes is filled in this loop. There was one line missing though. Fixed.

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Thanks! Fixed

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Yeah, fixed

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: isEpochFinished() is the end of voting phase

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: It can since each batch is processed independently and we store the offset of the latest index processed.
Also, most likely (I'd say for sure, but maybe I'm missing some nit of gas calculation) this will revert on the first attempt to process a batch if the batch size is too large, so technically it won't be adjusted when computation is already started.

packages/core-modules/contracts/modules/CoreElectionModule.sol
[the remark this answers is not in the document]
leomassazza: I'm fixing it and in fact changing the whole structure to a for loop with a break for readability purposes.

packages/core-modules/contracts/storage/ElectionStorage.sol
ajsantander: Sort of agree yeah. How are we calling an epoch's NFT holders?
We should call this `nextEpoch<whatever-we-call-those>`

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander: Sweet 💯

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander: Ah cool 👍

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -137,7 +131,27 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+
+    function getNextSeatCount() external view override returns (uint) {
+        return _electionStore().nextSeatCount;
+    }
+
+    function getNextEpochDuration() external view override returns (uint) {
+        return _electionStore().nextEpochDuration;
+    }
+
+    function getNextPeriodPercent() external view override returns (uint) {
```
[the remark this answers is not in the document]
leomassazza: I'll do it in the cleanup ticket #576

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: done

packages/core-modules/contracts/storage/ElectionStorage.sol
leomassazza: will be covered in #576

packages/core-modules/contracts/storage/ElectionStorage.sol
leomassazza: done

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,93 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+                }
+            }
+        }
+
+        electionData.processedBatchIdx += maxBatchSize;
+    }
+
+    function _findLessVoted(uint[] storage votes) private view returns (uint, uint) {
+        uint minVotes = votes[0];
+        uint minIdx = 0;
```
ajsantander, line 266:
```suggestion
        uint minIdx = 0;

```

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -180,8 +194,93 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        for (uint idx = 0; idx < votes.length; idx++) {
+            if (votes[idx] < minVotes) {
+                minVotes = votes[idx];
+                minIdx = idx;
+            }
+        }
```
ajsantander, line 272:
```suggestion
        }

```

ajsantander (approved):

mjlescano (approved):

i-stam (approved):

PR #573, opened 2021-12-09
Add election setup for first epoch

### LINKED ISSUE #551: Governance: Setup first epoch

mjlescano:
Closes #551

### DIFF
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -343,24 +343,18 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         return !isNominating() && !isEpochFinished();
     }
 
-    function setupFirstEpoch() external virtual override onlyOwner {
+    function setupFirstEpoch() external override onlyOwner {
         ElectionStore storage store = _electionStore();
 
         if (store.epochStart != 0) {
             revert FirstEpochAlreadySet();
         }
 
-        // TODO set epoch 0 seat to 1
-        // TODO set epoch 0 period
-        // TODO set token for owner
-
-        // Flip epochs
-        // solhint-disable-next-line not-rely-on-time
-        store.epochStart = block.timestamp;
-        store.seatCount = store.nextSeatCount;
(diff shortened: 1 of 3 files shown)


packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -214,22 +214,17 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
-
-        // Flip epochs
-        // solhint-disable-next-line not-rely-on-time
-        store.epochStart = block.timestamp;
-        store.seatCount = store.nextSeatCount;
-        store.epochDuration = store.nextEpochDuration;
-        store.nominationPeriodPercent = store.nextNominationPeriodPercent;
+        store.epochStart = block.timestamp; // solhint-disable-line not-rely-on-time
+        store.seatCount = 1;
+        store.epochDuration = 2 days;
```
mjlescano, line 226: This could be parametrizable?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -214,22 +214,17 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        store.nominationPeriodPercent = 0;
+        // TODO: Assign current owner as only council member
```
mjlescano, line 228: Is this a good idea, to start with a council of just the owner?

ajsantander: I think so. It's better to have 1 head than to be headless right?

ajsantander (approved):

ajsantander (commented):
Looking good AFAIK

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: In fact this "owner" will be the setup owner or pDAO setting up the module. But we should be careful if the owner is the council itself (and there's no council defined)

leomassazza (approved):

i-stam (approved):

PR #575, opened 2021-12-10
Splits cli integration tests in files

### LINKED ISSUE #574: Split CLI integration tests into multiple files, a helper, etc

### LINKED ISSUE #674: Simplify/Unify Batching Variables location
Currently `processedBatchIdx` and `isElectionEvaluated` are inside the `ElectionData` struct and `maxProcessingBatchSize` is at `ElectionStore` level. 

If there won't be multiple elections at the same time, all of them can be moved to `ElectionStore` level to keep the `ElectionData` a per election store

ajsantander:
Fix #574


packages/cli/test/integration/execute-call.test.js
```diff
@@ -0,0 +1 @@
+describe.skip('execute-call', function () {});
```
ajsantander, line 1: To be implemented later

mjlescano (approved):

PR #631, opened 2021-12-15
Add the ability to cast votes again

### LINKED ISSUE #534: Governance: Add ability to update elected council members

leomassazza:
Fix #534

### DIFF
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -178,7 +178,7 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
             VoteData memory previousVoteData = electionData.votes[voteDataIndex];
 
             for (uint i = 0; i < previousVoteData.candidates.length; i++) {
-                electionData.nomineeVotes[previousVoteData.candidates[i]] += previousVoteData.votePower;
+                electionData.nomineeVotes[previousVoteData.candidates[i]] -= previousVoteData.votePower;
             }
 
             // Cleanup previous vote in store
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -25,8 +25,6 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
 
     error FirstEpochAlreadySet();
 
-    error AlreadyVoted();
-
     error EpochNotFinished();
     error ElectionAlreadyEvaluated();
     error BatchSizeNotSet();
@@ -176,8 +174,15 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         }
 
         if (electionData.addressVoted[msg.sender]) {
-            // TODO: if msg.sender already voted, rollback previous votes and allow to re-vote
-            revert AlreadyVoted();
+            uint voteDataIndex = electionData.votesIndexes[msg.sender];
+            VoteData memory previousVoteData = electionData.votes[voteDataIndex];
+
+            for (uint i = 0; i < previousVoteData.candidates.length; i++) {
+                electionData.nomineeVotes[previousVoteData.candidates[i]] += previousVoteData.votePower;
+            }
+
+            // Cleanup previous vote in store
+            electionData.votes[voteDataIndex] = VoteData({candidates: new address[](0), votePower: 0});
         }
 
         uint votePower = ERC20(store.electionTokenAddress).balanceOf(msg.sender);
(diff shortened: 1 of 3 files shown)


packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -176,8 +174,15 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
         }
 
         if (electionData.addressVoted[msg.sender]) {
-            // TODO: if msg.sender already voted, rollback previous votes and allow to re-vote
-            revert AlreadyVoted();
+            uint voteDataIndex = electionData.votesIndexes[msg.sender];
+            VoteData memory previousVoteData = electionData.votes[voteDataIndex];
+
+            for (uint i = 0; i < previousVoteData.candidates.length; i++) {
+                electionData.nomineeVotes[previousVoteData.candidates[i]] += previousVoteData.votePower;
```
mjlescano, line 181: Shouldn't this have to substract (`-=`) the votePower instead of add (`+=`) it ?

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -176,8 +174,15 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+            }
+
+            // Cleanup previous vote in store
+            electionData.votes[voteDataIndex] = VoteData({candidates: new address[](0), votePower: 0});
```
mjlescano, line 185: Hmm, not 100% about this, but it doesn't initialize the values by itself? like so:

```suggestion
            electionData.votes[voteDataIndex] = VoteData();
```

packages/core-modules/test/contracts/modules/CoreElectionModule.test.js
```diff
@@ -347,13 +347,22 @@ describe('CoreElectionModule Setup, Getters, Setters and Voting', () => {
       assertBn.eq(await ElectionStorageMock.getVoterVoteVotePowerMock(voters[4].address), 100);
     });
 
-    describe('when attempting to vote again', () => {
-      it('reverts', async () => {
-        await assertRevert(
-          CoreElectionModule.connect(voters[1]).elect([candidates[0].address], [0]),
-          'AlreadyVoted'
+    describe('when casting a vote again', () => {
```
mjlescano, line 350: Inside this `describe`, I would cast 2 votes, so the test contains itself and is not dependant on running with after the previous one. E.g.: if you run this with `describe.only` it would fail or the test wouldn't be testing the double vote.

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander: +1

packages/core-modules/contracts/modules/CoreElectionModule.sol
ajsantander: Yeah, thought the same thing. Could work.

ajsantander (approved):

ajsantander (commented):
Looks good! I just adhere to @mjlescano 's comment about subtracting instead of adding.
And I think there should be a test case for it, otherwise the test would have caught that.

packages/core-modules/test/contracts/modules/CoreElectionModule.test.js
mjlescano: FYI, this is just an opinion I think we should start to think about, we are doing this same way of testing on other places, this PR is not the place to implement the fix.

i-stam (approved):

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: 🤦‍♂️ yes. Thanks! Fixed

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Nop. It's not allowed by the compiler. It failed with an error saying there are missing arguments. That was my first, natural choice.

packages/core-modules/test/contracts/modules/CoreElectionModule.test.js
leomassazza: Like it. Added a ticket for that #636

packages/core-modules/test/contracts/modules/CoreElectionModule.Plurality.test.js
```diff
@@ -260,5 +260,105 @@ describe('CoreElectionModule Count Votes using Simple Plurality strategy', () =>
         });
       });
     });
+
+    describe('when re-casting votes', async () => {
```
leomassazza, line 264: Added a test that will count the votes when some voter changes the vote.

i-stam (commented):
It seems to work now, I am not a big fan of so many loops but if we cannot do it in a better way, it is what it is.

PR #632, opened 2021-12-15
ELECTION 2 feature branch

leomassazza:
Feature branch for the milestone:"ELECTION 2: Finish first draft implementation of the ElectionModule"


ajsantander:
This branch is to be used only as reference for the final implementation in the next sprint. Closing without merge, but keeping the branch.

PR #645, opened 2021-12-16
Disable linter Solidity time rule

### LINKED ISSUE #643: Consider replacing solhint inline disables of time with a rule

ajsantander:
Fix #643


leomassazza (approved):
Love it :)

PR #649, opened 2021-12-16
Shuffle NFTs 

### LINKED ISSUE #460: Election resolution with NFT shuffling

leomassazza:
Fix #460

### DIFF
--- a/packages/core-modules/contracts/modules/CoreElectionModule.sol
+++ b/packages/core-modules/contracts/modules/CoreElectionModule.sol
@@ -319,12 +319,6 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
     }
 
     function resolveElection() external virtual override {
-        ElectionStore storage store = _electionStore();
-        MemberToken memberToken = MemberToken(store.memberTokenAddress);
-        address[] memory nextEpochMembers = _currentElectionData().nextEpochMembers;
-        uint currentMembersSize = store.members.length;
-        uint nextEpochMembersSize = nextEpochMembers.length;
-
         if (!isEpochFinished()) {
             revert EpochNotFinished();
         }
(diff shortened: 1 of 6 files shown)


packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -313,15 +318,37 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+
+        if (!isEpochFinished()) {
+            revert EpochNotFinished();
+        }
+
+        if (!isElectionEvaluated()) {
+            revert ElectionNotEvaluated();
+        }
+
+        for (uint i = 0; i < currentMembersSize; i++) {
```
leomassazza, line 336: Back to burn all and mint again to simplify the logic (and reduce the chance of errors) to transfer the right token and prevent re-minting unburned tokens

packages/core-modules/contracts/modules/CoreElectionModule.sol
```diff
@@ -313,15 +318,37 @@ contract CoreElectionModule is IElectionModule, ElectionStorage, OwnableMixin {
...
+        uint currentMembersSize = store.members.length;
+        uint nextEpochMembersSize = nextEpochMembers.length;
+
+        if (!isEpochFinished()) {
+            revert EpochNotFinished();
+        }
+
+        if (!isElectionEvaluated()) {
+            revert ElectionNotEvaluated();
+        }
```
ajsantander, line 334: Can we perform these two checks before line 322 to save reversion gas?

ajsantander (approved):
Super clean!

packages/core-modules/contracts/modules/CoreElectionModule.sol
leomassazza: Done, and please check this decision.

Added the code after the checking to an internal function that later can be implemented in mixings for the different strategies.

i-stam (approved):

PR #651, opened 2021-12-17
Simplify elect function

### LINKED ISSUE #641: Review elect function parameter interface: Do the two arrays really make sense? Can the code be simplified?

ajsantander:
Fix #641

### DIFF
--- a/packages/core-contracts/contracts/utils/ArrayUtil.sol
+++ b/packages/core-contracts/contracts/utils/ArrayUtil.sol
@@ -29,4 +29,24 @@ library ArrayUtil {
 
         delete positions[value];
     }
+
+    function hasDuplicates(address[] memory array) internal pure returns (bool) {
+        uint length = array.length;
+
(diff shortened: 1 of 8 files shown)


packages/core-contracts/contracts/utils/ArrayUtil.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    function hasDuplicates(address[] memory array) internal pure returns (bool) {
+        uint length = array.length;
+
+        for (uint i = 0; i < length; i++) {
+            address value = array[i];
+
+            if (i < length - 1) {
+                for (uint256 j = i + 1; j < length; j++) {
```
leomassazza, line 40: That's the N^2 we wanted to avoid and why the introduction of the second array.

leomassazza:
This PR simplifies the introduction of the 2nd array but brings back the problem of a nested for cycle

PR #667, opened 2021-12-20
Tweaks on custom errors

### LINKED ISSUE #273: Fix the unrecognized custom error issue in tests (probably hardhat error)
And remove the extra check in assert-revert.js when is fixed

Related to https://github.com/nomiclabs/hardhat/issues/1996

ajsantander:
Even though this does not solve #273, it emits a warning whenever an error cannot be parsed, being more explicit and not just ignoring the problem.

It also addresses what I believe to be a bug in OZ ERC721 implementation.

### DIFF
--- a/packages/core-contracts/contracts/token/ERC721.sol
+++ b/packages/core-contracts/contracts/token/ERC721.sol
@@ -11,8 +11,6 @@ import "./ERC721Storage.sol";
 import "../utils/AddressUtil.sol";
 import "../utils/StringUtil.sol";
 
-import "hardhat/console.sol";
-
 /*
     Reference implementations:
     * OpenZeppelin - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol
--- a/packages/core-contracts/contracts/token/ERC721.sol
+++ b/packages/core-contracts/contracts/token/ERC721.sol
@@ -11,6 +11,8 @@ import "./ERC721Storage.sol";
 import "../utils/AddressUtil.sol";
 import "../utils/StringUtil.sol";
 
+import "hardhat/console.sol";
+
 /*
     Reference implementations:
     * OpenZeppelin - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol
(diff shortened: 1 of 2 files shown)


packages/core-contracts/contracts/token/ERC721.sol
```diff
@@ -234,14 +236,8 @@ contract ERC721 is IERC721, IERC721Metadata, ERC721Storage {
         if (AddressUtil.isContract(to)) {
             try IERC721Receiver(to).onERC721Received(msg.sender, from, tokenId, data) returns (bytes4 retval) {
                 return retval == IERC721Receiver.onERC721Received.selector;
```
ajsantander, line 238: I believe this OZ implementation is wrong. The function is supposed to return false, not throw.

packages/core-contracts/contracts/token/ERC721.sol
```diff
@@ -11,6 +11,8 @@ import "./ERC721Storage.sol";
 import "../utils/AddressUtil.sol";
 import "../utils/StringUtil.sol";
 
+import "hardhat/console.sol";
```
leomassazza, line 14: rogue import

packages/core-contracts/contracts/token/ERC721.sol
leomassazza: I see your point (and in fact I don't agree at all with OZ's) but the reason they are doing it is to bubble up any revert from the underlying contract to better identify the issue.

So, not pro or against this change :)

packages/core-js/utils/assertions/assert-revert.js
```diff
@@ -12,12 +12,20 @@ module.exports = async function assertRevert(tx, expectedMessage) {
...
-      !receivedMessage.includes(expectedMessage) &&
-      !receivedMessage.includes('reverted with an unrecognized custom error')
-    ) {
+    if (!receivedMessage.includes(expectedMessage)) {
+      // ----------------------------------------------------------------------------
+      // TODO: Remove this check once the following issue is solved in hardhat:
+      // https://github.com/nomiclabs/hardhat/issues/1996
+      // Basically, the first time tests are run, the revert reason is not parsed,
+      // but the second time it is parsed just fine;
+      if (receivedMessage.includes('reverted with an unrecognized custom error')) {
```
leomassazza, line 21: like it. This is more clear.

leomassazza (approved):
LGTM (just a np and a comment)

i-stam (approved):

packages/core-contracts/contracts/token/ERC721.sol
ajsantander: Oh oops! 🙏

ajsantander: Removed

packages/core-contracts/contracts/token/ERC721.sol
ajsantander: Let's keep this one. I am pro maintaining OZ code, until its arguably incorrect.

PR #670, opened 2021-12-21
Use nominate/accept pattern for UUPS upgrades

ajsantander:


ajsantander:
Closing because this applies to EOA's (to make sure that private keys are not misplaced), and it doesn't really apply to contract accounts.

PR #676, opened 2021-12-21
More cli tests

### LINKED ISSUE #611: Finish implementing CLI integration tests

ajsantander:
Fix #611

### DIFF
--- a/packages/cli/internal/signatures.js
+++ b/packages/cli/internal/signatures.js
@@ -17,7 +17,9 @@ function getFullFunctionSignature(functionAbi, functionParameters) {
   }
 
   // Collect return values
-  const outputDescriptions = functionAbi.outputs.map((output) => `${output.type}${` ${output.name}` || ''}`);
+  const outputDescriptions = functionAbi.outputs.map(
+    (output) => `${output.type}${` ${output.name}` || ''}`
+  );
 
   // Function name
   let str = `${functionAbi.name}${multiline ? '(\n' : '('}`;
--- a/packages/cli/internal/signatures.js
+++ b/packages/cli/internal/signatures.js
@@ -17,12 +17,7 @@ function getFullFunctionSignature(functionAbi, functionParameters) {
   }
 
   // Collect return values
-  const outputDescriptions = [];
-  for (let i = 0; i < functionAbi.outputs.length; i++) {
-    const output = functionAbi.outputs[i];
-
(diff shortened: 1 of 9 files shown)


packages/cli/scripts/test.sh
```diff
@@ -11,3 +11,6 @@ npx hardhat deploy --network local --clear --quiet --no-confirm --instance test
 # Run tests
 cd ../../..
 npx mocha
+
+# Kill node
+pkill node
```
ajsantander, line 16: Was leaving `npx hardhat node` running in the background after process ended.

packages/cli/subtasks/pick-function.js
```diff
@@ -16,9 +16,7 @@ subtask(SUBTASK_PICK_FUNCTION, 'Pick a function from the given contract').setAct
       const selector = selectors.find((selector) => selector.name === functionAbi.name).selector;
 
       return {
-        title: `${chalk.gray(`${hre.cli.contractName}.`)}${fullSignature}${chalk.gray(
-          ` ${selector}`
-        )}`,
+        title: `${fullSignature}${chalk.gray(` ${selector}`)}`,
```
ajsantander, line 19: Looks like prompter lib autocomplete is not that good. E.g. when typing "set" it will not pick "SomeModule.setUintValue" but will pick "setUintValue" (when removing contract name).

packages/cli/internal/signatures.js
```diff
@@ -10,11 +16,30 @@ function getFullFunctionSignature(functionAbi, functionParameters) {
     parameterDescriptions.push(`${input.type} ${input.name}${valueDescription}`);
   }
 
+  // Collect return values
+  const outputDescriptions = [];
+  for (let i = 0; i < functionAbi.outputs.length; i++) {
+    const output = functionAbi.outputs[i];
+
+    outputDescriptions.push(`${output.type}${output.name ? ` ${output.name}` : ''}`);
+  }
```
mjlescano, line 25: Maybe this is clearer:

```suggestion
  const outputDescriptions = functionAbi.outputs.map((output) =>  {
    return `${output.type}${output.name ? ` ${output.name}` : ''}`
  });
```

mjlescano (approved):

leomassazza (approved):

i-stam (approved):

PR #677, opened 2021-12-22
Refactor readability of AST helpers usage

mjlescano:


.eslintrc.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   env: {
     mocha: true,
     node: true,
+    es6: true,
```
mjlescano, line 6: This is so it doesn't throw an error when using "new" features from ES6 (e.g. `new Set()`). It is not really needed for this PR, but is a good and simple addition nevertheless.

packages/core-js/utils/misc/unique.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Helper array unique function to be used with Array.filter method.
+ *   e.g.: [1, 1, 2, 3].filter(unique) // => [1, 2, 3]
+ */
+module.exports = function unique(value, index, self) {
+  return self.indexOf(value) === index;
+};
```
mjlescano, line 7: We've been using this pattern for getting the unique values from an array which IMHO is kinda difficult to read, so, I just moved it to an util which makes it more readable when using it.

packages/deployer/internal/contract-helper.js
```diff
@@ -40,22 +41,16 @@ async function getModulesSelectors() {
 }
 
 function findDuplicateSelectors(selectors) {
-  const duplicates = selectors
-    .map((s) => s.selector)
-    .filter((s, index, selectors) => selectors.indexOf(s) !== index);
```
mjlescano, line 45: We were repeating this pattern for getting duplicate values on an array, which IMHO is somewhat hard to read and understand. So, I moved it to a helper filter function on `core-js`.

Also, this had a bug were if the  repetition was present more than two times, it would repeat the error. For example, for  `[1,1]` would return `[1]`, but  for `[1,1,1] => [1,1]` and so on. (The fix  can be seen on the new `array-filters.js` file)

packages/core-js/utils/misc/array-filters.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Filter all the values from an array to get only repeated ones
+ *   e.g.: [1, 1, 2, 3].filter(onlyRepeated) // => [1]
+ */
+function onlyRepeated(value, index, self) {
+  const last = self.lastIndexOf(value);
+  return self.indexOf(value) !== last && index === last;
```
mjlescano, line 7: We return  only the  last appearance  of the repeated values.

packages/deployer/internal/router-ast-validator.js
```diff
@@ -16,25 +17,30 @@ class RouterASTValidator {
     const routerSelectors = findYulCaseValues('Router', this.asts['Router']);
 
     const errors = [];
-    moduleSelectors.forEach((contractSelector) => {
-      if (!routerSelectors.some((s) => s.selector === contractSelector.selector)) {
+
+    for (const contractSelector of moduleSelectors) {
+      const selectorExists = routerSelectors.some((s) => s.selector === contractSelector.selector);
+
+      if (!selectorExists) {
```
mjlescano, line 24: I just named the variable to improve readability.

PR #684, opened 2021-12-23
Increases coverage in core-js package

ajsantander:


i-stam (approved):
lgtm

packages/core-js/utils/ast/storage-struct.js
```diff
@@ -33,11 +18,21 @@ async function buildContractsStructMap(contractNodes) {
     }
   }
 
-  orderContractsStructMap(structs);
+  _orderContractsStructMap(structs);
 
   return structs;
 }
 
+function _orderContractsStructMap(structs) {
```
leomassazza, line 26: this will only order for contract.name, while before we ordered by contract.name AND struct.name if there were two structs in the same contract.

ajsantander: Yeah, I realized this. But couldn't find a case to test against. IMO we should increase this when we need it, and when we do, we should add tests for it.

PR #732, opened 2022-01-14
Ability to nominate

### LINKED ISSUE #730: Ability to nominate

ajsantander:
Fix #730


packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -37,4 +37,8 @@ interface IElectionModule {
     function getVotingPeriodStartDate() external view returns (uint64);
 
     function isEpochEvaluated() external view returns (bool);
+
+    function isNominated(address candidate) external view returns (bool);
+
+    function getNominees() external view returns (address[] memory);
```
leomassazza, line 43: this could be long and may require pagination. If not part of this PR, please add a ticket to investigate it

leomassazza (approved):
Approved. Only one comment to check

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: Let's ignore pagination in all view functions for now.
This means that we need only consider pagination during an election evaluation.

mjlescano (approved):

PR #735, opened 2022-01-14
Use initializableMixin where possible

### LINKED ISSUE #704: InitializebleModules: Update current modules to use InitializableMixin

leomassazza:
Fix #704

### DIFF
--- a/packages/core-contracts/contracts/ownership/Ownable.sol
+++ b/packages/core-contracts/contracts/ownership/Ownable.sol
@@ -13,11 +13,15 @@ contract Ownable is IOwnable, OwnableMixin {
     error NotNominated(address addr);
 
     function acceptOwnership() external override {
+        _acceptOwnershipBy(msg.sender);
+    }
+
+    function _acceptOwnershipBy(address newNominatedOwner) internal {
         OwnableStore storage store = _ownableStore();
 
         address currentNominatedOwner = store.nominatedOwner;
-        if (msg.sender != currentNominatedOwner) {
-            revert NotNominated(msg.sender);
+        if (newNominatedOwner != currentNominatedOwner) {
+            revert NotNominated(newNominatedOwner);
         }
 
         emit OwnerChanged(store.owner, currentNominatedOwner);
@@ -26,7 +30,7 @@ contract Ownable is IOwnable, OwnableMixin {
         store.nominatedOwner = address(0);
     }
 
-    function nominateNewOwner(address newNominatedOwner) external override onlyOwnerIfSet {
+    function nominateNewOwner(address newNominatedOwner) public override onlyOwnerIfSet {
         OwnableStore storage store = _ownableStore();
 
         if (newNominatedOwner == address(0)) {
--- a/packages/core-contracts/contracts/ownership/Ownable.sol
+++ b/packages/core-contracts/contracts/ownership/Ownable.sol
@@ -12,16 +12,12 @@ contract Ownable is IOwnable, OwnableMixin {
 
     error NotNominated(address addr);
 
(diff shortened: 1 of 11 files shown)


packages/core-contracts/contracts/ownership/Ownable.sol
```diff
@@ -13,11 +13,15 @@ contract Ownable is IOwnable, OwnableMixin {
     error NotNominated(address addr);
 
     function acceptOwnership() external override {
+        _acceptOwnershipBy(msg.sender);
```
leomassazza, line 16: added the internal function to accept ownership from the initialization function

packages/core-contracts/contracts/ownership/Ownable.sol
```diff
@@ -12,7 +12,7 @@ contract Ownable is IOwnable, OwnableMixin {
 
     error NotNominated(address addr);
 
-    function acceptOwnership() external override {
+    function acceptOwnership() public override {
```
ajsantander, line 15: Why change it to public?

packages/core-modules/contracts/interfaces/IOwnerModule.sol
```diff
@@ -0,0 +1,8 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IOwnerModule {
```
ajsantander, line 4: Should we be consistent and add interfaces to all modules? I believe the UpgradeModule is the only one without an interface right now.

ajsantander (approved):
Awesome!
(just a couple of questions, feel free to merge and answer in PM)

packages/core-modules/contracts/modules/OwnerModule.sol
```diff
@@ -2,8 +2,21 @@
...
+    }
 
+    function isOwnerModuleInitialized() external view override returns (bool) {
+        return _isInitialized();
+    }
+
+    function initializeOwnerModule(address initialOwner) external override onlyIfNotInitialized {
+        nominateNewOwner(initialOwner);
+        acceptOwnership();
+    }
```
mjlescano, line 21: Wouldn't this have the same functionality, having in mind that the call to `acceptOwnership();` uses `msg.sender`?:
```suggestion
    function initializeOwnerModule() external override onlyIfNotInitialized {
        nominateNewOwner(msg.sender);
        acceptOwnership();
    }
```

leomassazza: As discussed in Discord. 

We prefer to have the address explicitly. One caveat is that using the initialize we won't be able to use an EOA to nominate a multisig or gov entity since it will use the msg.sender to accept the ownership in line 20.
But having an _acceptOwnershipBy(address) can allow us to use a different address in the initializer and automatically accept the ownership by that address, but is risky since a typo there can set the owner to an invalid address.

leomassazza: As am afterthought.

If we want to set the owner to a Mutlisig or Gov Module, we should first initialize the OwnerModule to the EAO running the deploy and then set the new owner via usual way: Calling nominateNewOwner(multisigAddress) and as a post-deploy accept the ownership by the multisig or govModule using acceptOwnership().

The good part is that we ensure the OwnerModule is initialized to a known address.

mjlescano (approved):

packages/core-modules/contracts/modules/OwnerModule.sol
mjlescano: I like that option! Is less prone to errors during the initialization

PR #737, opened 2022-01-18
Create a satellite interface for modules and keep data on chain

mjlescano:


packages/core-contracts/contracts/satellite/SatelliteFactory.sol
```diff
@@ -1,6 +1,16 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+struct Satellite {
+    bytes32 id;
+    string contractName;
```
ajsantander, line 6: bytes32?

packages/core-contracts/contracts/satellite/SatelliteFactory.sol
```diff
@@ -1,6 +1,16 @@
...
+    address deployedAddress;
+}
+
 abstract contract SatelliteFactory {
```
ajsantander, line 10: IMO keep SatelliteFactory only, and always return an array

mjlescano: Done!

ajsantander (approved):

packages/core-contracts/contracts/satellite/SatelliteFactory.sol
mjlescano: Done!

PR #739, opened 2022-01-18
InitializableModules: Update CLI

### LINKED ISSUE #706: InitializableModules: Update CLI to raise a warning when interacting with non-initialized modules

leomassazza:
Fix #706

The PR adds the ability to check if the contract selected is initializable and if it is already initialized, raising a Warn message if not. That is accomplished with the introduction of a new subtask that is executed in the pick-function SUBTASK before showing the list of functions available.

### DIFF
--- a/packages/cli/subtasks/check-initializable.js
+++ b/packages/cli/subtasks/check-initializable.js
@@ -1,0 +1,31 @@
+const { subtask } = require('hardhat/config');
+const { SUBTASK_CHECK_INITIALIZATION } = require('../task-names');
+const logger = require('@synthetixio/core-js/utils/io/logger');
+const { capitalize } = require('@synthetixio/core-js/utils/misc/strings');
+
+subtask(
+  SUBTASK_CHECK_INITIALIZATION,
+  'Check if current contract is initializable and if it is initialized'
+).setAction(async (taskArguments, hre) => {
+  const target = hre.deployer.deployment.general.contracts[hre.cli.contractName];
+  const address = target.proxyAddress || target.deployedAddress;
+  const abi = hre.deployer.deployment.abis[hre.cli.contractName];
+  const abiFunctions = abi.filter((abiItem) => abiItem.name && abiItem.type === 'function');
+
+  const capitalizedContractName = capitalize(hre.cli.contractName.split(':')[1]);
+  const isInitializablefunctionName = `is${capitalizedContractName}Initialized`;
+  const initializafunctionName = `initialize${capitalizedContractName}()`;
+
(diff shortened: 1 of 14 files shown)


packages/cli/subtasks/pick-function.js
```diff
@@ -21,6 +21,8 @@ subtask(SUBTASK_PICK_FUNCTION, 'Pick a function from the given contract').setAct
       };
     });
 
+    await hre.run(SUBTASK_CHECK_INITIALIZATION);
```
leomassazza, line 24: Do the checking before showing the functions

mjlescano (approved):

packages/cli/subtasks/check-initializable.js
```diff
@@ -0,0 +1,31 @@
...
+  SUBTASK_CHECK_INITIALIZATION,
+  'Check if current contract is initializable and if it is initialized'
+).setAction(async (taskArguments, hre) => {
+  const target = hre.deployer.deployment.general.contracts[hre.cli.contractName];
+  const address = target.proxyAddress || target.deployedAddress;
+  const abi = hre.deployer.deployment.abis[hre.cli.contractName];
+  const abiFunctions = abi.filter((abiItem) => abiItem.name && abiItem.type === 'function');
+
+  const capitalizedContractName = capitalize(hre.cli.contractName.split(':')[1]);
+  const isInitializableFunctionName = `is${capitalizedContractName}Initialized`;
```
ajsantander, line 16: Small semantic improvement: this should be called isInitializedFunctionName. Its existance determines if the contract is initializable. Becomes more relevant when you actually call the function a few lines below.

packages/cli/subtasks/check-initializable.js
```diff
@@ -0,0 +1,31 @@
...
+  const initializeFunctionName = `initialize${capitalizedContractName}`;
+
+  const isInitializable = abiFunctions.some((f) => f.name === isInitializableFunctionName);
+
+  if (isInitializable) {
+    const contract = new hre.ethers.Contract(address, abi, hre.ethers.provider);
+    const result = await contract[isInitializableFunctionName]();
```
ajsantander, line 23: Here’s where the small semantic problem pointed out above makes the code weird to read.

ajsantander (approved):
Awesome! Just a mini nit

packages/cli/subtasks/check-initializable.js
leomassazza: in fact, is not semantic, it was the wrong name (a mental typo) :)

packages/cli/subtasks/check-initializable.js
leomassazza: fixed

PR #747, opened 2022-01-20
Fix sample-project tests

mjlescano:
This PR fixes the following error when running the sample project tests:

Captura de Pantalla 2022-01-20 a la(s) 12 28 36

Which can be seen here: https://github.com/Synthetixio/synthetix-v3/runs/4880610175?check_suite_focus=true

### DIFF
--- a/packages/deployer/.mocharc.json
+++ b/packages/deployer/.mocharc.json
@@ -1,5 +1,6 @@
 {
   "spec": ["test/unit/**/*.test.js", "test/integration/**/*.test.js"],
+  "timeout": 10000,
   "watch-files": [
     "internal/**/*.js",
     "subtasks/**/*.js",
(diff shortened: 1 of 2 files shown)


packages/deployer/.mocharc.json
```diff
@@ -1,5 +1,6 @@
 {
   "spec": ["test/unit/**/*.test.js", "test/integration/**/*.test.js"],
+  "timeout": 10000,
```
mjlescano, line 3: This is necessary because *sometimes* the github actions CI is taking a little longer than the default `2secs`.

packages/deployer/subtasks/clear-deployments.js
```diff
@@ -12,8 +11,6 @@ subtask(
   SUBTASK_CLEAR_DEPLOYMENTS,
   'Delete all previous deployment data on the current environment'
 ).setAction(async ({ instance }, hre) => {
-  await hre.run(TASK_CLEAN);
```
mjlescano, line 15: The error was being caused because we were calling this Task programmatically from the deploy script on the same process, which removed the artifacts folders but it didn't remove the cache from hardhat, causing a missing error on the second+ run.

Also, this souldn't be necessary anymore as we started running the compile task with the `force: true` flag, which causes the regenartion of the artifacts.

ajsantander (approved):

PR #751, opened 2022-01-20
Add ERC20HistoricalBalance

### LINKED ISSUE #750: Create ERC20 with historical balance to use as basic VotePower

leomassazza:
Fix #750

### DIFF
--- a/packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
+++ b/packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
@@ -78,9 +78,6 @@ contract ERC20HistoricalBalance is ERC20, IERC20HistoricalBalance, ERC20Historic
 
         _moveVotingPower(from, address(0), amount);
         _writeCheckpoint(_erc20HistoricalBalanceStore().totalSupplyCheckpoints, _subtract, amount);
-        ERC20Store storage store = _erc20Store();
-        store.balanceOf[from] -= amount;
-        store.totalSupply -= amount;
     }
 
     function _moveVotingPower(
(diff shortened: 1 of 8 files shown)


packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
```diff
@@ -0,0 +1,135 @@
...
+                high = mid;
+            } else {
+                low = mid + 1;
+            }
+        }
+
+        return high == 0 ? 0 : ckpts[high - 1].balance;
+    }
+
+    function _transfer(
```
leomassazza, line 59: @ajsantander what is preferred here, overriding and calling super as I do here or adding a hook on ERC20 pre/post some actions (mint, burn, transfer) and override the hooks as done by OZs? 
[OZ solution](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b3b83b558ebb9982e27ae5ee0bb5f33f278863dd/contracts/token/ERC20/ERC20.sol#L248)

mjlescano (approved):

packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
ajsantander: I think overrides are more explicit, but don't have a strong opinion.

packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
```diff
@@ -0,0 +1,135 @@
...
+        address from,
+        address to,
+        uint256 amount
+    ) internal override {
+        super._transfer(from, to, amount);
+
+        _moveVotingPower(from, to, amount);
```
ajsantander, line 66: Can we rename this to something that does not imply governance?

ajsantander (approved):

PR #752, opened 2022-01-20
Initializable/use-boolean-flag

### LINKED ISSUE #749: InitializableModules: Change samples to use boolean flag

leomassazza:
Fix #749

### DIFF
--- a/packages/core-contracts/contracts/ownership/OwnableStorage.sol
+++ b/packages/core-contracts/contracts/ownership/OwnableStorage.sol
@@ -3,9 +3,9 @@ pragma solidity ^0.8.0;
 
 contract OwnableStorage {
     struct OwnableStore {
+        bool initialized;
         address owner;
         address nominatedOwner;
-        bool initialized;
     }
 
     function _ownableStore() internal pure returns (OwnableStore storage store) {
--- a/packages/core-contracts/contracts/ownership/OwnableStorage.sol
+++ b/packages/core-contracts/contracts/ownership/OwnableStorage.sol
@@ -5,6 +5,7 @@ contract OwnableStorage {
     struct OwnableStore {
         address owner;
         address nominatedOwner;
+        bool initialized;
     }
 
     function _ownableStore() internal pure returns (OwnableStore storage store) {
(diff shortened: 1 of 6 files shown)


mjlescano (approved):

packages/core-contracts/contracts/mocks/utils/SetUtilMock.sol
```diff
@@ -0,0 +1,72 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../utils/SetUtil.sol";
+
+contract Bytes32SetMock {
```
ajsantander, line 6: Should you update the feature branch from main so that this diff isn't visible?

packages/core-contracts/contracts/ownership/OwnableStorage.sol
```diff
@@ -5,6 +5,7 @@ contract OwnableStorage {
     struct OwnableStore {
         address owner;
         address nominatedOwner;
+        bool initialized;
```
ajsantander, line 8: Should we put all these at the beginning of namespaces before we go live into any real networks?

packages/deployer/subtasks/clear-deployments.js
```diff
@@ -1,6 +1,5 @@
 const del = require('del');
 const { subtask } = require('hardhat/config');
-const { TASK_CLEAN } = require('hardhat/builtin-tasks/task-names');
```
ajsantander, line 3: Also this diff should go once the feature branch is updated to main.

ajsantander (approved):
Looking good! Just a couple of nits.

packages/deployer/subtasks/clear-deployments.js
leomassazza: fixed

packages/core-contracts/contracts/ownership/OwnableStorage.sol
leomassazza: done

packages/core-contracts/contracts/mocks/utils/SetUtilMock.sol
leomassazza: fixed

PR #755, opened 2022-01-21
Adhere to initializable pattern in ElectionModule

### LINKED ISSUE #720: Adhere to the new initializable pattern in ElectionModule

ajsantander:
Fix #720

### DIFF
--- a/packages/core-modules/contracts/submodules/election/ElectionSchedule.sol
+++ b/packages/core-modules/contracts/submodules/election/ElectionSchedule.sol
@@ -17,6 +17,10 @@ contract ElectionSchedule is ElectionBase {
     }
 
     function _getCurrentPeriodType() internal view returns (ElectionPeriod) {
+        if (!_electionStore().initialized) {
+            return ElectionPeriod.Null;
+        }
+
         EpochData storage epoch = _getCurrentEpoch();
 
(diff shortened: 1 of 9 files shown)


packages/core-modules/contracts/submodules/election/ElectionSchedule.sol
```diff
@@ -17,6 +17,10 @@ contract ElectionSchedule is ElectionBase {
     }
 
     function _getCurrentPeriodType() internal view returns (ElectionPeriod) {
```
leomassazza, line 19: You can use the onlyIfInitialized modifier

ajsantander: Yep, this is something I considered and chose to go with this design, so I wouldn't have to put `onlyIfInitialized` all over the place.

But I do see your point. This may be a form of semantic overloading. If you consider it to be so, would you mind merging this PR and letting me know, so that I create a new issue to remove the overload? It would simplify my work on the evaluations feature.

leomassazza: Sure, creating a ticket for that.

leomassazza (approved):

PR #760, opened 2022-01-24
Election module/resolution

### LINKED ISSUE #723: Election evaluation

ajsantander:
Fix #723

### DIFF
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -12,7 +12,7 @@ interface IElectionModule {
         uint64 epochEndDate
     ) external;
 
-    function isElectionModuleInitialized() external bool returns (bool);
+    function isElectionModuleInitialized() external view returns (bool);
 
     // ---------------------------------------
     // Owner functions
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -93,8 +93,6 @@ interface IElectionModule {
 
     function getCurrentPeriodType() external view returns (uint);
 
-    function isEpochEvaluated() external view returns (bool);
-
     // Nominations
     // ~~~~~~~~~~~~~~~~~~
 
@@ -102,7 +100,7 @@ interface IElectionModule {
 
     function getNominees() external view returns (address[] memory);
 
-    // Votes / ballots
+    // Votes
     // ~~~~~~~~~~~~~~~~~~
 
     function calculateBallotId(address[] calldata candidates) external pure returns (bytes32);
@@ -116,4 +114,9 @@ interface IElectionModule {
     function getBallotVotes(bytes32 ballotId) external view returns (uint);
 
     function getBallotCandidates(bytes32 ballotId) external view returns (address[] memory);
+
+    // Resolutions
+    // ~~~~~~~~~~~~~~~~~~
+
+    function isElectionEvaluated() external view returns (bool);
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -119,4 +119,6 @@ interface IElectionModule {
     // ~~~~~~~~~~~~~~~~~~
 
     function isElectionEvaluated() external view returns (bool);
+
+    function getCandidateVotes(address candidate) external view returns (uint);
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -38,6 +38,8 @@ interface IElectionModule {
 
     function setMaxDateAdjustmentTolerance(uint64 newMaxDateAdjustmentTolerance) external;
 
+    function setDefaultBallotEvaluationBatchSize(uint newDefaultBallotEvaluationBatchSize) external;
+
     // ---------------------------------------
     // Nomination functions
     // ---------------------------------------
@@ -56,7 +58,7 @@ interface IElectionModule {
     // Election resolution
     // ---------------------------------------
 
-    function evaluate() external;
+    function evaluate(uint numBallots) external;
 
     function resolve() external;
 
@@ -77,6 +79,8 @@ interface IElectionModule {
         );
 
     function getMaxDateAdjustmenTolerance() external view returns (uint64);
+
+    function getDefaultBallotEvaluationBatchSize() external view returns (uint);
 
     // Epoch and periods
     // ~~~~~~~~~~~~~~~~~~
@@ -121,4 +125,6 @@ interface IElectionModule {
     function isElectionEvaluated() external view returns (bool);
 
     function getCandidateVotes(address candidate) external view returns (uint);
+
+    function getElectionWinners() external view returns (address[] memory);
 }
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -3,6 +3,12 @@ pragma solidity ^0.8.0;
 
 interface IElectionModule {
     // ---------------------------------------
+    // Initialization
+    // ---------------------------------------
+
+    function isElectionModuleInitialized() external bool returns (bool);
+
+    // ---------------------------------------
     // Owner functions
     // ---------------------------------------
 
--- a/packages/core-modules/contracts/interfaces/IElectionModule.sol
+++ b/packages/core-modules/contracts/interfaces/IElectionModule.sol
@@ -6,17 +6,17 @@ interface IElectionModule {
     // Initialization
     // ---------------------------------------
 
-    function isElectionModuleInitialized() external bool returns (bool);
-
-    // ---------------------------------------
-    // Owner functions
-    // ---------------------------------------
-
     function initializeElectionModule(
         uint64 nominationPeriodStartDate,
         uint64 votingPeriodStartDate,
         uint64 epochEndDate
     ) external;
+
+    function isElectionModuleInitialized() external bool returns (bool);
+
+    // ---------------------------------------
+    // Owner functions
+    // ---------------------------------------
 
     function adjustEpochSchedule(
         uint64 newNominationPeriodStartDate,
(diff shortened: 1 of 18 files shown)


packages/core-modules/contracts/submodules/election/ElectionTally.sol
```diff
@@ -0,0 +1,109 @@
...
+        ElectionData storage election = _getCurrentElection();
+        uint totalBallots = election.ballotIds.length;
+
+        uint firstBallotIndex = election.numEvaluatedBallots;
+
+        uint lastBallotIndex = firstBallotIndex + numBallots;
+        if (lastBallotIndex >= totalBallots) {
+            lastBallotIndex = totalBallots - 1;
+
+            election.evaluated = true;
```
mjlescano, line 23: Maybe this could be put after the call of `_evaluateBallotRange`, in the chance that that method is overridden with some custom logic that checks for `election.evaluated` and it wouldn't be true because the last ballot batch is still not processed.

packages/core-modules/contracts/submodules/election/ElectionTally.sol
```diff
@@ -0,0 +1,109 @@
...
+        uint ballotVotes = ballot.votes;
+        if (ballotVotes == 0) {
+            return;
+        }
+
+        address[] memory ballotCandidates = ballot.candidates;
+
+        uint numCandidates = ballotCandidates.length;
+        for (uint candidateIndex = 0; candidateIndex < numCandidates; candidateIndex++) {
+            address candidate = ballot.candidates[candidateIndex];
```
mjlescano, line 59: This should be using the previously assigned value:
```suggestion
            address candidate = ballotCandidates[candidateIndex];
```

packages/core-modules/contracts/submodules/election/ElectionTally.sol
```diff
@@ -0,0 +1,109 @@
...
+        uint numSeats
+    ) internal {
+        uint ballotVotes = ballot.votes;
+        if (ballotVotes == 0) {
+            return;
+        }
+
+        address[] memory ballotCandidates = ballot.candidates;
+
+        uint numCandidates = ballotCandidates.length;
```
mjlescano, line 57: Looks like is not necessary to assign the complete array from `storage` to `memory` on this use case:
```suggestion
        uint numCandidates = ballot.candidates.length;
```

mjlescano (approved):

packages/core-modules/contracts/submodules/election/ElectionTally.sol
ajsantander: Yeah, I was trying to save gas, but it makes too much noise, so I'll change it.

However, I'm going to try putting this in ElectionModule.sol, because it should be independent from the counting method, in the `evaluate()` function. Can you take a look?

mjlescano: Oh yeah, makes more sense now :)

packages/core-modules/contracts/interfaces/IElectionModule.sol
```diff
@@ -116,4 +122,13 @@ interface IElectionModule {
     function getBallotVotes(bytes32 ballotId) external view returns (uint);
 
     function getBallotCandidates(bytes32 ballotId) external view returns (address[] memory);
+
+    // Resolutions
+    // ~~~~~~~~~~~~~~~~~~
+
+    function isElectionEvaluated() external view returns (bool);
+
+    function getCandidateVotes(address candidate) external view returns (uint);
```
leomassazza, line 131: nit: getCandidateVotes will only apply to certain counting (identifying winners) strategies. Nothing that hurts right now but to keep in mind for the future

packages/core-modules/contracts/modules/ElectionModule.sol
```diff
@@ -78,29 +86,37 @@ contract ElectionModule is IElectionModule, ElectionSchedule, ElectionVotes, Own
...
     // ---------------------------------------
 
     function nominate() external override onlyInPeriod(ElectionPeriod.Nomination) {
-        SetUtil.AddressSet storage nominees = _getCurrentEpoch().nominees;
-
-        if (nominees.contains(msg.sender)) {
-            revert AlreadyNominated();
-        }
+        SetUtil.AddressSet storage nominees = _getCurrentElection().nominees;
+        if (nominees.contains(msg.sender)) revert AlreadyNominated();
```
leomassazza, line 112:
```suggestion

        if (nominees.contains(msg.sender)) revert AlreadyNominated();
```

packages/core-modules/contracts/modules/ElectionModule.sol
```diff
@@ -78,29 +86,37 @@ contract ElectionModule is IElectionModule, ElectionSchedule, ElectionVotes, Own
...
     }
 
     function withdrawNomination() external override onlyInPeriod(ElectionPeriod.Nomination) {
-        SetUtil.AddressSet storage nominees = _getCurrentEpoch().nominees;
-
-        if (!nominees.contains(msg.sender)) {
-            revert NotNominated();
-        }
+        SetUtil.AddressSet storage nominees = _getCurrentElection().nominees;
+        if (!nominees.contains(msg.sender)) revert NotNominated();
```
leomassazza, line 119:
```suggestion

        if (!nominees.contains(msg.sender)) revert NotNominated();
```

packages/core-modules/contracts/modules/ElectionModule.sol
```diff
@@ -111,9 +127,7 @@ contract ElectionModule is IElectionModule, ElectionSchedule, ElectionVotes, Own
 
     function elect(address[] calldata candidates) external override onlyInPeriod(ElectionPeriod.Vote) {
         uint votePower = _getVotePower(msg.sender);
-        if (votePower == 0) {
-            revert NoVotePower();
-        }
+        if (votePower == 0) revert NoVotePower();
```
leomassazza, line 130:
```suggestion

        if (votePower == 0) revert NoVotePower();
```

packages/core-modules/contracts/submodules/election/ElectionTally.sol
```diff
@@ -0,0 +1,97 @@
...
+
+            return;
+        }
+
+        // Otherwise see if there is a winning candidate with less votes to take its seat.
+        for (uint8 winnerPosition = 1; winnerPosition <= numSeats; winnerPosition++) {
+            address winner = winners.valueAt(winnerPosition);
+            uint winnerVotes = election.candidateVotes[winner];
+
+            if (candidateVotes > winnerVotes) {
```
leomassazza, line 90: Issue: This can replace the 2nd candidate and not the least voted. You need to find the least voted, not the first that received less votes than the new candidate in the list.

packages/core-modules/test/contracts/modules/ElectionModule.resolve.test.js
```diff
@@ -33,50 +44,202 @@ describe('ElectionModule (resolve)', () => {
...
-          });
-        });
+      it('shows that nominations exist', async function () {
+        assertBn.eq((await ElectionModule.getNominees()).length, 5);
       });
 
-      describe('when evaluating the epoch', function () {
-        before('evaluate', async function () {
-          await ElectionModule.evaluate();
+      describe('when entering the election period', function () {
```
leomassazza, line 64: add a test that consider the issue in _updateWinnerSet(). It can be simulated with 3 ballots and 2 seats. 
You need to get the ballots to check get in that order (Candidate1: 2 votes, Candidate2: 1 vote, Candidate3: 10 votes).  
Candidate3 will replace Candidate1 => the council will end as C3, C2 while C1 received more votes than C2

leomassazza (changes requested):
Some nits but the critical issue is the one in _updateWinnerSet()

packages/core-modules/contracts/interfaces/IElectionModule.sol
ajsantander: Hmm... ok. This will expose whatever counting algorithms store in the candidateVotes mapping, within the storage namespace. If this will not be used in other strategies, then this is not a nit, and we need to review our strategy.

packages/core-modules/contracts/modules/ElectionModule.sol
ajsantander: Committed

packages/core-modules/contracts/modules/ElectionModule.sol
ajsantander: Committed

packages/core-modules/contracts/modules/ElectionModule.sol
```diff
@@ -128,20 +142,23 @@ contract ElectionModule is IElectionModule, ElectionSchedule, ElectionVotes, Own
...
-    function evaluate() external override onlyInPeriod(ElectionPeriod.Evaluation) {
-        // TODO
+    function evaluate(uint numBallots) external override onlyInPeriod(ElectionPeriod.Evaluation) {
+        if (isElectionEvaluated()) revert ElectionAlreadyEvaluated();
+
+        _evaluateNextBallotBatch(numBallots);
 
-        _getCurrentEpoch().evaluated = true;
+        ElectionData storage election = _getCurrentElection();
+        if (election.numEvaluatedBallots == election.ballotIds.length) {
```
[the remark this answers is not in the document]
ajsantander: Again, if this is as you say, then this is not minor. Let's discuss.

packages/core-modules/contracts/submodules/election/ElectionTally.sol
ajsantander: Yep! You're totally right. Will fix 👍

ajsantander: Done. Can you give it another look?

packages/core-modules/test/contracts/modules/ElectionModule.resolve.test.js
ajsantander: Also done.

leomassazza (approved):

packages/core-modules/contracts/interfaces/IElectionModule.sol
leomassazza: I'd say, keep it that way until we implement another strategies. Maybe just overriding those functions on the other implementations is enough

packages/core-modules/contracts/modules/ElectionModule.sol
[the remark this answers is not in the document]
leomassazza: same as the response above.

PR #761, opened 2022-01-24
Add replace function to SetUtil

ajsantander:


packages/core-contracts/contracts/utils/SetUtil.sol
```diff
@@ -18,6 +18,14 @@ library SetUtil {
         remove(set.raw, bytes32(uint256(uint160(value))));
     }
 
+    function replace(
+        AddressSet storage set,
+        address value,
+        address newValue
+    ) internal {
+        replace(set.raw, bytes32(uint256(uint160(value))), bytes32(uint256(uint160(newValue))));
```
leomassazza, line 26: np: add a helper to transform from address to bytes32 (the chained casting)

ajsantander: Done https://github.com/Synthetixio/synthetix-v3/issues/763

PR #771, opened 2022-01-26
Let constants pass as safe

### LINKED ISSUE #702: Bug: constant variable declarations are seen as unsafe by the deployer and should not

leomassazza:
Fix #702


mjlescano (approved):

packages/deployer/internal/storage-ast-validator.js
```diff
@@ -122,6 +122,10 @@ class ModuleStorageASTValidator {
     // Look for state variable declarations
     for (const contractNode of candidates) {
       for (const node of findContractStateVariables(contractNode)) {
+        if (node.mutability === 'constant') {
```
ajsantander, line 125: Should we also apply this to "immutable"?

ajsantander (approved):
Looks good! Just commented on something else we might wanna do here.

PR #773, opened 2022-01-27
Deployer to check for missing interfaces

### LINKED ISSUE #736: Make sure all modules have interfaces

leomassazza:
Fix #736, #610

Looks huge, but is the addition of a new SUBTASK to the deployer that uses a new validator `packages/deployer/internal/interface-ast-validator.js`. 

The rest is adding interfaces to all the test modules so that the deployer approves the contracts.


packages/deployer/internal/interface-ast-validator.js
```diff
@@ -0,0 +1,66 @@
+const {
+  findContractDefinitions,
+  findContractDependencies,
+  findFunctions,
+} = require('@synthetixio/core-js/utils/ast/finders');
+const { contractIsModule } = require('../internal/contract-helper');
+
+class InterfaceCoverageASTValidator {
```
leomassazza, line 8: This is the core of the PR

packages/deployer/subtasks/validate-interfaces.js
```diff
@@ -0,0 +1,32 @@
...
+const { ContractValidationError } = require('../internal/errors');
+const { SUBTASK_VALIDATE_INTERFACES } = require('../task-names');
+
+subtask(SUBTASK_VALIDATE_INTERFACES).setAction(async (_, hre) => {
+  logger.subtitle('Validating all visible functions are defined in interfaces');
+
+  const { deployment } = hre.deployer;
+
+  const asts = mapValues(deployment.sources, (val) => val.ast);
+  const validator = new InterfaceCoverageASTValidator(asts);
```
mjlescano, line 14: The validator takes an object of AST nodes, but on the constructor it uses is with `Object.values(asts)`. And here, we use `mapValues`.

I think we could just do `Object.values(deployment.sources).map((val) => val.ast)` here, and remove the `Object.values(asts)` from inside the validator.

packages/deployer/subtasks/validate-interfaces.js
```diff
@@ -0,0 +1,32 @@
...
+
+  let errorsFound = [];
+  errorsFound.push(...validator.findFunctionsNotDefinedInInterfaces());
+
+  if (errorsFound.length > 0) {
+    errorsFound.forEach((error) => {
+      logger.error(error.msg);
+    });
+
+    errorsFound.map((err) => logger.debug(JSON.stringify(err, null, 2)));
```
mjlescano, line 24:
```suggestion
    for (const error of errorsFound) {
      logger.error(error.msg);
      logger.debug(JSON.stringify(error, null, 2))
    }
```

mjlescano, line 24:
```suggestion
    for (const error of errorsFound) {
      logger.error(error.msg);
      logger.debug(JSON.stringify(error, null, 2));
    }
```

packages/deployer/subtasks/validate-interfaces.js
```diff
@@ -0,0 +1,32 @@
...
+subtask(SUBTASK_VALIDATE_INTERFACES).setAction(async (_, hre) => {
+  logger.subtitle('Validating all visible functions are defined in interfaces');
+
+  const { deployment } = hre.deployer;
+
+  const asts = mapValues(deployment.sources, (val) => val.ast);
+  const validator = new InterfaceCoverageASTValidator(asts);
+
+  let errorsFound = [];
+  errorsFound.push(...validator.findFunctionsNotDefinedInInterfaces());
```
mjlescano, line 17: No need for a `let` here, and we could directly use the result:
```suggestion
  const errorsFound = validator.findFunctionsNotDefinedInInterfaces();
```

packages/deployer/internal/interface-ast-validator.js
```diff
@@ -0,0 +1,66 @@
...
+
+    for (const dependency of findContractDependencies(module.name, this.contractNodes)) {
+      if (dependency.contractKind != 'interface') {
+        continue;
+      }
+
+      interfacedFunctions.push(...findFunctions(dependency.name, this.contractNodes));
+    }
+
+    return interfacedFunctions;
```
mjlescano, line 62: IMHO, a more functional way of writing it would be clearer:
```suggestion
    return findContractDependencies(module.name, this.contractNodes).filter((dep) =>
      dep.contractKind === 'interface'
    );
```

mjlescano (approved):

packages/deployer/internal/interface-ast-validator.js
leomassazza: we are collecting the functions inside the for loop  `interfacedFunctions.push(...findFunctions(dependency.name, this.contractNodes))`  to collect them and if I'm right that is not possible inside the filter.

packages/deployer/subtasks/validate-interfaces.js
leomassazza: done

ajsantander (approved):

PR #785, opened 2022-02-01
Fix contracts dependencies find

### LINKED ISSUE #772: The way we are using the AST dependency finders is not compatible with having contracts on different files with the same name

mjlescano:
Closes #772

The main objective of this PR is to stop relying on using `"id"`'s for querying the AST tree which started to be unreliable on edge cases and change the `finders.js` interfaces to use fully qualified names.

### DIFF
--- a/packages/core-js/utils/ast/finders.js
+++ b/packages/core-js/utils/ast/finders.js
@@ -239,6 +239,17 @@ function _findCotractSourceByFullyQualifiedName(contractFullyQualifiedName, astN
   return { baseNode, contractNode };
 }
 
+function _findInheritedLocalNodeNames(contractNode, sourceUnitNode) {
+  return Array.from(findAll('InheritanceSpecifier', contractNode))
+    .map((inheritNode) => inheritNode.baseName.referencedDeclaration)
+    .map(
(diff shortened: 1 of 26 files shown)


packages/core-js/utils/ast/finders.js
```diff
@@ -9,42 +10,14 @@ function findContractDefinitions(astNode) {
...
-function findContractNodeWithName(contractName, astNodes) {
-  if (Array.isArray(astNodes)) {
-    for (const astNode of astNodes) {
-      const contractDefiniton = findContractNodeWithName(contractName, astNode);
-      if (contractDefiniton) return contractDefiniton;
-    }
-  }
-
-  for (const contractDefiniton of findAll('ContractDefinition', astNodes)) {
+function _findContractNodeWithName(contractName, sourceUnitNode) {
```
mjlescano, line 19: This function is not exposed anymore as it is not reliable to be used because there could be contracts with the same name on different SourceUnits, but the user could not notice that and re-add the bug we had before.

ajsantander (approved):
LGTM! Just pls review coverage so that there is no drop.

PR #791, opened 2022-02-02
Election module/cleanups

### LINKED ISSUE #738: ElectionModule final cleanups
### Tasks
- [x] Remove date vars in tests that don't need them
- [x] Move nomination logic to its own file?
- [x] Naming clean up: ensure no "voting", "nominating", "evaluating", etc in ElectionModule sources
- [x] No TODOs
- [x] Order of date parameters in schedule functions
- [x] Isn't it ugly that errors are all in ElectionBase?
- [x] Review gas efficiency. Don't destroy readability, but also don't be excessive
- [x] Nested structs at end
- [x] Consider getIdlePeriodDuration view
- [x] More tests for date changes with more extreme edge cases
- [x] Restore initializable checks in deployer (commented out until fixed)
- [x] Review reverts in view functions
- [x] Review external functions that just wrap an internal function without adding any value.
- [x] Review all uint sizes
- [x] Optimize packing in ElectionStorage
- [x] Clearly mark all contracts that are intended for override, and make sure they contain no generic logic or views
- [x] Review all comments
- [x] Only use internal in functions intended to be overriden in other contracts, and use virtual. Otherwise use private.
- [x] Simplify schedule. Multiple config functions are a bit confusing.
- [x] Review schedule adjustments again. Doesn't feel like the right solution.
- [x] Consider putting all test files in a folder and using shorter names
- [x] Make all views public? Some of them are already public.
- [x] Rename getCurrentPeriodType

ajsantander:
Fix #738

### DIFF
--- a/packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
+++ b/packages/core-contracts/contracts/token/ERC20HistoricalBalance.sol
@@ -63,24 +63,24 @@ contract ERC20HistoricalBalance is ERC20, IERC20HistoricalBalance, ERC20Historic
     ) internal override {
         super._transfer(from, to, amount);
 
-        _moveVotingPower(from, to, amount);
+        _moveBalance(from, to, amount);
     }
 
     function _mint(address to, uint256 amount) internal override {
         super._mint(to, amount);
 
-        _moveVotingPower(address(0), to, amount);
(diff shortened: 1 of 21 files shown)


leomassazza (approved):
Looks good.

PR #820, opened 2022-02-25
Add missing subtask name

leomassazza:


ajsantander:
It's a total mystery to me that hardhat was fetching and executing the task anyway.

I just checked in main. The name is undefined, but it executes the correct task correctly nonetheless 🤷

ajsantander (approved):

Issue #782, opened 2022-01-31
Constants in contracts. Apply same Should we also apply this to "immutable"?


leomassazza:
Should we also apply this to "immutable"?

_Originally posted by @ajsantander in https://github.com/Synthetixio/synthetix-v3/pull/771#discussion_r795630639_

PR #823, opened 2022-03-04
Warn about Nested structs in namespaces

### LINKED ISSUE #716: Deployer: Warn about nested structs
Create a validator to check if there are nested structs and warn about it (non-blocking)

leomassazza:
Fix #716 

This PR adds a validation for nested structs. The core of the validation is in `storage-ast-validator.js` where I add a new validator function named `findNestedStructDeclarations()`. It will check the type of the members of the storage struct to see if there is another struct inside.

### DIFF
--- a/packages/deployer/internal/storage-ast-validator.js
+++ b/packages/deployer/internal/storage-ast-validator.js
@@ -126,6 +126,23 @@ class ModuleStorageASTValidator {
     return errors;
   }
 
+  async findNestedStructDeclarations() {
+    const errors = [];
+
+    const structsMap = await buildContractsStructMap(this.contractNodes);
+
+    for (const structMap of structsMap) {
+      const structMembers = structMap.struct.members.filter((v) => v.type.startsWith('struct'));
+      if (structMembers.length > 0) {
+        errors.push({
+          msg: `Nested structs at ${structMap.contract.name}.${structMap.struct.name}`,
+        });
+      }
+    }
+
+    return errors;
+  }
+
   async findInvalidNamespaceMutations() {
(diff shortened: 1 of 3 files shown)


packages/deployer/test/unit/internal/storage-ast-validator.test.js
```diff
@@ -281,8 +309,7 @@ describe('internal/storage-ast-validator.js', function () {
       });
     });
 
-    // TODO enable this test when #782 is closed
-    describe.skip('valid immutable declaration', () => {
+    describe('valid immutable declaration', () => {
```
leomassazza, line 312: Re-enabling test. #782 was already merged.

packages/core-contracts/contracts/utils/SetUtil.sol
```diff
@@ -7,43 +7,43 @@ library SetUtil {
     // ----------------------------------------
 
     struct AddressSet {
-        Bytes32Set raw;
+        mapping(uint => Bytes32Set) raw;
```
leomassazza, line 10: TBH, I think using a mapping here is overkilling, but I couldn't find a way to allow this only lib to pass the tests without hardcoding it.

packages/deployer/internal/storage-ast-validator.js
```diff
@@ -126,6 +126,23 @@ class ModuleStorageASTValidator {
     return errors;
   }
 
+  async findNestedStructDeclarations() {
```
leomassazza, line 129: This is the core of the PR.

mjlescano (approved):

PR #840, opened 2022-03-09
Ability to change minimumActiveMembers setting

ajsantander:


packages/core-modules/contracts/storage/ElectionStorage.sol
```diff
@@ -28,6 +28,8 @@ contract ElectionStorage {
     struct ElectionSettings {
         // Number of council members in the next epoch
         uint8 nextEpochSeatCount;
```
ajsantander, line 30: It's ok to move things around because this hasn't hit production yet anywhere.

leomassazza (approved):
Love it!

PR #857, opened 2022-03-15
Election module refactor inheritance

### LINKED ISSUE #855: Refactor ElectionModule for better inheritance override

ajsantander:
Fix #855 

Moving ElectionModule inheritance a bit, so that it can be inherited from other packages in a way that does not import sample code.


leomassazza (commented):
LGTM (After merge approval)
In fact, I really like it. Actual module will be more readable.

PR #891, opened 2022-04-08
Add base cannonfile with router generation and proxy deployment

mjlescano:
Closes https://github.com/Synthetixio/gmv3-interface/issues/18

This PR adds a `governance-module` [cannon](https://usecannon.com/) build to facilitate development usage and integration with existing platforms.


packages/governance-modules/cannon.json
```diff
@@ -0,0 +1,11 @@
+{
+  "chains": [
+    {
+      "deploy": [
+        "synthetix-governance:0.0.1"
```
dbeal-eth, line 5: should this version be bounded to the `package.json` value?

packages/governance-modules/cannonfile.toml
```diff
@@ -0,0 +1,92 @@
+name = "synthetix-governance"
+description = "Governance cotracts for on chain voting"
+version = "0.0.1"
```
dbeal-eth, line 3: same for binding to the package.json value

if we want to keep cannon versino separate from package for the time being, then please disregard this comment.

packages/governance-modules/cannonfile.toml
```diff
@@ -0,0 +1,92 @@
...
+defaultValue = "SGT"
+
+[setting.first_council_member]
+defaultValue = "0xe3B0C44298FC1C149AfBF4C8996fb92427aE41E4"
+
+[setting.epoch_duration]
+defaultValue = 90 # days
+
+[setting.voting_period_duration]
+defaultValue = 7 # days
```
dbeal-eth, line 22: great settings! :+1:

packages/governance-modules/cannonfile.toml
```diff
@@ -0,0 +1,92 @@
...
+[contract.upgrade_module]
+artifact = "contracts/modules/UpgradeModule.sol:UpgradeModule"
+step = 1
+
+[run.generate_router]
+exec = "scripts/generate-router.js"
+func = "generateRouter"
+args = [
+  "contracts/modules/ElectionModule.sol:ElectionModule",
+  "<%= outputs.self.contracts.election_module.address %>",
```
dbeal-eth, line 41: srsly this works? again, great job!

dbeal-eth: wonder if we could mak ethe params even more streamlined though. thoughts?

dbeal-eth (approved):
this is awesome!

packages/governance-modules/cannonfile.toml
mjlescano: hmm, my original idea was to only give it 'outputs.self.contracts', so we would avoid repetition, but we cannot do it because cannon doesn't allow us to put objects thru templating. Maybe we could ask the mainteners for that feature? 😬

dbeal-eth: hmm maybe we should ask :thinking:

packages/governance-modules/cannon.json
mjlescano: Yess, updating.

packages/governance-modules/cannonfile.toml
mjlescano: I think we should use another version for now, for the case that we build the same version of `governance-modules` with a different cannon version (which we are probably going to publish a lot on the following weeks)

packages/governance-modules/cannonfile.toml
mjlescano: Ended up using a literal JSON on a multi line string, I think it works great :)

packages/governance-modules/cannon.json
mjlescano: Responded on next comment 👇

leomassazza (approved):
Everything looks good though I didn't check the cannonfile.toml

PR #895, opened 2022-04-13
Bootstrap AccountModule

### LINKED ISSUE #893: Bootstrap AccountModule with permission modifier

leomassazza:
Fix #893  
This PR adds the AccountModule and Account as ERC721 satellite (just bootstrap) ~and implements the delegation logic via some internal functions at the moment and a modifier.~


packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -0,0 +1,16 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@synthetixio/core-contracts/contracts/interfaces/ISatelliteFactory.sol";
+
+interface IAccountModule is ISatelliteFactory {
```
leomassazza, line 6: At the moment I think it will be a SatteliteFactory, but still TBD.

packages/synthetix-main/contracts/modules/SynthsModule.sol
```diff
@@ -100,7 +100,7 @@ contract SynthsModule is ISynthsModule, OwnableMixin, SynthsStorage, Initializab
         ISynth(getSynthImplementation()).setNewAuthorized(authorized);
     }
 
-    function getBeacon() external view override returns (address) {
+    function getSynthBeacon() external view override returns (address) {
```
leomassazza, line 103: Renamed to prevent future naming collisions

packages/synthetix-main/contracts/storage/AccountStorage.sol
```diff
@@ -0,0 +1,19 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@synthetixio/core-contracts/contracts/satellite/SatelliteFactory.sol";
+
+contract AccountStorage {
+    struct AccountStore {
+        bool initialized;
+        SatelliteFactory.Satellite account;
+        mapping(uint256 => mapping(address => uint16)) accountDelegations;
```
leomassazza, line 10: This mapping contains the account delegations. Each account can be delegated to more than one address and each address can have one or several atomic permissions (mapped into a uint16, 1 bit per permission). 

I'm tempted to up the limit from uint16 to uint64 or something like that to have room for more permissions, but up to now only 7 are defined.

packages/synthetix-main/contracts/submodules/account/AccountBase.sol
```diff
@@ -0,0 +1,112 @@
...
+import "../../storage/AccountStorage.sol";
+
+/// @dev Common utils, errors, and events to be used by any contracts that conform the AcccountModule
+contract AccountBase is ERC721, AccountStorage, InitializableMixin {
+    // ---------------------------------------
+    // Constants
+    // ---------------------------------------
+
+    /// @dev use permissions packed in an uint16 (16 permissions). Set the values right shifting uint(1) from 0 to 15
+    uint16 private constant _PERMISSION_DELEGATE = uint16(1);
```
leomassazza, line 15: one permission per atomic operation. Each in one bit position of the uint16

packages/synthetix-main/contracts/submodules/account/AccountBase.sol
```diff
@@ -0,0 +1,112 @@
...
+            packedPermissions = packedPermissions | permissions[i];
+        }
+
+        _accountStore().accountDelegations[account][authorized] = packedPermissions;
+
+        emit PermissionsSet(account, authorized, packedPermissions);
+    }
+
+    /// @dev Used to allow certain functions to operate on the account if owner or authorized to that operation
+    modifier onlyOwnerOrAuthorized(uint256 accountId, uint16 permission) {
```
leomassazza, line 105: This is the modifier that is going to be used on each operation that can be delegated or requires the account owner

packages/synthetix-main/contracts/token/SNXToken.sol
```diff
@@ -8,8 +8,6 @@ import "@synthetixio/core-contracts/contracts/token/ERC20.sol";
 import "@synthetixio/core-contracts/contracts/errors/AccessError.sol";
 
 contract SNXToken is ERC20, UUPSImplementation, Ownable, Authorizable {
-    address private _minterSystem;
```
leomassazza, line 11: duh

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
leomassazza: needs to be a SatelliteFactory

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -0,0 +1,16 @@
...
+    function initializeAccountModule() external;
+
+    function isAccountModuleInitialized() external view returns (bool);
+
+    function upgradeAccountImplementation(address newAccountImplementation) external;
+
+    function getAccountAddress() external view returns (address);
+
+    function getAccountModuleSatellites() external view returns (Satellite[] memory);
```
dbeal-eth, line 15: im assuming its returning an array only because that is how general satellite implementations work? Normally the synthetix system should only need a single satellite for purposes of the account token, but I may not understand how satellites work in this case or what the purpose should be here

packages/synthetix-main/contracts/interfaces/IDelegation.sol
```diff
@@ -0,0 +1,15 @@
...
+interface IDelegation {
+    enum Permission {
+        Owner,
+        Stake,
+        Unstake,
+        Mint,
+        Burn,
+        ClaimRewards,
+        ManageLocking,
+        Delegate
```
dbeal-eth, line 13: permission to delegate is basically all the permission needed to be an owner (because you can just delegate permissions to yourself), right?

packages/synthetix-main/contracts/interfaces/IDelegation.sol
```diff
@@ -0,0 +1,15 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IDelegation {
+    enum Permission {
```
dbeal-eth, line 5: im assuming we can just add more items to the enum for future upgrades. Is there a check in the deployment tool that verifies that enums are not updated incorrectly?

packages/synthetix-main/contracts/modules/AccountModule.sol
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    function isAccountModuleInitialized() external view override returns (bool) {
+        return _isInitialized();
+    }
+
+    function initializeAccountModule() external override onlyOwner onlyIfNotInitialized {
+        AccountModuleStore storage store = _accountModuleStore();
+
+        Account firstAccountImplementation = new Account();
```
dbeal-eth, line 27: will there ever be anything beyond the "first account" implementation?

packages/synthetix-main/contracts/satellites/account/AccountBase.sol
```diff
@@ -0,0 +1,121 @@
...
+
+    function _hasPermission(
+        uint256 account,
+        address authorized,
+        Permission permission
+    ) internal view returns (bool) {
+        return _accountStore().delegatedPermissions[account][authorized] & _getPermissionMask(permission) != 0;
+    }
+
+    function _grantPermission(
```
dbeal-eth, line 49: what is the purpose of having these extra functions if a user/web ui can just call `_setPermissions` to accomplish same task?

dbeal-eth (approved):
just minor thoughts from me. Another explaination of what satellite does for us.

Also, we should really consider making the account fully `ERC-721` compatible. The Permissions system is nice, but we are already closely tracing what an ERC-721 token would look like, so why don't we just implement it all the way?

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
leomassazza: Exactly, it is an array for all kind of satellites so that the CLI let's you interact with this kind of Modules (only one satellite) or with multiple satellites

packages/synthetix-main/contracts/interfaces/IDelegation.sol
leomassazza: you are right... I added it before checking the UI where I saw the "owner"

packages/synthetix-main/contracts/interfaces/IDelegation.sol
leomassazza: not yet

packages/synthetix-main/contracts/modules/AccountModule.sol
leomassazza: yes, the AccountModule is the "implementation" manager. And allows the system to be upgraded with another implementation of Account contract

packages/synthetix-main/contracts/satellites/account/AccountBase.sol
```diff
@@ -0,0 +1,121 @@
...
+pragma solidity ^0.8.0;
+
+import "@synthetixio/core-contracts/contracts/token/ERC721.sol";
+import "@synthetixio/core-contracts/contracts/utils/AddressUtil.sol";
+import "@synthetixio/core-contracts/contracts/initializable/InitializableMixin.sol";
+import "../../storage/AccountStorage.sol";
+import "../../interfaces/IDelegation.sol";
+
+/// @dev Common utils, errors, and events to be used by any contracts that conform the AcccountModule
+contract AccountBase is ERC721, AccountStorage, IDelegation, InitializableMixin {
```
leomassazza, line 11: This is where we declare the Account as ERC721

packages/synthetix-main/contracts/satellites/account/AccountBase.sol
leomassazza: I added both because the interface wasn't declared yet to support single call to set all the permissions in one step or single permission granting/revoking operations.
If they are not required I'll delete the single perm operations.

PR #921, opened 2022-05-05
fix generation of router breaking when two modules have same name

dbeal-eth:
LMK how I did

fixes https://github.com/Synthetixio/synthetix-v3/issues/792

notes:
* test was added to verify this specific case in the future
* the name might look kind of long/ugly, lmk if its unacceptable, but people
probably won't be looking at the router very often so its not issue i'm guessing


mjlescano (approved):

dbeal-eth:
this is no longer relevant with the router being migrated to typescript, and many other changes.

PR #929, opened 2022-05-10
Adds withdrawVote()

### LINKED ISSUE #928: Implement withdrawVote()
Atm, users are able to change their vote, but not withdraw it. There should be a `withdrawVote()` function that simply removes the user's previous vote, if it exists.

ajsantander:
Closes #928

### DIFF
--- a/packages/core-modules/contracts/submodules/election/ElectionBase.sol
+++ b/packages/core-modules/contracts/submodules/election/ElectionBase.sol
@@ -31,7 +31,7 @@ contract ElectionBase is ElectionStorage, InitializableMixin {
     error NotNominated();
     error NoCandidates();
     error NoVotePower();
-    error HasNotVoted();
+    error VoteNotCasted();
     error DuplicateCandidates();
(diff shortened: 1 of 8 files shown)


packages/core-modules/contracts/submodules/election/ElectionBase.sol
```diff
@@ -31,6 +31,7 @@ contract ElectionBase is ElectionStorage, InitializableMixin {
     error NotNominated();
     error NoCandidates();
     error NoVotePower();
+    error HasNotVoted();
```
mjlescano, line 34: Not sure about this, but this error looks different to the other ones, maybe something like `VoteNotCasted`, `NoCastedVote`

andytcf (approved):

packages/core-modules/contracts/submodules/election/ElectionBase.sol
ajsantander: Ok, will rename to  `VoteNotCasted`.

PR #963, opened 2022-05-23
Blocking wrong initializer on EM

### LINKED ISSUE #962: Revert if user calls wrong initializer in EM

ajsantander:
Fix #962

`synthetix-governance` adds adds a parameter to the initializer. Still, nothing stops devs from calling the wrong initializer. Even though that situation could be recoverable, this PR enforces it by blocking access to the base initializer.


mjlescano (approved):

leomassazza:
Post merge PR: Approved LGTM

PR #987, opened 2022-05-25
Fix CLI tests

mjlescano:


packages/cli/scripts/test.sh
```diff
@@ -11,6 +11,9 @@ npx hardhat deploy --network local --clear --quiet --no-confirm --instance test
 # Run tests
 cd ../../..
 npx mocha
+test_exit_code=$?
 
 # Kill node
 pkill node
+
+exit "$test_exit_code"
```
mjlescano, line 19: This fixes the error from the tests, which was causing for the Github Action no never fail, regardless if the tests failed.

packages/cli/test/setup.js
```diff
@@ -0,0 +1,4 @@
+const chalk = require('chalk');
+
+// Disable logging coloring on all environments
+chalk.level = 0;
```
mjlescano, line 4: Unit tests were not working on local because they were printing colors, and the assertions didn't work.

packages/deployer/subtasks/load-deployment.js
```diff
@@ -25,6 +25,10 @@ subtask(SUBTASK_LOAD_DEPLOYMENT, 'Loads deployment artifacts for a particular in
     const previousDeploymentFile =
       deploymentFiles.length > 1 ? deploymentFiles[deploymentFiles.length - 2] : null;
 
+    if (!currentDeploymentFile) {
+      throw new Error(`Could not find deployment files with ${JSON.stringify(info)}`);
+    }
```
mjlescano, line 30: Just added a more clear error when missing a deployment file, this would have saved me some time.

PR #1015, opened 2022-06-06
Add MulDivUp Math util 

leomassazza:


packages/core-contracts/test/contracts/utils/MathUtil.test.js
```diff
@@ -1,7 +1,12 @@
 const { ethers } = hre;
 const assertBn = require('@synthetixio/core-js/utils/assertions/assert-bignumber');
+const assertRevert = require('@synthetixio/core-js/utils/assertions/assert-revert');
 const { bnSqrt } = require('@synthetixio/core-js/utils/ethers/bignumber');
 
+function e(base, exp) {
+  return ethers.BigNumber.from(base).mul(ethers.BigNumber.from(10).pow(exp));
```
dbeal-eth, line 7: this function name `e` seems confusing since in mathematics, `e` corresponds to Euler's number which is used for logarithmic functions and other purposes which could conceivably coincide with this.

Instead, we could name it `scale` or `s` would be better.

Better yet, we could use the `@synthetixio/wei` library which is specifically designed for these operations. For example Line 67 would change to:

```
await assertMulDivDown({ x: wei(250, 25), y: wei(50, 25), denominator: wei(100, 25) }, wei(125, 25));
```

very similar to what you are already doing. and `Wei` converts directly to big number with `n.toBN()`

dbeal-eth (changes requested):

packages/synthetix-main/contracts/modules/MulticallModule.sol
```diff
@@ -11,7 +11,8 @@ contract MulticallModule is IMulticallModule {
             (bool success, bytes memory result) = address(this).delegatecall(data[i]);
 
             if (!success) {
-                // Next 5 lines from https://ethereum.stackexchange.com/a/83577
+                // Next 6 lines from https://ethereum.stackexchange.com/a/83577
+                // solhint-disable-next-line reason-string
```
leomassazza, line 15: Fix on the go

packages/core-contracts/test/contracts/utils/MathUtil.test.js
dbeal-eth: to clarify, the purpose of the other bignumber library is only for complex math functions. it does not store or use Big.js for any other purpose.

dbeal-eth (approved):

PR #1019, opened 2022-06-07
Mint next account

noahlitvin:
In general, we'd like accounts to be generated with incrementing IDs. We started by allowing accounts to be created at any ID, mostly because in the case of a multicall, you're going to want to be able to queue up later transactions to reference the account ID you're creating. But if two accounts were created on the same block, the account creation would revert for the second, leading to a poor user experience.

This function allows our peripheral `Onboarding` contract (as well as a UI for just creating an account token without onboarding) to create an account with an incrementing ID with no risk of reversion. Open to renaming to something other than `mintNext`.

### DIFF
--- a/packages/synthetix-main/contracts/interfaces/IAccountToken.sol
+++ b/packages/synthetix-main/contracts/interfaces/IAccountToken.sol
@@ -5,6 +5,9 @@ import "@synthetixio/core-contracts/contracts/interfaces/IERC721.sol";
 
 /// @title NFT token identifying an Account
 interface IAccountToken is IERC721 {
-    /// @notice mints a new token (NFT) with the "requestedAccountId" id owned by "owner". It can ol=nly be called by the system
+    /// @notice mints a new token (NFT) with the "requestedAccountId" id owned by "owner".
     function mint(address owner, uint requestedAccountId) external;
+
+    /// @notice mints a new token (NFT) with at the next available id owned by "owner".
+    function mintNext(address owner) external returns (uint256 accountId);
 }
--- a/packages/synthetix-main/contracts/satellites/AccountToken.sol
+++ b/packages/synthetix-main/contracts/satellites/AccountToken.sol
@@ -47,12 +47,21 @@ contract AccountToken is IAccountToken, ERC721, AccountTokenStorage, Initializab
     // ---------------------------------------
     // Mint/Transfer
     // ---------------------------------------
-    function mint(address owner, uint256 accountId) external override {
-        _mint(owner, accountId);
+    function mint(address _owner, uint256 accountId) public override {
+        _mint(_owner, accountId);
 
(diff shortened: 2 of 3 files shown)


packages/synthetix-main/contracts/interfaces/IAccountToken.sol
```diff
@@ -5,6 +5,9 @@ import "@synthetixio/core-contracts/contracts/interfaces/IERC721.sol";
 
 /// @title NFT token identifying an Account
 interface IAccountToken is IERC721 {
-    /// @notice mints a new token (NFT) with the "requestedAccountId" id owned by "owner". It can ol=nly be called by the system
+    /// @notice mints a new token (NFT) with the "requestedAccountId" id owned by "owner".
     function mint(address owner, uint requestedAccountId) external;
```
leomassazza, line 9:
```suggestion
    function mint(address accountOwner, uint requestedAccountId) external;
```

packages/synthetix-main/contracts/interfaces/IAccountToken.sol
```diff
@@ -5,6 +5,9 @@ import "@synthetixio/core-contracts/contracts/interfaces/IERC721.sol";
...
+
+    /// @notice mints a new token (NFT) with at the next available id owned by "owner".
+    function mintNext(address owner) external returns (uint256 accountId);
```
leomassazza, line 12:
```suggestion
    function mintNext(address accountOwner) external returns (uint256 accountId);
```

packages/synthetix-main/contracts/satellites/AccountToken.sol
```diff
@@ -47,12 +47,21 @@ contract AccountToken is IAccountToken, ERC721, AccountTokenStorage, Initializab
...
 
-        IAccountModule(_accountStore().mainProxy).transferAccount(owner, accountId);
+        IAccountModule(_accountStore().mainProxy).transferAccount(accountOwner, accountId);
 
-        emit AccountMinted(owner, accountId);
+        emit AccountMinted(accountOwner, accountId);
+    }
+
+    function mintNext(address accountOwner) external override returns (uint256 accountId) {
+        while (_exists(_accountStore().recentIdUsed)) {
```
leomassazza, line 59: you can check before if the number of tokens created is less than recentIdUsed and fast forward to it, also write on the while loop to an internal variable and at the end go to the store since it's cheaper. 
i.e.

```uint checkId =  _accountStore().recentIdUsed;
if (checkId  < _accountStore().balance) {
   checkId = _accountStore().balance;
}
while (_exists(checkId) {
  checkId++;
}
accountId = checkId;
_accountStore().recentIdUsed = accountId;
```
and you need to add the balance in the store (`uint256 balance,`) and an increment to the public mint function (`_accountStore().balance++;`)

leomassazza (approved):
Added a suggestion (optional) and committed a small change (adjust variable names in the Interface to match the contract).

PR #1024, opened 2022-06-14
Split Fund Module

leomassazza:
Splits fund module in 3 separate concerns:
- Fund Creation, Ownership and Admin
- SCCP params
- Fund Vaults

Added some cleanup also in this PR: 
- Removes the fundToken since the Id can be managed directly without an NFT
- Cleanup (remove old example modules)


packages/synthetix-main/contracts/interfaces/IFundSCCP.sol
```diff
@@ -0,0 +1,20 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @title Module for managing preferred and approved funds via SCCPs
+interface IFundSCCP {
```
dbeal-eth, line 5: maybe we could do better for naming this file/interface

packages/synthetix-main/contracts/interfaces/IFundVault.sol
```diff
@@ -0,0 +1,78 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./IFundVaultStorage.sol";
+
+/// @title Module for managing funds and assignments per account
+interface IFundVault is IFundVaultStorage {
```
dbeal-eth, line 7: i think `IVault` is fine here np

packages/synthetix-main/contracts/interfaces/IFundVault.sol
```diff
@@ -0,0 +1,78 @@
...
+    function delegateCollateral(
+        uint fundId,
+        uint accountId,
+        address collateralType,
+        uint amount,
+        uint leverage
+    ) external;
+
+    /// @notice mints sUSD for a fund/account from a collateralType. if CRatio is valid
+    function mintsUSD(
```
dbeal-eth, line 21: im realizing now it might just be better to call this `withdrawLoan` or something like that, `sUSD` is a terrible thing to have in function names unfortunately

packages/synthetix-main/contracts/modules/SUSDTokenModule.sol
```diff
@@ -0,0 +1,74 @@
...
+
+        UUPSProxy sUSDTokenProxy = new UUPSProxy(address(firstSUSDTokenImplementation));
+
+        address sUSDTokenProxyAddress = address(sUSDTokenProxy);
+        SUSDToken sUSDToken = SUSDToken(sUSDTokenProxyAddress);
+
+        sUSDToken.nominateNewOwner(address(this));
+        sUSDToken.acceptOwnership();
+        sUSDToken.setNewAuthorized(address(this));
+        sUSDToken.initialize("SyntUSD Token", "sUSD", 18);
```
dbeal-eth, line 44: typo

dbeal-eth: also I the official title should probably be something like "Synthetic USD Token v3"

packages/synthetix-main/contracts/token/SUSDToken.sol
```diff
@@ -20,7 +21,11 @@ contract SNXToken is ERC20, UUPSImplementation, Ownable, Authorizable {
         _upgradeTo(newImplementation);
     }
 
-    function mint(address recipient, uint256 amount) external onlyAuthorized {
-        _mint(recipient, amount);
+    function burn(address to, uint256 amount) external override onlyAuthorized {
```
dbeal-eth, line 24: I know `onlyAuthorized` should be very secure but we might want to tighten this all the way down to `onlyOwner`. That will make it harder for an accident or attack somehow granting access to mint sUSD to an account outside of the core mvp system, and makes the governance process more clear.

dbeal-eth (approved):
great!

packages/synthetix-main/contracts/modules/SUSDTokenModule.sol
leomassazza: Fixed

packages/synthetix-main/contracts/token/SUSDToken.sol
leomassazza: Fixed

packages/synthetix-main/contracts/interfaces/IFundVault.sol
leomassazza: We might have vaults also for markets later. And also I'd like to have all Fund sections close and findable by name

packages/synthetix-main/contracts/interfaces/IFundSCCP.sol
leomassazza: Fixed. Changed to FundConfiguration

packages/synthetix-main/contracts/interfaces/IFundVault.sol
leomassazza: I'll keep that way until we find/choose a better option. We were discussing it a little bit with Noah and couldn't find a better option.

PR #1028, opened 2022-06-15
Add users debts for mainnet and staging

mjlescano:


packages/treasury-council/package.json
```diff
@@ -19,7 +19,6 @@
     "@nomiclabs/hardhat-ethers": "2.0.5",
     "hardhat": "2.9.3",
     "hardhat-contract-sizer": "2.1.1",
-    "solidity-coverage": "0.7.17",
-    "synthetix": "2.64.2"
```
mjlescano, line 23: Could not find anywhere  that the `synthetix` dep is being used, removing.

PR #1031, opened 2022-06-17
Add common config package

mjlescano:
This PR abstract all the common configs from hardhat projects and puts it on the `common-config` package to be easily reausable.

## Advantages
* It removes **a lot** of config repetition
* Instead of having to maintain repeated `.env` files on each package, we can just have a single `common-config/.env` file that is going to be used by all packages.
* If we want to update the Solidity version we are using, we just have to edit `common-config/hardhat.config.js` instead of updating each package file.
* Dependencies for hardhat projects are only defined once at `common-config/package.json`. This makes it a lot easier to update `hardhat` for example.


packages/spartan-council/package.json
```diff
@@ -14,12 +14,5 @@
...
   "author": "",
-  "license": "MIT",
-  "devDependencies": {
-    "@nomiclabs/hardhat-ethers": "2.0.5",
-    "hardhat": "2.9.3",
-    "hardhat-contract-sizer": "2.1.1",
-    "solidity-coverage": "0.7.17",
-    "synthetix": "2.64.2"
-  }
+  "license": "MIT"
```
leomassazza, line 17: Do we need the license here also (repeat for all the inherited package.json)

PR #1034, opened 2022-07-01
Updates to the validate council script

ajsantander:
Updates the validate council script to reflect the desired state after the first elections were resolved.


mjlescano (changes requested):
@ajsantander we already had this PR open: https://github.com/Synthetixio/synthetix-v3/pull/1032 with some changes and refactors already in place, so, I would prefer to merge that one with these changes included.

I already migrated the new data validations (including members, etc), just need to update the Date parsing stuff which is super nice and will be ready :)

PR #1039, opened 2022-07-11
(252) added subgraph workspace

fritzschoff:
close #252


packages/subgraphs/governance/.gitignore
```diff
@@ -0,0 +1,64 @@
...
+*.tgz
+
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+
+# next.js build output
+.next
```
mjlescano, line 64: IMHO, all the templated stuff is not necessary, we should have only the things we need as it makes this file a lot easier to maintain :)

mjlescano (approved):

PR #1040, opened 2022-07-18
Add erc721 enumerability

sunnyvempati:
- add ability to get owner's token Ids by using `ERC721Enumerable`.  
- this is required for frontend's UX when owners who have accounts visit the app.


packages/core-contracts/contracts/token/ERC721Storage.sol
```diff
@@ -6,10 +6,14 @@ contract ERC721Storage {
         string name;
         string symbol;
         string baseTokenURI;
+        uint256[] allTokens;
```
leomassazza, line 9: Add all enumerable storage related (all the additions) to a specific storage i.e. ERC721EnumerableStorage.sol so that it doesn't add complexity to non-enumerable ERC721's

leomassazza: Also, this change won't pass the deployer since you are adding variables in the middle of the storage struct (if it passed, let us know, it shouldn't).
Any new variable should be added at the bottom of the storage struct. 

On any upgrade this contract contract will have the state data corrupted. Let's jump in a call if you need more info about that.

leomassazza (changes requested):
Great job, mate! Just some comments related on how v3 works and naming.

Also, we'll need to add some tests to cover the new functionality before going back to main. Not needed right now, but helpful

packages/core-contracts/contracts/token/ERC721Enumerable.sol
```diff
@@ -0,0 +1,141 @@
...
+    function _initialize(
+        string memory tokenName,
+        string memory tokenSymbol,
+        string memory baseTokenURI
+    ) internal virtual override {
+        super._initialize(tokenName, tokenSymbol, baseTokenURI);
+        ERC721EnumerableStore storage store = _erc20EnumerableStore();
+        if (store.allTokens.length > 0) {
+            revert InitError.AlreadyInitialized();
+        }
```
sunnyvempati, line 25: @leomassazza can you double check this?  

calling `super` here and calling the erc721 initialializer; and i'm ok if the store is reinitialized as long as there's no tokens created so thats the check on line 23.  

please suggest or shed light on anything im missing here.  thanks!

leomassazza: initialization should only happens once. super._initialize will revert if tokenName, tokenSymbol or baseTokenURI was already set.

Your double checking makes sense here but not likely to revert

leomassazza: you can simplify a little the code since store is not used anymore in that function and just get the allTokens directly:
if (_erc20EnumerableStore().allTokens.length >0) { ... }

leomassazza (approved):
LGTM, just a np

packages/core-contracts/contracts/token/ERC721Enumerable.sol
sunnyvempati: makes sense! thanks for feedback :)

PR #1041, opened 2022-07-25
cannon testing update + tests are ts, rest of synthetix-main merge

dbeal-eth:
major summary of changes (reccomended to read through them in this order):
* updated cannonfile to deploy 4 systems instead of 1: account nft, usd token, esnx token, and core system (includes funds, vaults, marketmanager, etc.). All other systems are then connected using the newly created "AssociatedSystemsModule"
* created `AssociatedSystemsModule` which is sort of like a modernized version of the addressresolver for v3. You specify an implementation address, (which can be a *router* deployed by the deployer with cannon) and then it will deploy a proxy to officially connect it to the system. Supports doing the upgrades in the same call. also supports adding unmanaged contracts (ex. when we deploy to prod, using the old SNX contract)
* change deployer ability to *not* deploy proxies. This should be done with cannon now
* change deployer to take in `modules` argument (positional), which has the ability to specify which modules to deploy. This is used by cannon to build subsets of the module directory for the 4 different systems specified above.
* add `TokenModule` and `NftModule` to `core-modules` which implements everything needed for a basic module as the title implies
* remove checks for initialization modules (though not the initialization module itself), since initialization will be a much more customized process going forward and will be done by just implementing "isInitialized" for your module.
* Implement SIP-305, which mostly entails:
  * adding "rewards" concept to Vaults. More than one reward can be added, and any rewards can be added with the consent of the fund owner. This will be used to distribute eSNX
  * create eSNX token, which implements the RewardsDistributor interface from the vault mentioned in previous bullet. When told to distribute a reward, it mints tokens and sends directly to the specified user account
  * create `redeemRewards` function in `CollateralModule` which takes in "rewardToken" (which is, eSNX) and mints SNX.
  * add logic to process the existance of rewards as part of your collateral in an account with the curves library from earlier

notes:
* deleted a bunch of logic for checking initialization behavior, since initialization will now be a much more case-by-case setup. I think this is for the best 
*


packages/synthetix-main/contracts/storage/FundVaultStorage.sol
```diff
@@ -31,6 +33,27 @@ contract FundVaultStorage is IVaultModuleStorage {
         mapping(uint256 => SetUtil.Bytes32Set) liquidityItemsByAccount;
         /// @dev minted USD
         mapping(uint256 => uint256) usdByAccount;
+        /// @dev rewards
+        RewardDistribution[] rewards;
+    }
+
+    struct RewardDistribution {
```
leomassazza, line 40: Rewards Distribution can be moved to a specific module/storage pair and if something is needed from the vault (or viceversa) use mixins in order to maintain single responsibilities on modules (and more important) storage

packages/synthetix-main/contracts/modules/core/AssociatedSystemsModule.sol
```diff
@@ -0,0 +1,53 @@
...
+import "@synthetixio/core-contracts/contracts/ownership/OwnableMixin.sol";
+import "@synthetixio/core-contracts/contracts/proxy/UUPSProxy.sol";
+import "../../interfaces/IAssociatedSystemsModule.sol";
+import "../../storage/AssociatedSystemsStorage.sol";
+
+import "@synthetixio/core-contracts/contracts/interfaces/IUUPSImplementation.sol";
+import "@synthetixio/core-modules/contracts/interfaces/IOwnerModule.sol";
+import "@synthetixio/core-modules/contracts/interfaces/ITokenModule.sol";
+
+contract AssociatedSystemsModule is IAssociatedSystemsModule, OwnableMixin, AssociatedSystemsStorage {
```
leomassazza, line 13: Since AssociatedModules replaces the satellites, it should be a core-module (with its storage and mixin)

packages/synthetix-main/contracts/modules/account/AccountTokenModule.sol
```diff
@@ -0,0 +1,6 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@synthetixio/core-modules/contracts/modules/NftModule.sol";
+
+contract AccountTokenModule is NftModule {}
```
leomassazza, line 6: Please, apply this change taking into account @sunnyvempati 's PR to change the Account NFT as ERC721_Enumerable

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -23,70 +25,15 @@ contract AccountModule is IAccountModule, OwnableMixin, AccountRBACMixin, Initia
...
-        override
-        onlyOwner
-        onlyIfInitialized
-    {
-        AccountToken(getAccountAddress()).upgradeTo(newAccountTokenImplementation);
-    }
-
-    function getAccountAddress() public view override returns (address) {
-        return _accountModuleStore().account.deployedAddress;
+    function getAccountAddress() public view override returns (INftModule) {
```
leomassazza, line 28: np: rename to be transparent on what is the returned value (not the address, but the NFT itself)

packages/synthetix-main/contracts/modules/core/AssociatedSystemsModule.sol
```diff
@@ -0,0 +1,53 @@
...
+    string public constant KIND_ERC20 = "erc20";
+    string public constant KIND_ERC721 = "erc721";
```
leomassazza, line 15: Kind ERC721 (and KIND_OTHER) are unused. In fact in the code, Account that is an NFT is registered as ERC20

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -23,70 +25,15 @@ contract AccountModule is IAccountModule, OwnableMixin, AccountRBACMixin, Initia
...
+        return _getNft(ACCOUNT_SYSTEM);
```
leomassazza, line 29: AccountTokenModule originaly created a Satellite (AccountToken) that is a specialized NFT that will use the transferAccount hook. 

I didn't find that hook call in the AccountTokenModule (maybe I'm missing something).

package.json
```diff
@@ -14,10 +14,11 @@
   "license": "MIT",
   "homepage": "https://github.com/Synthetixio/synthetix-v3#readme",
   "scripts": {
+    "build": "npm run --workspaces --if-present build",
     "test": "npm run --workspaces --if-present test",
     "coverage": "npm run --workspaces --if-present coverage",
-    "lint:js": "prettier --check '**/*.js' && eslint '**/*.js'",
-    "lint:js:fix": "prettier --write '**/*.js' && eslint --fix '**/*.js'",
+    "lint:js": "eslint '**/*.ts'",
+    "lint:js:fix": "eslint --fix '**/*.ts'",
```
ajsantander, line 21: Why are we removing prettier here?

packages/core-contracts/contracts/mocks/utils/MathUtilMock.sol
```diff
@@ -7,4 +7,44 @@ contract MathUtilMock {
     function sqrt(uint x) public pure returns (uint z) {
         return MathUtil.sqrt(x);
     }
+
```
ajsantander, line 10: This is introducing code duplication. If its a mock that simply tests the MathUtil library, it should simply provide wrapper functions that expose the library's internal functions.

packages/core-contracts/contracts/ownership/Authorizable.sol
```diff
@@ -0,0 +1,32 @@
...
+import "../errors/ChangeError.sol";
+
+contract Authorizable is Ownable, IAuthorizable, AuthorizableMixin {
+    event AuthorizedChanged(address oldAuthorizad, address newAuthorized);
+
+    function setNewAuthorized(address newAuthorized) public override onlyOwner {
+        AuthorizableStore storage store = _authorizableStore();
+        address oldAuthorized = store.authorized;
+
+        if (newAuthorized == address(0)) {
```
ajsantander, line 17: This check can be hoisted up in the function to save gas in the case of a revert.

packages/core-contracts/test/contracts/proxy/UUPSProxy.test.js
```diff
@@ -21,7 +21,11 @@ describe('UUPSProxy', () => {
     describe('when setting an EOA as the first implementation', () => {
       it('reverts', async () => {
         const factory = await ethers.getContractFactory('UUPSProxy');
-        await assertRevert(factory.deploy(user.address), `NotAContract("${user.address}")`);
+        //console.log(await ethers.getContractAt('UUPSProxy', ethers.constants.ZeroAddress));
```
ajsantander, line 24: nit: remove console.log

ajsantander (approved):

package.json
dbeal-eth: it was causing errors for "warnings" in typescript code which can't be handled right now and, if i'm not mistaken, eslint covers prettier problems already so its redundant.

packages/core-contracts/contracts/mocks/utils/MathUtilMock.sol
dbeal-eth: this code was from @leomassazza because this merge from `feature-v3-mvp` to main includes all the code that leo worked on previously. @leomassazza want to comment on this?

PR #1042, opened 2022-07-26
Cannon fixes

sunnyvempati:
- upgrade cannon
- fix call to USDToken
- fix breaking test with setting fund position

### DIFF
--- a/packages/synthetix-main/test/contracts/modules/FundModuleFundAdmin.test.js
+++ b/packages/synthetix-main/test/contracts/modules/FundModuleFundAdmin.test.js
@@ -89,7 +89,7 @@ describe('FundModule - Funds Admin', function () {
     describe('when attempting to set the positions of a non existent fund', async () => {
       it('reverts', async () => {
         await assertRevert(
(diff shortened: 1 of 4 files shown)


packages/synthetix-main/test/contracts/modules/FundModuleFundAdmin.test.js
```diff
@@ -136,7 +136,12 @@ describe('FundModule - Funds Admin', function () {
       });
 
       before('adjust fund positions', async () => {
-        const tx = await FundModule.connect(fundAdmin).setFundPosition(1, [1, 2], [1, 1]);
+        const tx = await FundModule.connect(fundAdmin).setFundPosition(
+          1,
+          [1, 2],
+          [1, 1],
+          [1000, 1000]
```
noahlitvin, line 143: This is the maxValuePerShare, not yet implemented? Maybe we set to 0, presumably what we'll use for an undefined maximum?

noahlitvin (changes requested):

noahlitvin (approved):

PR #1048, opened 2022-08-18
Account permissions view

sunnyvempati:
- fix enumerability for accounts.
- `getAccountPermissions` that returns Array of `{ target: address, roles: Array<string> }`

### DIFF
--- a/packages/synthetix-main/contracts/interfaces/IAccountModule.sol
+++ b/packages/synthetix-main/contracts/interfaces/IAccountModule.sol
@@ -3,11 +3,19 @@ pragma solidity ^0.8.0;
 
 import "@synthetixio/core-contracts/contracts/interfaces/ISatelliteFactory.sol";
 import "@synthetixio/core-modules/contracts/interfaces/INftModule.sol";
+import "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
 
 /// @title Module for managing account token (NFT) and accounts, each account is represented by an NFT
 interface IAccountModule is ISatelliteFactory {
+    struct AccountPermission {
+        address target;
+        bytes32[] roles;
+    }
+
     /// @notice gets the AccountToken address.
     function getAccountAddress() external view returns (INftModule);
+
+    function getAccountPermissions(uint accountId) external view returns (AccountPermission[] memory);
 
     /// @notice creates a new accountToken (NFT)
     function createAccount(uint256 accountId) external;
--- a/packages/synthetix-main/contracts/interfaces/IAccountModule.sol
+++ b/packages/synthetix-main/contracts/interfaces/IAccountModule.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.8.0;
 
 import "@synthetixio/core-contracts/contracts/interfaces/ISatelliteFactory.sol";
 import "@synthetixio/core-modules/contracts/interfaces/INftModule.sol";
-import "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
 
 /// @title Module for managing account token (NFT) and accounts, each account is represented by an NFT
 interface IAccountModule is ISatelliteFactory {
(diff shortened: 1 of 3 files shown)


packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -3,12 +3,20 @@ pragma solidity ^0.8.0;
 
 import "@synthetixio/core-contracts/contracts/interfaces/ISatelliteFactory.sol";
 import "@synthetixio/core-modules/contracts/interfaces/INftModule.sol";
+import "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
```
ajsantander, line 6: Wouldn't it be more appropriate to import SetUtil in AccountModule?

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -3,12 +3,20 @@ pragma solidity ^0.8.0;
...
 
 /// @title Module for managing account token (NFT) and accounts, each account is represented by an NFT
 interface IAccountModule is ISatelliteFactory {
+    struct AccountPermission {
+        address target;
+        bytes32[] roles;
+    }
```
ajsantander, line 13: IMO data structures are better declared in storage contracts, in this case `AccountModuleStorage.sol`.

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -9,6 +9,7 @@ import "../../storage/AccountModuleStorage.sol";
 
 import "@synthetixio/core-modules/contracts/mixins/AssociatedSystemsMixin.sol";
 import "../../mixins/AccountRBACMixin.sol";
+import "hardhat/console.sol";
```
ajsantander, line 12: Please don't leave these in prod code.

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -28,6 +29,22 @@ contract AccountModule is IAccountModule, OwnableMixin, AccountRBACMixin, Associ
         return _getNft(_ACCOUNT_SYSTEM);
     }
 
+    function getAccountPermissions(uint accountId) external view returns (AccountPermission[] memory) {
+        AccountRBAC storage accountRbac = _accountModuleStore().accountsRBAC[accountId];
+
+        uint allPermissionsLength = accountRbac.permissionAddresses.length();
+        AccountPermission[] memory allPermissions = new AccountPermission[](allPermissionsLength);
+        for (uint i = 0; i < allPermissionsLength; i++) {
```
ajsantander, line 37: For set iterations, I recommend starting `i` at `1`, and not doing the `i + 1` thing when accessing the position.

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -28,6 +29,22 @@ contract AccountModule is IAccountModule, OwnableMixin, AccountRBACMixin, Associ
...
+            address permissionAddress = accountRbac.permissionAddresses.valueAt(i + 1);
+            allPermissions[i] = AccountPermission({
```
ajsantander, line 39: Is it really necessary to create a struct in memory? Can't we just return a tuple with arrays?

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
sunnyvempati: only reason i just added it to the interface is because it's only a data structure for a return type and not something that's actually stored in storage.

packages/synthetix-main/contracts/modules/core/AccountModule.sol
sunnyvempati: instead of having two arrays; having an array of objects is easier to consume on the FE.  i can change this, let me know if you feel strongly about this (also a rec from db to create a struct)

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
sunnyvempati: removed

PR #1055, opened 2022-08-23
Updates for publication of JS/TS toolings

mjlescano:
Closes https://www.notion.so/Publish-core-utils-hardhat-router-4fac81be6be64f50a4ae860fad85b309

This PR includes the changes necessary for publishing and let everyone use the `@synthetixio/hardhat-router` plugin.

## core-utils structure
It updates `@synthetixio/core-utils` ts config to be able to be used as `import '@synthetixio/core-utils/utils/*'` instead of `import '@synthetixio/core-utils/dist/utils/*'`

## TODOs
- [x] Publish `@synthetixio/core-utils`
- [x] Publish `@synthetixio/core-contracts`
- [x] Publish `@synthetixio/hardhat-router`


.eslintignore
```diff
@@ -1,3 +1,4 @@
 build/*
 **/coverage*
 packages/*/dist
+packages/core-utils/utils
```
dbeal-eth, line 4: why is this added?

dbeal-eth (commented):
there are a lot of random lint execution fixes. while it is good to run lint, perhaps that should be done on a separate PR

.eslintignore
mjlescano: This is because of this change: https://github.com/Synthetixio/synthetix-v3/pull/1055/files#diff-c213856001d7550e3bd55730a1a2b2dcb919c90e11946a3b500f10bf4557386bR6

Which makes the build to be on the root instead of the `dist` folder, which allows us to have a clearer import like `import '@synthetixio/core-utils/utils/*'`. It's a common pattern used also by hardhat eg. [here](https://github.com/NomicFoundation/hardhat/tree/master/packages/hardhat-core/src)

mjlescano:
> there are a lot of random lint execution fixes. while it is good to run lint, perhaps that should be done on a separate PR

Yup, totally agree that this would be better to have it in separate PRs, but, for this case I would ask you for a free pass as renaming the packages makes me update a lot of files, and each time I edit any file it gets linted by vscode.

PR #1057, opened 2022-08-24
Remove satellites, etc.

noahlitvin:


mjlescano (approved):

packages/core-contracts/contracts/interfaces/IBeacon.sol
```diff
@@ -1,8 +0,0 @@
-//SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
-
-interface IBeacon {
-    function getImplementation() external view returns (address);
-
-    function upgradeTo(address newImplementation) external;
-}
```
ajsantander, line 8: How is this related to satellites? I think beacons are still useful and we might use at some point. However, I do agree to remove anything not being used atm. So your call...

packages/core-contracts/contracts/mocks/proxy/BeaconProxyMock.sol
```diff
@@ -1,17 +0,0 @@
-//SPDX-License-Identifier: MIT
```
ajsantander, line 1: Again, depends on wether you want to keep beacons around or not.

packages/core-modules/contracts/mixins/AssociatedSystemsMixin.sol
```diff
@@ -14,6 +14,10 @@ contract AssociatedSystemsMixin is AssociatedSystemsStorage {
     bytes32 internal constant _KIND_ERC721 = "erc721";
     bytes32 internal constant _KIND_UNMANAGED = "unmanaged";
 
+    function _getSystemAddress(bytes32 id) internal view returns (address) {
```
ajsantander, line 17: Maybe `_getAssociatedSystemAddress` would be a better name (more verbose, but more communicative)

packages/core-modules/contracts/modules/NftModule.sol
```diff
@@ -12,11 +12,6 @@ import "../storage/NftStorage.sol";
 import "../interfaces/INftModule.sol";
 
 contract NftModule is INftModule, ERC721Enumerable, NftStorage, InitializableMixin, OwnableMixin {
-    event Mint(address owner, uint nftId);
-
-    // ---------------------------------------
-    // Chores
-    // ---------------------------------------
```
ajsantander, line 19: 😆

ajsantander (approved):

PR #1062, opened 2022-08-25
Nft trasnfer hook

ajsantander:
Adds a hook that allows the account system to notify the core system of when an NFT is transferred.

The system needs this, because it caches the owner of each account in its RBAC.


packages/core-modules/contracts/storage/NftStorage.sol
```diff
@@ -4,7 +4,6 @@ pragma solidity ^0.8.0;
 contract NftStorage {
     struct NftStore {
         bool initialized;
-        address mainProxy;
```
ajsantander, line 7: Not needed anymore. The account system knows who the system is because it should be its owner.

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -39,6 +39,15 @@ interface IAccountModule {
...
 
+    /**
+     * @dev Allows the account token to notify the system when a transfer has occurred
+     *
+     * Requirements:
+     *
+     * - `msg.sender` must be the account token.
+     */
+    function notifyAccountTransfer(address to, uint256 accountId) external;
+
```
ajsantander, line 50: This is the function in the core system that gets called by the account system whenever an NFT is transferred. The core system uses it to update its cache of who the owner of a particular NFT is.

packages/synthetix-main/contracts/interfaces/IAccountModule.sol
```diff
@@ -89,4 +98,9 @@ interface IAccountModule {
      * @dev Returns the address for the account token used by the module.
      */
     function getAccountTokenAddress() external view returns (address);
+
+    /**
+     * @dev Returns the address that owns a given account, as recorded by the system.
+     */
+    function accountOwner(uint accountId) external view returns (address);
```
ajsantander, line 105: Added this function to have visibility on who the core system caches to be the owner of an account.

packages/synthetix-main/contracts/modules/account/AccountTokenModule.sol
```diff
@@ -3,11 +3,20 @@ pragma solidity ^0.8.0;
...
         emit Mint(owner, nftId);
     }
+
+    function _postTransfer(
+        address from,
+        address to,
+        uint256 tokenId
+    ) internal virtual override {
+        IAccountModule(_getOwner()).notifyAccountTransfer(to, tokenId);
+    }
```
ajsantander, line 21: This is the hook that runs after every transfer, and lets the account system update the core system after a transfer.

packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -20,6 +20,14 @@ contract AccountModule is IAccountModule, OwnableMixin, AccountRBACMixin, Associ
...
     error InvalidRole();
 
+    modifier onlyAccountToken() {
+        if (msg.sender != address(getAccountTokenAddress())) {
+            revert OnlyTokenProxyAllowed(msg.sender);
+        }
+
+        _;
+    }
+
```
ajsantander, line 30: Only moved this up from below (I like modifiers up)

noahlitvin: Rename to `OnlyAccountTokenProxy`?

noahlitvin (approved):

PR #1064, opened 2022-08-25
Fix router validation

mjlescano:
This PR fixes a bug when generating deployment sources artifacts that caused to some validations to fail. It removes the old caching strategy, which contained the bug, and and instead simply debounces the persistance to files to only once every tick, this way minimising the amount to writes to disk.


packages/hardhat-router/internal/autosave-object.js
```diff
@@ -23,6 +24,8 @@ module.exports = function autosaveObject(file, initialState = {}) {
 
   const data = JSON.parse(fs.readFileSync(file));
 
+  const _write = debounce(() => write(file, data), 0);
```
mjlescano, line 27: Debounce persitance to files to once every tick. This improves a lot the perfomance during saves in a single sync loop.

packages/hardhat-router/internal/process-contracts.js
```diff
@@ -72,18 +45,25 @@ async function _initContractSource(contractFullyQualifiedName) {
...
   }
 
   // Save the ABIs of all the contracts
   for (const [sourceName, contracts] of Object.entries(buildInfo.output.contracts)) {
     for (const [contractName, attributes] of Object.entries(contracts)) {
-      deployment.abis[`${sourceName}:${contractName}`] = attributes.abi;
+      const key = `${sourceName}:${contractName}`;
+      if (!deployment.abis[key]) {
+        deployment.abis[key] = attributes.abi;
+      }
```
mjlescano, line 66: Added some extra checks just for performance nicities.

packages/hardhat-router/tasks/deploy.js
```diff
@@ -77,7 +76,6 @@ task(TASK_DEPLOY, 'Deploys all system modules')
       await hre.run(SUBTASK_VALIDATE_STORAGE);
       await hre.run(SUBTASK_VALIDATE_MODULES);
       await hre.run(SUBTASK_VALIDATE_INTERFACES);
-      await hre.run(SUBTASK_VALIDATE_SATELLITES);
```
mjlescano, line 80: There were also some pending removals for the old satellites implementation.

PR #1069, opened 2022-08-28
Account module tests review more

ajsantander:


packages/synthetix-main/cannonfile.toml
```diff
@@ -37,8 +37,8 @@ modified = ["contracts/modules/esnx", "contracts/modules/common", "scripts"]
 [run.deploy_core]
 exec = "scripts/deploy.js"
 func = "deploy"
-args = ["Core", ".*/(core|common)/.*"]
-modified = ["contracts/modules/core", "contracts/modules/common", "contracts/utils", "scripts"]
+args = ["Core", ".*/(core|common|mocks)/.*"]
+modified = ["contracts/modules/core", "contracts/modules/common", "contracts/modules/mocks", "contracts/utils", "scripts"]
```
ajsantander, line 41: This is interesting. We need to decide how we want to mock modules.

The way I chose to do it is include the mocks in the system, because I found an unfinished implementation for that in the code, and I want to deviate from what you guys are doing as little as possible.

If we do decide this way is ok, I recommend that we see if there is a way to exclude mocks in an official build with cannon.

packages/synthetix-main/contracts/mocks/AccountRBACMixinModuleMockStorage.sol
```diff
@@ -3,13 +3,14 @@ pragma solidity ^0.8.0;
...
-        uint256 rbacValue;
+        uint256 stakeMock;
+        uint256 mintMock;
     }
 
     function _mixinModuleMockStore() internal pure returns (AccountRBACMixinModuleMockStore storage store) {
         assembly {
             // bytes32(uint(keccak256("io.synthetix.snx.accountrbacmixinmodule")) - 1)
-            store.slot := 0x1ce4ca68760c4591c093e69c4f3107e99b4faf6a64521759223b68428b166546
+            store.slot := 0x186e65a22afe779f65be55d2f02e866d66351197187020749997986235b258c1
```
ajsantander, line 13: Bad hash?

@mjlescano 👀 
I recommend that we consider using this instead of hardcoded hashes:

```
contract SomeStorage {
  bytes32 private constant slot = bytes32(uint(keccak256("io.synthetix.snx.some_storage")) - 1);
  
function _someStore() internal pure returns (SomeStore storage store) {
  assembly {
    store.slot = slot;
  }  
}
```

packages/synthetix-main/contracts/mocks/IAccountRBACMixinModuleMock.sol
```diff
@@ -2,7 +2,11 @@
 pragma solidity ^0.8.0;
 
 interface IAccountRBACMixinModuleMock {
-    function interactWithAccount(uint accountId, uint inputValue) external;
+    function mock_AccountRBACMixin_stake(uint accountId, uint newStakeMock) external;
```
ajsantander, line 5: Just making the mock fn names more explicit (and unique).

packages/synthetix-main/contracts/mocks/IAccountRBACMixinModuleMock.sol
```diff
@@ -2,7 +2,11 @@
...
 
-    function getRBACValue() external view returns (uint);
+    function mock_AccountRBACMixin_mint(uint accountId, uint newMintMock) external;
+
+    function mock_AccountRBACMixin_getStakeMock() external view returns (uint);
+
+    function mock_AccountRBACMixin_getMintMock() external view returns (uint);
```
ajsantander, line 11: And adding a few mock fns...

packages/synthetix-main/test/integration/modules/CollateralModuleConfiguration.test.ts
```diff
@@ -6,7 +6,7 @@ import { ethers } from 'hardhat';
 
 import { bootstrap } from '../bootstrap';
 
-describe('CollateralManagerConfiguration (SCCP)', function () {
+describe.skip('CollateralManagerConfiguration (SCCP)', function () {
```
ajsantander, line 9: Now you'll see a bunch of skips added.

I simply removed all only's from AccountModule reviewed tests, and added skip's to everything else.

packages/synthetix-main/cannonfile.toml
ajsantander, line 41: How do we mock modules?

I found this pattern in the code, and I finished it to make the most minimal noise with previous work.

If we go this way, we might want to enable cannon to exclude folders in prod builds.

packages/synthetix-main/contracts/mocks/AccountRBACMixinModuleMockStorage.sol
mjlescano: For me that's ok! But is going to be more expensive to deploy (probably not an issue).

mjlescano (approved):

noahlitvin (approved):

PR #1070, opened 2022-08-29
sip 304 markets

dbeal-eth:
Implements SIP-304

notes:
* as per spec, fund managers can specify "maxDebt" when entering the pool, which gives fund. However, to control the amount of
liquidity from a system level, a SCCP controllable value `liquidityRatio` which controls how much liquidity may be delegated to a market, and changes
dynamically based on the current collateral value/debt saddling
* uses a heap to track when funds are out of range. added a few tests to make sure this functionality is rock solid :rock:
* in order for many market liquidity calculation functions to take effect, it is necessary to trigger fund or vault functions
   * note, however, vaults can still be liquidated if debt goes up (without an update) and anyone can trigger an update by calling
     `vaultDebt` with the affected vault
* there may be an edge case attack vector where an attacker who manages to gain complete control over a vault's balance measurement
causes the registered balance to repetedly oscillate between a low balance and high balance, incurring more debt than the fund owner likely intended.
To mitigate this, we might be able to use
a second heap to measure when a fund is entering/leaving a pool. However, there is probably a cheaper and easier way to handle this.


packages/hardhat-router/internal/process-contracts.js
```diff
@@ -36,12 +36,15 @@ async function initContractData(contractFullyQualifiedName, extraData = {}) {
     'source-abis',
     deployedBytecodeHash + '.json'
   );
-  if (fs.existsSync(cachedSourceInfo)) {
+
+
+  await _initContractSource(contractFullyQualifiedName);
```
mjlescano, line 41: This fix is already merged to `main`, a rebase should fix it :)

noahlitvin (approved):

PR #1088, opened 2022-09-01
fix vault rewards module tests

dbeal-eth:
fixes boil down to:
* using `evm_setNextBlockTimestamp` instead of just `fastForward` within
  `fastForwardTo`-- ensures exact timestamp is delivered
* recognition that `evm_mine` always causse the block timestamp to
  incraese by 1 second, so we should be able to account for this in the test to get
the exact time

also btw the refactoring of these tests was not exactly the same as
before, which is part of the reason why the tests were failing when they
succeeded before

### DIFF
--- a/packages/core-utils/src/utils/hardhat/rpc.ts
+++ b/packages/core-utils/src/utils/hardhat/rpc.ts
@@ -21,13 +21,9 @@ export async function fastForward(seconds: number, provider: ethers.providers.Js
 }
 
 export async function fastForwardTo(time: number, provider: ethers.providers.JsonRpcProvider) {
-  const now = await getTime(provider);
+  await provider.send('evm_setNextBlockTimestamp', [time]);
 
-  if (time < now) {
-    throw 'Cannot fast forward to a past date.';
-  }
-
-  await fastForward(time - now, provider);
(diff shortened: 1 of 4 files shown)


dbeal-eth:
`VaultRewardsModule` tests are succeeding, but other upstream tests were failing

noahlitvin (approved):

mjlescano (approved):

packages/core-utils/src/utils/hardhat/rpc.ts
```diff
@@ -21,13 +21,9 @@ export async function fastForward(seconds: number, provider: ethers.providers.Js
 }
 
 export async function fastForwardTo(time: number, provider: ethers.providers.JsonRpcProvider) {
-  const now = await getTime(provider);
+  await provider.send('evm_setNextBlockTimestamp', [time]);
```
mjlescano, line 24: Nice! didn't know about `evm_setNextBlockTimestamp` 👏

dbeal-eth: these docs are very helpful https://book.getfoundry.sh/reference/anvil/?highlight=time#supported-rpc-methods

PR #1093, opened 2022-09-06
Moved Heap to core-contracts

ajsantander:
Moved Heap.sol from synthetix-main to core-contracts, since its generic.

Note that it was renamed to HeapUtil.sol to stay consistent with other utils in core-contracts.


dbeal-eth (approved):
it also renames "heap" to "heapUtil" for some reason but I guess that is fine. we verified the tests are actually running on the `core-contracts` package as well, right?

noahlitvin (approved):

PR #1100, opened 2022-09-09
SIP-306 LegacyMarket + incomplete sample markets

dbeal-eth:
Includes code complete `LegacyMarket`, and some sample markets we started working on during the offsite.

Some minor changes were required in order to implement `LegacyMarket` with v3-core as we had it, specifically, we needed to 

Currently `LegacyMarket` uses about 2 million gas to migrate a single account with 3 escrow entries. On mainnet, this comes out to about $40 at current gas prices. This is part of the reason that a forced migration is inadvisible.

when the time comes, we can test the legacy market with much larger scale once we have v2x changes deployed

Companion PR on v2x (SIP-299): https://github.com/Synthetixio/synthetix/pull/1897

cc @barrasso


dbeal-eth:
just a quick note that the test failures I am not able to reproduce at all on my local machine so if you see a ton of dumb commits after this point that is why. There is just nothign I can do if I cannot reproduce whatever is wrong with the CI from a fresh checkout on my local machine...

dbeal-eth:
please ignore the reduction in test coverage, the coverage loss will be changed/overwritten by the looming router-restructure changes whiych are likely to be merged anyway.

noahlitvin (approved):

PR #1103, opened 2022-09-13
Gas profile

dbeal-eth:
adds support for measuring gas usage using `hardhat-gas-reporter`

adds support for coverage (though its slow)

dramatically speeds up the base case of running the tests with no coverage or gas recording

upgrades cannon to 1.0.31

some of hte tests. This is because the lint fixes revealed some test cases were silently failing due to forgetting to use `await`. For now I have left these since the errors technically existed before, but they should be corrected reasonably quickly ofc


ajsantander (approved):
Awesome!

PR #1104, opened 2022-09-13
Fix gas profile PR minor errors

ajsantander:
Generally minor errors, but one problem with Anvil which was reported to the tooling team.

### DIFF
--- a/packages/core-utils/src/utils/assertions/assert-event.ts
+++ b/packages/core-utils/src/utils/assertions/assert-event.ts
@@ -21,11 +21,11 @@ export default async function assertEvent(
           : ''
       })`;
 
+      seenEvents.push(text);
+
       if (text.match(expectedMatch) || text.toLowerCase().includes(expectedMatch.toLowerCase())) {
         return;
       }
-
-      seenEvents.push(text);
     } catch {
       // noop
     }
--- a/packages/core-utils/src/utils/assertions/assert-revert.ts
+++ b/packages/core-utils/src/utils/assertions/assert-revert.ts
@@ -23,7 +23,7 @@ function getErrorData(err: ErrorObject): string | null {
 
 export default async function assertRevert(
   tx: Promise<ethers.providers.TransactionResponse>,
-  expectedMessage = 'transaction failed',
+  expectedMessage?: string,
   contract?: ethers.Contract
 ) {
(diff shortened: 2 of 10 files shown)


packages/core-utils/src/utils/assertions/assert-event.ts
```diff
@@ -21,11 +21,11 @@ export default async function assertEvent(
           : ''
       })`;
 
+      seenEvents.push(text);
+
```
ajsantander, line 25: If there is a partial match, seenEvents would never be populated, and the console would show something like:

Expected event "ABC()" but found "" (nothing).

packages/core-utils/src/utils/assertions/assert-revert.ts
```diff
@@ -23,7 +23,7 @@ function getErrorData(err: ErrorObject): string | null {
 
 export default async function assertRevert(
   tx: Promise<ethers.providers.TransactionResponse>,
-  expectedMessage: string,
+  expectedMessage = 'transaction failed',
```
ajsantander, line 26: This allows you to just check for a revert, without reason.

packages/synthetix-main/test/integration/modules/MarketManagerModule.test.ts
```diff
@@ -74,14 +74,21 @@ describe('MarketManagerModule', function () {
...
+      //   MockMarket().connect(user1).buySynth(One),
+      //   `MarketDepositNotApproved("${await user1.getAddress()}", "${
+      //     MockMarket().address
+      //   }", ${One.toString()}, 0)`,
+      //   systems().Core
+      // );
+
+      // TODO: Replace this with the commented check above
+      // when fixed. For now we check that it simply reverts.
+      await assertRevert(MockMarket().connect(user1).buySynth(One));
```
ajsantander, line 91: This is the problem with Anvil that we found: If the entry point is not the contract that emits the event, Anvil fails to understand the revert reason.

mjlescano: This is being tracked here: https://www.notion.so/Cannon-Errors-through-proxies-are-not-correctly-displayed-5dd75d2b447d496d86aa1c05ccc62dcb

PR #1111, opened 2022-09-19
Market-provided Collateral

noahlitvin:


packages/synthetix-main/contracts/mixins/MarketManagerMixin.sol
```diff
@@ -139,7 +139,24 @@ contract MarketManagerMixin is MarketManagerStorage, PoolModuleStorage {
         return IMarket(marketData.marketAddress).reportedDebt();
     }
 
+    function _depositedCollateralValue(MarketData storage marketData) internal view returns (uint) {
+        uint totalDepositedCollateralValue = 0;
+        address[] memory collateralTypes = _collateralStore().collateralTypes.values();
```
dbeal-eth, line 144: rather than iterating all the collateral types, it would be better to iterate over a short list of *collateral entries* (aka, collateral address/amount pair in a struct). Since most markets will only work with one/a small subset of collatreal type in our stated use case

packages/synthetix-main/contracts/modules/core/CollateralModule.sol
```diff
@@ -144,6 +144,28 @@ contract CollateralModule is
         return _getAccountUnassignedCollateral(accountId, collateralType);
     }
 
+    function depositMarketCollateral(
```
dbeal-eth, line 147: these functions could be potentially moved to their own module. modules are, after all, public function interfaces to manipulate underlying pools of data

packages/synthetix-main/contracts/storage/MarketManagerStorage.sol
```diff
@@ -25,6 +25,8 @@ contract MarketManagerStorage {
         // used to attach/reattach pools to a market if it goes below a certain debt per debt share
         HeapUtil.Data outRangePools;
         SharesLibrary.Distribution debtDist;
+        // @notice the amount of a given collateral type deposited by this market
+        mapping(address => uint128) depositedCollateral;
```
dbeal-eth, line 29: as mentioned up above, it might be better to store this as an `DepositCollateralEntry[] depositedCollateral` and iterate over the list rather than creating a mapping which we have to scan all collateral types for

dbeal-eth (commented):
good start

noahlitvin:
Noting that tests pass with `npm run test` but not `npm run coverage`. The issue stems from `assertRevert`.

packages/synthetix-main/contracts/mixins/MarketManagerMixin.sol
```diff
@@ -143,7 +143,23 @@ contract MarketManagerMixin is MarketManagerStorage, PoolModuleStorage {
...
+
+        return totalDepositedCollateralValue;
+    }
+
     function _totalBalance(uint marketId, MarketData storage marketData) internal view returns (int) {
-        return int(IMarket(marketData.marketAddress).reportedDebt(marketId)) + marketData.issuance;
+        return
+            int(IMarket(marketData.marketAddress).reportedDebt(marketId)) +
+            marketData.issuance -
+            int(_depositedCollateralValue(marketData));
```
sunnyvempati, line 163: maybe just pass in `marketData.depositedCollateral` to `_depositedCollateralValue` instead of the entire data.

packages/synthetix-main/contracts/modules/core/MarketCollateralModule.sol
```diff
@@ -0,0 +1,104 @@
...
+import "../../storage/CollateralStorage.sol";
+import "../../storage/MarketManagerStorage.sol";
+
+import "../../utils/ERC20Helper.sol";
+
+contract MarketCollateralModule is IMarketCollateralModule, CollateralStorage, MarketManagerStorage, OwnableMixin {
+    using ERC20Helper for address;
+
+    error InsufficientMarketCollateralDepositable(uint marketId, address collateralType, uint amountToDeposit);
+    error InsufficientMarketCollateralWithdrawable(uint marketId, address collateralType, uint amountToWithdraw);
```
sunnyvempati, line 15: i've been moving errors and events into the interface; not sure what the standard is here but felt cleaner

sunnyvempati (approved):
pre-approving; a few comments.  didn't review the logic closely but tests look good

PR #1113, opened 2022-09-19
Allow nominations in voting period

ajsantander:


mjlescano (approved):

leomassazza (approved):
LGTM. As mentioned in a private chat. The only concern is communicational. It should be clear to users that nominations are open on voting period BUT they cannot withdraw a nomination in that period, as before.

PR #1116, opened 2022-09-20
Spot Market Impl

sunnyvempati:
- initial draft of spot market implementation


markets/spot-market/interfaces/ISynth.sol
```diff
@@ -0,0 +1,11 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@synthetixio/main/contracts/interfaces/external/IMarket.sol";
```
Rickk137, line 4: can be removed

Rickk137 (changes requested):

noahlitvin (approved):

PR #1123, opened 2022-09-23
Fix synthetix-main coverage

mjlescano:


packages/synthetix-main/test/integration/bootstrap.ts
```diff
@@ -34,7 +34,9 @@ before(async function () {
   // allow extra time to build the cannon deployment if required
   this.timeout(300000);
 
-  const cannonInfo = await hre.run('cannon:build');
+  const cmd = hre.network.name === 'cannon' ? 'build' : 'deploy';
```
dbeal-eth, line 37: perfect

dbeal-eth (approved):
looks great!

packages/core-utils/src/utils/assertions/assert-event.ts
```diff
@@ -16,9 +17,7 @@ export default async function assertEvent(
       const parsed = contract.interface.parseLog(log);
 
       const text = `${parsed.name}(${
-        parsed.args
-          ? parsed.args.map((v) => (v.toString ? '"' + v.toString() + '"' : v)).join(', ')
-          : ''
+        parsed.args ? AbiHelpers.formatValues(parsed.args as unknown[]) : ''
```
mjlescano, line 20: With this fix we are formatting the errors the same way as Hardhat does it.

PR #1124, opened 2022-09-26
Router restructure

dbeal-eth:
this PR introduces a proposed `hardhat-router` affiliated project architecture change. Initially, `hardhat-router` enforced creation of storage contracts which would then be extended by their ultimate modules. The storage contracts would have a structure similar to the following:

```
contract SomeStorage {
    struct SomeStore {
        uint myval1;
        uint myval2;
        mapping(uint => uint) something;
    }

    function _someStorage() internal returns SomeStorage [
        assembly {

        }
    }
}
```

Having a manifest definition for the structure is an important feature in the design, but it doesn't address mutations, computed properties, or other "control". Engineers with familiarity with Java will be familiar with object oriented code. Engineers familiar with frontend will be familiar with MVC. Engineers familiar with databases will be familiar with routines. All in all, these concepts all revolve around a design principal where the code which operates closely with the data it is modifying is also linked. This creates code which is highly modular and readable, and increases safety since it is expected that mutations on the data maybe should be done with the included control functions. Currently, `hardhat-router` encourages the use of `Mixin`s which are less suitable for the  Since the storage is also defined as a contract extension for the downstream modules, it could also lead to contract size limit issues in the future. All in all, this system

Thankfully, solidity includes a different type of contract called a `library`, which (with some minor workarounds), can provide a convenient mechanism for resolving these problems. These libraries (hereafter called "storage library") might have a structure similar to the following:

```
library Some {
    struct Data {
        uint myval1;
        uint myval2;
        mapping(uint => uint) something;
    }

    function load(uint id) internal pure returns (Data storage data) {
        bytes32 s = keccak256(abi.encode("Some", id));
        assembly {
            data.slot := s
        }
    }

    function loadSomeIndex() internal pure returns (SetUtil.UintSet storage data) {
        bytes32 s = keccak256(abi.encode("Some_someIndex"));
        assembly {
            data.slot := s
        }
    }



    function create(uint myval1) internal returns (Data storage self) {
        
    }

    function doSomething(Data stroage self) internal {
        self.myval1 = self.myval1 * 2;
        self.myval2 = self.myval1 + 2;
    }
}
```

Note that, using soliditiy library library, it is possibe in a depending contract to declare `

observed benefits benefits:
* the `load` function now performs a lot like how an engineer would load data from a database or an API. They just specify the entry they want and out pops the correct storage object. For example, the code to get an account (and its associated nested structures, such as its balance of all collaterals) is `Account.Data storage account = Account.load(accountId)` or `Account.load(accountId).doSomething()`
* Code associated with modifying/doing complex reads on a data structure is located in the same file which it is defined. Code is naturally split with the splitting of data structures.
* Even the most fundamental of data structures can be reused repeatedly in a nested context. For example, `Distributor`, which used to be located in the shares library, is now just another storage library which is depended upon by `Pool`, `Vault`, and `Market`, fulfilling rewards distributor-like functionalities for each component. Its well encapsulated.
* Since the mutation functions and structure definitions for this data is located within a solidity `library`, if in the future we run into contract size limit issues, we can just convert the functions of a large storage library to `external`. This library could hypothetically be shared/reuesd/recycled between all contracts which use the same data structures through `DELEGATECALL`. Moreover, this can be done retroactively (aka, after the system has already accumulated storage) without implementing anything
* With this code structuring, modules begin to form more of the role of an operator on the data. Since they are depending upon the storage classes which use them, they can 

Note that testing of storage classes can be done consistently by creating and deploying a separate `Test` module (ex `PoolTestModule`) for the functoinality which needs to be verified directly, and then selectively linking it through the cannon cli.

notes:
* I tried to get rid of the assembly in the `load` function, but it is just not possible to set a storage pointer within solidity without doing otherwise. Perhaps in the future solidity could change this? But for now its something it seems we are going to have to deal with it 
* this PR should satisfy the refactoring requirement of equivalent code results while still restructuring. This is because 
* while this PR restructures the actual solidity code in `synthetix-main`, it does nothing to change the structure of the code in the other directories (ex. `core-modules`). the purpose of this PR is to provide a first step and review implementation for the proposed router architecture change
* likewise, other than removing the checks for the valid storage class mutations, these changes do not provide any of the necessary changes for validating storage mutations. that code shoud be updated and verified before we move onto production v3. (cc @mjlescano )

general issues which have been encountered while implementing this structure:
* what should be done about errors/reverts which occur from within a storage library? its possible to define errors from within the library, but that is not the standard that we have been following previously and certain useful parts of the data may not be available. it seems likely that events should always be emitted from modules and never from storage, so I am not listing that as part of the concerns for here
* modifiers apparently cannot be defined from within libraries, so even though it would be most convenient to define a modifier `Pool.exists(id)` from within a storage library, we have to define such modifier. Where should they be stored instead? My goal was to get rid of mixins, so for now I have just put a copy of the needed modifiers in every contract they belong to, which is a bit of duplication.
* this system technically is a bit less efficient than the one from before because the functions operate using `Data storage self` references, and the `self` reference may need to reference the ID or other index of itself in order to communicate with other data structures. This results in extra storage usage, and it is important that the developer remembers to call `create` or similar function when initializing for the first time. This likely is not a problem, but it does seem like kind of an awkward outcome.


packages/synthetix-main/contracts/storage/Account.sol
```diff
@@ -0,0 +1,66 @@
...
+
+        return (totalDeposited, totalAssigned); //, totalLocked, totalEscrowed);
+    }
+
+    function getAssignedCollateral(Data storage self, address collateralType) internal view returns (uint) {
+        uint totalAssigned = 0;
+
+        uint[] memory pools = self.collaterals[collateralType].pools.values();
+
+        console.log("GETTING ASSIGNED", pools.length);
```
mjlescano, line 53: Should this console.log be removed?
```suggestion
```

packages/synthetix-main/contracts/storage/Account.sol
```diff
@@ -0,0 +1,66 @@
...
+        for (uint i = 0; i < pools.length; i++) {
+            uint128 poolIdx = uint128(pools[i]);
+
+            Pool.Data storage pool = Pool.load(poolIdx);
+
+            (uint collateralAmount, , ) = pool.currentAccountCollateral(collateralType, self.id);
+            totalAssigned += collateralAmount;
+            console.log("ASSIGNED", collateralAmount, self.id);
```
mjlescano, line 61: Same
```suggestion
```

ajsantander (approved):

PR #1128, opened 2022-10-03
Oracle manager

sunnyvempati:


packages/oracle-manager/contracts/modules/core/NodeFactoryModule.sol
```diff
@@ -0,0 +1,93 @@
...
+        if (nodeDefinition.nodeType == NodeType.REDUCER) {
+            return ReducerNodeLibrary.process(prices, nodeDefinition.parameters);
+        } else if (nodeDefinition.nodeType == NodeType.EXTERNAL) {
+            return ExternalNodeLibrary.process(prices, nodeDefinition.parameters);
+        } else if (nodeDefinition.nodeType == NodeType.CHAINLINK) {
+            return ChainlinkNodeLibrary.process(nodeDefinition.parameters);
+        } else if (nodeDefinition.nodeType == NodeType.PYTH) {
+            return PythNodeLibrary.process(nodeDefinition.parameters);
+        } else {
+            revert("Unsupported Node Type");
```
dbeal-eth, line 90: use an actual solidity error here. define it above like `error UnsupportedNodeType(uint nodeType)`, then use it like `revert UnsupportedNodeType(nodeDefinition.nodeType)`

packages/oracle-manager/contracts/storage/NodeFactoryStorage.sol
```diff
@@ -0,0 +1,36 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract NodeFactoryStorage {
```
dbeal-eth, line 4: note this will probably need to be updated very soon due to the impending router refactor stuff

packages/oracle-manager/contracts/utils/ChainlinkNodeLIbrary.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import "../storage/NodeFactoryStorage.sol";
+import "../interfaces/external/IAggregatorV3Interface.sol";
+
+library ChainlinkNodeLibrary {
```
dbeal-eth, line 7: the filename is `ChainlinkNodeLIbrary`, but it should be `ChainlinkNodeLibrary` (if the filename doesnt match the name of its internal contract/library, it tends to cause problems)

packages/oracle-manager/contracts/utils/ChainlinkNodeLIbrary.sol
```diff
@@ -0,0 +1,13 @@
...
+    function process(bytes memory parameters) internal view returns (NodeFactoryStorage.NodeData memory) {
+        IAggregatorV3Interface chainlinkAggr = IAggregatorV3Interface(abi.decode(parameters, (address)));
+        (, int256 price, , uint256 updatedAt, ) = chainlinkAggr.latestRoundData();
+        return NodeFactoryStorage.NodeData(price, updatedAt, 0, 0);
```
dbeal-eth, line 11: :+1:

packages/oracle-manager/contracts/utils/ReducerNodeLibrary.sol
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import "../storage/NodeFactoryStorage.sol";
+
+library ReducerNodeLibrary {
```
dbeal-eth, line 6: this library is a great example for why I think this strategy for nodes works so well. its looking really good, even has a `quickSort` function which, tbh, might be overkill for solidity, but its great

packages/oracle-manager/scripts/deploy.js
```diff
@@ -0,0 +1,90 @@
...
+const {
+  TASK_DEPLOY,
+  SUBTASK_GET_MULTICALL_ABI,
+  SUBTASK_GET_DEPLOYMENT_INFO,
+} = require('@synthetixio/hardhat-router/task-names');
+
+/**
+ * Generate the file contracts/Router.sol including the given modules in its source.
+ */
+module.exports.deploy = async function deploy(runtime, prefix, modules) {
```
dbeal-eth, line 11: ideally (don't worry about for this PR but just want to make a note) this deployment file should be embedded within the `hardhat-router` or something in order to make it easy for any project to just call cannon run step here without needing to copy and paste it like this

packages/oracle-manager/test/integration/modules/NodeModule.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('Test price on leaf nodes', async () => {
+    let priceData = await systems().Core.connect(owner).process(nodeId1());
+    assertBn.equal(priceData.price, ethers.utils.parseEther('1'));
+
+    priceData = await systems().Core.connect(owner).process(nodeId2());
+    assertBn.equal(priceData.price, ethers.utils.parseEther('0.9'));
+  });
+
+  it('register a max reducer', async () => {
```
dbeal-eth, line 33: specific tests regarding the reducers should go in their own file which specifically tests the library functions outside of the interference of the Node factory. I can show you how to do this later if you want, its actually not too bad.

dbeal-eth (approved):
there seem to be a lot of interesting extra changes bundled into this so might want to double check the merge base

most of my suggestions are very minor

PR #1138, opened 2022-10-20
Adds documentation and minor code readability improvements to Distributions

ajsantander:


packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -4,42 +4,140 @@ pragma solidity ^0.8.0;
 import "@synthetixio/core-contracts/contracts/utils/MathUtil.sol";
 
 import "./DistributionActor.sol";
+import "../errors/ParameterError.sol";
 
+/**
+ * @title Tracks value distributed amongst a set of actors using shares. A share represents a unit in which the total value is distributed, and thus the granilarity of the distribution.
```
dbeal-eth, line 10: "tracks value distributed" usually (at least in english) repeating terms which are used in the title of something is not helpful in a description. maybe it would be better to say somethign like "In a pattern typical to the longstanding `StakingRewards` contract, divides a quanitity of `value` between a set of actors proportional to the number of shares they have"

packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -48,81 +146,121 @@ library Distribution {
...
-        actor.shares = uint128(shares);
-        // lastValuePerShare remains unset because they contributed
+        // Note: No need to udpate actor.lastValuePerShare
+        // because they contributed value to the distribution.
     }
 
     /**
-     * get the number of shares recorded for an actor
+     * @dev Upadates an actor's lastValuePerShare to the distribution's current valuePerShare, and
+     * returns the total change in value for the actor.
```
dbeal-eth, line 220: returns the change in value from the actor from the last update

dbeal-eth (approved):

packages/synthetix-main/contracts/storage/Distribution.sol
ajsantander: Oh I really like the last phrase! I would prefer to not refer to any v2x contracts, or even assume previous knowledge of that. Better if we can explain everything with no previous context.

ajsantander: Updated

packages/synthetix-main/contracts/storage/Distribution.sol
ajsantander: Updated

packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -4,42 +4,141 @@ pragma solidity ^0.8.0;
...
+ * - valuePerShare: 50 USD
+ * - totalValue: 100 USD
+ * 3) Actor 1's shares are duplicated
+ * - shares:
+ *   - actor1: 2 (100 USD)
+ *   - actor2: 1 (50 USD)
+ * - totalShares: 3 (150 USD)
+ * - valuePerShare: 50 USD
+ * - totalValue: 150 USD
+ */
```
sunnyvempati, line 68: awesome comment; ty.

sunnyvempati (approved):
this is great; much appreciated.

PR #1140, opened 2022-10-21
Feature flags for pools and markets

sunnyvempati:
Introduced feature flags to our contracts.

- `owner` (pdao) can add, remove addresses for a given feature flag (bytes32), or disable the feature entirely.

TODO:
- [x] Add to cannonfile
- [x] Write tests
- [x] Add events


packages/core-modules/contracts/modules/FeatureFlagModule.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function addToFeatureFlag(bytes32 feature, address permissioned) external override {
+        OwnableStorage.onlyOwner();
+        FeatureFlag.load().featureFlags[feature].permissionedAddresses.add(permissioned);
+
+        emit FeatureFlagAddressAdded(feature, permissioned);
+    }
+
+    function addMultipleToFeatureFlag(bytes32 feature, address[] calldata permissioned) external override {
```
dbeal-eth, line 30: in the interest of streamlining the contract ABI, `addMultipleToFeatureFlag` and `removemultipleFromFeatureFlag` should be removed in favor of simply using `MulticallModule` with multiple `addToFeatureFlag` calls.

packages/core-modules/contracts/modules/mocks/SampleFeatureFlagModule.sol
```diff
@@ -0,0 +1,17 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../interfaces/ISampleFeatureFlagModule.sol";
+import "../../storage/SampleStorage.sol";
+import "../../storage/FeatureFlag.sol";
+
+contract SampleFeatureFlagModule is ISampleFeatureFlagModule {
```
dbeal-eth, line 8: fyi sample/mock modules like this will be replaced with some sort of codegen in the near future so don't invest a ton of time in this.

packages/core-modules/contracts/storage/FeatureFlag.sol
```diff
@@ -0,0 +1,34 @@
...
+import "@synthetixio/core-contracts/contracts/errors/AccessError.sol";
+import "./Permission.sol";
+
+library FeatureFlag {
+    using SetUtil for SetUtil.AddressSet;
+
+    error FeatureUnavailable();
+
+    struct Data {
+        mapping(bytes32 => Permission.Data) featureFlags;
```
dbeal-eth, line 13: rather than having a single property mapping `bytes32 => Permission.Data`, have the `load` function take a parameter `bytes32` and then add to the `abi.encode`. Then, load that as a slot directly.

better pattern when you look at the code, saves a sidgen of gas

dbeal-eth: in fact, you could probably just have `Permission` on its own and more or less completely remove this file (or do the opposite, move the `Permission` structure into this file and completely remove the current mapping)

packages/core-modules/test/contracts/modules/FeatureFlagModule.test.js
```diff
@@ -0,0 +1,137 @@
...
+
+  it('does not allow non-owners to set feature flags', async () => {
+    await assertRevert(
+      FeatureFlagModule.connect(user).setFeatureFlag(FEATURE_FLAG_NAME, true),
+      'Unauthorized'
+    );
+  });
+
+  let setupTx;
+  before('setup feature flag', async () => {
```
dbeal-eth, line 36: this before (afaik) runs before *any* of the `it` clauses in this describe, so its confusing to put it in the middle here

dbeal-eth: please let me know if I am wrong about taht

packages/core-modules/test/contracts/modules/FeatureFlagModule.test.js
```diff
@@ -0,0 +1,137 @@
+const { ethers } = hre;
```
dbeal-eth, line 1: in general this file is not following the same testing approach we have settled upon in the `synthetix-main`. In general, tests should be focused on breakign down into each writable function, testing all the different permutations of arguments you might submit to the function, verifying that the corresponding read functions are behaving as expected, etc. This tends to result in very high and quality test coveage.

Please take a look at one of the `synthetix-main` test files, such as `VaultModule` for example. Then can you come back and restructure these tests a bit to have a similar structure?

packages/core-modules/test/contracts/modules/FeatureFlagModule.test.js
```diff
@@ -0,0 +1,137 @@
...
+
+  describe('disable feature flag for all', async () => {
+    before('disable feature flag', async () => {
+      await FeatureFlagModule.connect(owner).setFeatureFlag(FEATURE_FLAG_NAME, false);
+    });
+
+    it('reverts when feature flag is disabled', async () => {
+      await assertRevert(
+        SampleFeatureFlagModule.connect(permissionedUser).setFeatureFlaggedValue(25),
+        'FeatureUnavailable'
```
dbeal-eth, line 133: you will likely need to put the contract which has the ABI containin the `FeatureUnavailable` flag in order for parsing... or have you not found that necessary?

packages/synthetix-main/cannonfile.toml
```diff
@@ -161,6 +166,7 @@ modified = ["contracts/modules/common", "contracts/modules/account"]
 depends = [
   "contract.OwnerModule",
   "contract.UpgradeModule",
+  "contract.FeatureFlagModule",
```
dbeal-eth, line 169: do all of the systems really need feature flags? is it a common module?

packages/synthetix-main/contracts/modules/core/MarketManagerModule.sol
```diff
@@ -23,6 +24,7 @@ contract MarketManagerModule is IMarketManagerModule {
     error MarketDepositNotApproved(address market, address from, uint requestedAmount, uint approvedAmount);
 
     function registerMarket(address market) external override returns (uint128 marketId) {
+        FeatureFlag.onlyIfFeatureFlag("market");
```
dbeal-eth, line 27: you could probably just rename fhe function to be like `FeatureFlag.ensureEnabled`

dbeal-eth: `"market"` should be a constant, not an inline string

we should probably just make a generla purpose `constants` contract which everything using constants inherits from. I believe solidity is smart enough to optimize out any unused constants so its generally the best option

dbeal-eth (dismissed):

packages/core-modules/contracts/modules/mocks/SampleFeatureFlagModule.sol
sunnyvempati: sweet 👍🏽 ; pretty basic here

packages/core-modules/contracts/modules/FeatureFlagModule.sol
sunnyvempati: @ajsantander requested i add these as a convenience function.

packages/core-modules/test/contracts/modules/FeatureFlagModule.test.js
sunnyvempati: yep you right; moving it to the top

packages/core-modules/test/contracts/modules/FeatureFlagModule.test.js
sunnyvempati: yea these can be organized a bit better.   all paths are tested but are a bit hard to follow, i'll reorg in a separate PR and have it look similar to VaultModule, thanks for the link

packages/synthetix-main/contracts/modules/core/MarketManagerModule.sol
sunnyvempati: cool, i'll make both these changes 👍🏽

packages/core-modules/contracts/storage/FeatureFlag.sol
sunnyvempati: ooooh smart!  great idea, thank you!!

packages/core-modules/contracts/modules/FeatureFlagModule.sol
dbeal-eth: @ajsantander can we get some documentation here as to the reasoning?

sunnyvempati: i think the idea was just gas considerations since we eliminate the `CALL` ops with multicall

dbeal-eth: 1. this function is infrequently called so the amounto f gas saved lifetime is almost certainly <10m
2. the small cost of having increased gas is much lower than the significant cost of having additional surface area which we have to test, verify, and maintain

sunnyvempati: @dbeal-eth yeah good points; and after talking to noah and afif, we're planning on like adding one or two addresses here and there, and then just removing the flag altogether to allow everyone so I think its safe to just remove these.  lemme know if you think otherwise @ajsantander

packages/synthetix-main/cannonfile.toml
sunnyvempati: moved to core; only pool and markets, but i only foresee us using it for core stuff anyway.  👍🏽

packages/core-modules/contracts/interfaces/IFeatureFlag.sol
```diff
@@ -0,0 +1,14 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @title Interface for feature flags
+interface IFeatureFlag {
```
noahlitvin, line 5: Should be `IFeatureFlagModule`?

noahlitvin (approved):

PR #1141, opened 2022-10-24
SIP-309 Market Locked Collateral

dbeal-eth:
Implements [SIP-309](https://sips.synthetix.io/sips/sip-309/).

The code ended up a little less neat than I would have hoped for, especailly in the PoolModule.

Please note these other changes:
* **INTERFACE CHANGE** `PoolModule.setPoolConfiguration` has been changed to take `MarketDistribution` object instead of 3 separate arrays. This improves the code clarity, , and reduces the prevelance of stack too deep error.
* **INTERFACE CHANGE** `PoolModule.getPoolConfiguration` has been changed to correspond to the parameter changes in `PoolModule.setPoolConfiguration`
* `VaultModule.mintUsd` and `VaultModule.burnUsd` functions (and their dependencies) have been moved to new module `IssueUSDModule` in order to reduce the deployment size of the `VaultModule`. The corresponding tests have been relocated as well.


packages/synthetix-main/contracts/interfaces/IIssueUSDModule.sol
```diff
@@ -0,0 +1,62 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../utils/CurvesLibrary.sol";
+
+/// @title Liquidates the collateral for an account in a pool
+interface IIssueUSDModule {
```
noahlitvin, line 7: Unsure if this is right name for this? Maybe `UsdIssuer`, but it also burns...

noahlitvin (approved):

packages/synthetix-main/contracts/interfaces/IIssueUSDModule.sol
dbeal-eth: it always burns.

im sure @ajsantander will come up with something :laughing:

PR #1144, opened 2022-10-25
deeper router restructure

dbeal-eth:
AssociatedSystems, OwnableModule, UpgradeModule, etc. should all be migrated to the new hardhat-router standard before doing further deployments.

This markably increases code readability, which is nice.


ajsantander (approved):
I like it a lot!

Only thing I would recommend is that we bring back mixins only for modifiers, or quick read/check stuff... I.e. OwnableMixin with just modifiers. In most cases, you wouldn't want to inherit Ownable just for the modifier right? You'd give the contract the ability to set its owner, and maybe you only want to use the modifier. I've seen this in synthetix-main too with the RBAC onlyWithPermission; the intention to not have mixins caused some code repetition which I attempt to fix here by reviving the mixin just for modifiers.

PR #1152, opened 2022-11-04
Feat/rewards distribution

Rickk137:
split setRewardsDistribution logic into registerRewardsDistribution

### DIFF
--- a/packages/synthetix-main/contracts/modules/core/RewardsManagerModule.sol
+++ b/packages/synthetix-main/contracts/modules/core/RewardsManagerModule.sol
@@ -28,7 +28,7 @@ contract RewardsManagerModule is IRewardsManagerModule {
     // Associated Rewards
     // ---------------------------------------
 
-    function registerRewardsDistribution(
+    function registerRewardsDistributor(
         uint128 poolId,
         address collateralType,
         uint index,
@@ -64,7 +64,7 @@ contract RewardsManagerModule is IRewardsManagerModule {
         existingDistribution.distributor = IRewardDistributor(distributor);
     }
 
-    function setRewardsDistribution(
+    function distributeRewards(
         uint128 poolId,
         address collateralType,
         uint index,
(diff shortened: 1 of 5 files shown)


packages/synthetix-main/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -28,14 +28,11 @@ contract RewardsManagerModule is IRewardsManagerModule {
     // Associated Rewards
     // ---------------------------------------
 
-    function setRewardsDistribution(
+    function registerRewardsDistribution(
```
noahlitvin, line 31: Maybe we rename to `registerRewardsDistributor`?

packages/synthetix-main/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -64,6 +61,33 @@ contract RewardsManagerModule is IRewardsManagerModule {
             revert InvalidParameters("distributor", "must be non-zero");
         }
 
+        existingDistribution.distributor = IRewardDistributor(distributor);
+    }
+
+    function setRewardsDistribution(
```
noahlitvin, line 67: Maybe we rename to `distributeRewards` (though this might be confused with a payout). Possible `notify` is the right verb here.

packages/synthetix-main/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -28,14 +28,11 @@ contract RewardsManagerModule is IRewardsManagerModule {
...
         address collateralType,
         uint index,
-        address distributor,
-        uint amount,
-        uint start,
-        uint duration
+        address distributor
     ) external override {
         if (index > _MAX_REWARD_DISTRIBUTIONS) {
             revert InvalidParameters("index", "too large");
```
noahlitvin, line 38: Further down, where we can only do 'next index', I follow why it's implemented this way, but I wonder if we can just treat them like slots and allow people to insert wherever? Or have it auto-increment? I can think on this more.

Rickk137:
closing this one:
https://github.com/Synthetixio/synthetix-v3/pull/1156

PR #1170, opened 2022-11-16
Update cannon deployment on Oracle Manager

mjlescano:
This PR updates the `cannonfile.toml` to be compatible with the latest changes introduced in `hardhat-router`.

It also introduces some extra changes:
1. Removes the `contracts/modules/core/` and moves everything to `../modules`, as this project does not need several systems for now (e.g.: `core`, `snx`, etc...) 
2. Adds `OwnerModule` & `UpgradeModule` to make the `Proxy` upgradable & ownable.
3. Adds the `contracts/Dependencies.sol` file with imports of the required contracts from other projects, this is just so we don't have to create a file for each import.
4. Adds upgradability functionality to the `cannonfile.toml`


packages/oracle-manager/hardhat.config.ts
```diff
@@ -1,6 +1,5 @@
 //import '@typechain/hardhat'
 
-import '@nomiclabs/hardhat-ethers';
```
mjlescano, line 3: Not necessary as is already being loaded by the common config :)

dbeal-eth (approved):
:clap:

PR #1171, opened 2022-11-17
Remove InitialModuleBundle dependency from cannonfile.toml

mjlescano:


packages/sample-project/cannonfile.toml
```diff
@@ -2,20 +2,22 @@ name = "synthetix-sample-project"
...
-[contract.OwnerModule]
-artifact = "contracts/modules/OwnerModule.sol:OwnerModule"
+[setting.core_router_version]
+description = "Version of @synthetixio/core-router"
+defaultValue = "1.0.0"
 
-[contract.UpgradeModule]
-artifact = "contracts/modules/UpgradeModule.sol:UpgradeModule"
+[import.core_router]
+source = "core-router:<%= settings.core_router_version %>"
```
dbeal-eth, line 18: would be better to have setting for `[setting.core_router_package]` and inject the whole package name in here rather than individual part

packages/sample-project/test/bootstrap.js
```diff
@@ -1,64 +1,11 @@
-const hre = require('hardhat');
+const { bootstrap } = require('@synthetixio/core-router/util/bootstrap');
 
-let baseSystemSnapshot;
-let provider;
-let signers;
-let outputs;
+const result = bootstrap();
```
dbeal-eth, line 3: nice code cutting!

packages/sample-project/cannonfile.toml
```diff
@@ -49,17 +51,20 @@ depends = [
...
   "InitializableModule",
   "SettingsModule",
   "SomeModule",
   "AnotherModule"
 ]
 depends = ["run.generate_router"]
 
-[contract.Proxy]
-artifact = "contracts/Proxy.sol:Proxy"
+[invoke.upgrade_proxy]
```
dbeal-eth, line 62: i agree with the change, but unfortunately for v3 synthetix-main we have already committed to the deployment steps that we have for the current router, so we can't do an upgrade without completely resetting the proxy. for future projects, this is definitely the approach, but I just want to make sure we are not making any changes to that?

sunnyvempati (approved):
nicee

PR #1174, opened 2022-11-18
Introduces SafeCast

ajsantander:


packages/core-contracts/contracts/mocks/utils/SafeCastMock.sol
```diff
@@ -0,0 +1,40 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../utils/SafeCast.sol";
+
+contract SafeCastMock {
```
dbeal-eth, line 6: this mock could possibly be replaced with generated code by the new generator (it would be a storage library with an empty Data structure)

packages/core-contracts/contracts/utils/SafeCast.sol
```diff
@@ -0,0 +1,73 @@
...
+ * 2) If int256 A = -1, casting A to uint256 returns type(uint256).max.
+ */
+library SafeCast {
+    error CastError(bytes32 fromType, bytes32 toType);
+
+    // Note: Overloading doesn't seem to work for similar types, i.e. int256 and int128, uint256 and uint128, etc, so explicitly naming the functions differently here.
+
+    function uint256toUint128(uint256 x) internal pure returns (uint128) {
+        if (x > type(uint128).max) {
+            revert("Failed cast uint256 to uint128");
```
dbeal-eth, line 18: normally I wouldn't complain about stuff like this, but since this library is giong to be used at the deepest levels of most of our contracts, these strings unfortunaetly add a lot of extra weight, would be better to define a ton of errors instead, the ABIs can take it...

packages/synthetix-main/contracts/storage/Market.sol
```diff
@@ -21,6 +22,10 @@ library Market {
     using DecimalMath for uint256;
     using DecimalMath for int256;
     using DecimalMath for int128;
+    using SafeCast for uint128;
```
dbeal-eth, line 25: thats a lot of `using`!

note: solidity does support function overloading btw, so rather than having function names in the lib like `int256TouInt128()`, it could just be `.toInt()` and the function takes in a `int128` parameter. This would make for much more readable casts.

dbeal-eth (approved):

packages/core-contracts/contracts/mocks/utils/SafeCastMock.sol
ajsantander: Sure. Do you want to do something in particular in this PR, or are you just commenting for later improvements?

packages/core-contracts/contracts/utils/SafeCast.sol
ajsantander: Yeah, I actually tried to use custom errors, but was having trouble with them. Solidity started getting weird with all this low level stuff. Do you mind if we improve the errors in another PR?

packages/synthetix-main/contracts/storage/Market.sol
ajsantander: It does support it, but it is very buggy for similar types. We can improve on it later, raise issues to the Solidity team, etc.

PR #1175, opened 2022-11-18
add minDelegation feature

dbeal-eth:
prevent users from sybilling or creating a ton of fake pools and etc. by
requiring a minimum delegration amount. Note that this minimum amount is
not enforced when the user is setting collateralAmount to 0 (in this case, the system correctly
removes the user/pool/anything from consideration in future
calculations)

notes:
* if the minDelegation amount isnt set, the `liquidationReward` is used
  instead (as this is the lower limit for delegation size as it causes
  problem for the system if the user has too tiny of an account
* *BREAKING INTERFACE CHANGE* CollateralConfiguration now takes in a
  `CollateralConfiguration.Data` object becuaes it really makes sense for
  what the function is doing.
* had to update one of the tests because it was actually below the limit
  already


packages/synthetix-main/contracts/modules/core/VaultModule.sol
```diff
@@ -61,45 +63,18 @@ contract VaultModule is IVaultModule {
...
-            } else if (collateral.pools.contains((uint(poolId)))) {
-                collateral.pools.remove(poolId);
-            }
-
-            vault.currentEpoch().updateAccountPosition(accountId, collateralAmount, leverage);
-
-            // no update for usd because no usd issued
-            collateralPrice = pool.recalculateVaultCollateral(collateralType);
-        }
+        (uint collateralPrice) = _updatePosition(accountId, poolId, collateralType, collateralAmount, oldCollateralAmount, leverage);
```
ajsantander, line 77: Beautiful!

ajsantander (approved):
All good 👍 

Awesome job in splitting up logic in VaultModule.

PR #1176, opened 2022-11-19
symmetric market max debt per share handling

dbeal-eth:
the current `main` will only re-add a pool to a market when it gets pinged by a user in the market.

however, this can lead to a situation where a pool can be drained of its funds if the market is able to control the `reportedDebt` through some manipulation. to do this, the attacker would move the `reportedDebt` such that it surpasses the limit set by the pool, and then "bump" the pool, and then move back below the reportedDebt. After bumping again, The market will then effectively take the difference between the reduced reported debt and the market limit becuase.

To prevent this from happening, this PR changes it so that a pool can be both added and removed from a market by the specified limit.

notes:
* found one bug in `hardhat-router` recent merge which caused fully qualified path to be submitted for a module for some reason, so I added a check to use `_contractName` function to prevent it from spitting out a FQN
* new internal structure in `Market` is now required because an additional property, `liquidityAmount` must be stored for when a pool is re-added to a market. Otherwise, everything functions as before.
* the `Market.distributeDebt` function is now even more narly than before, but I really have no idea how to split it meaninfully in a way that reduces code duplication. The process followed by the first loop and the second loop has some significant alterations. @ajsantander suggestions?

### DIFF
--- a/packages/hardhat-router/src/tasks/deploy.ts
+++ b/packages/hardhat-router/src/tasks/deploy.ts
@@ -82,7 +82,7 @@ task(TASK_DEPLOY, 'Deploys the given modules behind a Proxy + Router architectur
     const modulesData = await deployContracts(contracts, hre);
 
     await hre.run(SUBTASK_GENERATE_ROUTER, {
-      router: /*_contractName(*/router/*)*/,
+      router: /*_contractName(*/ router /*)*/,
       template: routerTemplate,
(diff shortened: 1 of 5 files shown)


dbeal-eth:
closing in favor of fixed v ersion https://github.com/Synthetixio/synthetix-v3/pull/1183

PR #1179, opened 2022-11-22
Use SafeCast in Distribution.sol

ajsantander:


packages/core-contracts/contracts/utils/SafeCast.sol
```diff
@@ -57,6 +58,14 @@ library SafeCast {
         return int256(x);
     }
 
+    function int128toUint128(int128 x) internal pure returns (uint128) {
+        if (x < 0) {
+            revert("Failed cast int128 to uint128");
```
dbeal-eth, line 63: these types of revert strings will likely put us over the contract limit for some of our modules. do you have an idea how to handle this?

dbeal-eth (approved):

PR #1181, opened 2022-11-23
Improvements to SafeCast (Custom errors and more)

ajsantander:


packages/core-contracts/contracts/utils/SafeCast.sol
```diff
@@ -2,70 +2,92 @@
...
     error CastError(bytes32 fromType, bytes32 toType);
 
     // Note: Overloading doesn't seem to work for similar types, i.e. int256 and int128, uint256 and uint128, etc, so explicitly naming the functions differently here.
 
     function uint256toUint128(uint256 x) internal pure returns (uint128) {
+        // -------------------------------o===============================>
+        // -------------------------------o===============>xxxxxxxxxxxxxxxx
         if (x > type(uint128).max) {
-            revert("Failed cast uint256 to uint128");
+            revert CastError("uint256", "uint128");
```
dbeal-eth, line 26: these constants are only better assuming solidity is able to optimize the strings down from repetitions, but that seems doubtful. It might be better to just create one error per failure, ie

```
error InvalidU256ToU128(uint256 invalid);
```

dbeal-eth (changes requested):

packages/core-contracts/contracts/utils/SafeCast.sol
ajsantander: Cool. Implemented.

dbeal-eth: wohoo!

dbeal-eth (approved):

PR #1190, opened 2022-11-30
Remove precision fns

ajsantander:


packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -238,16 +232,16 @@ library Distribution {
      */
     function getActorValueChange(Data storage dist, bytes32 actorId) internal view returns (int valueChange) {
         DistributionActor.Data storage actor = dist.actorInfo[actorId];
-        int128 deltaValuePerShare = dist.valuePerShare - actor.lastValuePerShare;
+        int128 deltaValuePerShare = dist.valuePerShareD27 - actor.lastValuePerShareD27;
 
-        int changedValueHighPrecision = deltaValuePerShare * actor.shares.uint128toInt256();
-        valueChange = changedValueHighPrecision.fromHighPrecisionDecimalToInteger();
+        int changedValueHighPrecision = deltaValuePerShare * actor.sharesD18.uint128toInt256();
```
leomassazza, line 237: mixing D27 and D18 here. Check if maths work

packages/synthetix-main/contracts/storage/DistributionEntry.sol
```diff
@@ -83,14 +83,14 @@ library DistributionEntry {
 
         // determine whether this is an instant distribution or a delayed distribution
         if (entry.duration == 0 && entry.lastUpdate < entry.start) {
-            valuePerShareChange = (int(entry.scheduledValue) * 1e18) / int(totalSharesAmount);
+            valuePerShareChange = (int(entry.scheduledValueD18) * 1e18) / int(totalSharesAmount);
```
leomassazza, line 86: scheduledValueD18.divideDecimal(int(totalSharesAmount)) ?

packages/synthetix-main/contracts/storage/Pool.sol
```diff
@@ -219,10 +219,10 @@ library Pool {
             thing = int(DecimalMath.UNIT); // If minLiquidityRatio is zero, then TODO
```
leomassazza, line 219: thing... really? LoL

packages/synthetix-main/contracts/storage/Distribution.sol
ajsantander: Yep, it goes to D45, then to D18 on the next line

packages/synthetix-main/contracts/storage/Pool.sol
ajsantander: Yes, this has to do with the "passes" strategy. Seems like during the last pass I identified that to be a thing, but not what it is. We can probably rename it properly on the next pass.

PR #1192, opened 2022-11-30
better casting

dbeal-eth:
make casting and extremely easy to reason about by splitting the cast operations into separate libraries for each operation. to allow for casting, add the following `using` calls to the top of your solidity file:

```
    using SafeCastU128 for uint128;
    using SafeCastI128 for int128;
```

then, cast any numeric type to another like `x.to128()` or `x.toInt()`.

2-step casts such as Uint256 -> Int128 should be done with 2 functions like `.toInt().to128()`, but these cases are very rare.

casts from *128 to *256 are implied by solidity so there is no need to create functions for these conversions.


packages/core-contracts/contracts/utils/SafeCast/SafeCastU128.sol
```diff
@@ -0,0 +1,20 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @title Utility that avoids silent overflow errors.
+ *
+ */
+library SafeCastU128 {
+    error OverflowInt128ToUint128();
+
+    function toInt(uint128 x) internal pure returns (int128) {
```
sunnyvempati, line 11:
```suggestion
    function toInt128(uint128 x) internal pure returns (int128) {
```

packages/core-contracts/contracts/utils/SafeCast/SafeCastI128.sol
```diff
@@ -0,0 +1,20 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @title Utility that avoids silent overflow errors.
+ *
+ */
+library SafeCastI128 {
+    error OverflowInt128ToUint128();
+
+    function toUint(int128 x) internal pure returns (uint128) {
```
sunnyvempati, line 11:
```suggestion
    function toUint128(int128 x) internal pure returns (uint128) {
```

sunnyvempati (commented):
couple of suggestions on naming but looks good otherwise.  it would be nice to go from like int256 -> uint128 without the extra call but we can optimize later

ajsantander (approved):

PR #1193, opened 2022-11-30
Split Distribution docs into ScalableMapping

ajsantander:


packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -197,14 +84,15 @@ library Distribution {
      *
      * Whenever an actor's shares are changed in this way, we record the distribution's current valuePerShare into the actor's lastValuePerShare record.
      *
-     * Note: Consider that calling this function wipes out historical value change data. So, make sure to call `getActorValueChange()` and use this data before calling this function.
-     * TODO: Consider reinserting this return value because almost always you NEED to use it.
+     * Returns the the amount by which the actors value changed since the last update.
```
dbeal-eth, line 87: would be clearer to say "since last call to `setActorShares`"

dbeal-eth (approved):

PR #1225, opened 2022-12-01
Todos decimal math

ajsantander:


packages/core-contracts/contracts/mocks/utils/DecimalMathMock.sol
```diff
@@ -16,6 +16,14 @@ contract DecimalMathMock {
         return DecimalMath.divDecimal(x, y);
     }
 
+    function upscale(uint x, uint factor) public pure returns (uint) {
```
leomassazza, line 19: np: use uint256 here or uint in the previous two functions

packages/core-contracts/contracts/mocks/utils/DecimalMathMock.sol
```diff
@@ -40,6 +56,14 @@ contract DecimalMathMock {
         return DecimalMath.divDecimal(x, y);
     }
 
+    function upscale(int x, uint factor) public pure returns (int) {
```
leomassazza, line 59: same np than with int256 vs int

leomassazza (approved):

PR #1230, opened 2022-12-02
add tests (and fix bugs) for pools

dbeal-eth:
add tests to verify that the system imposed max debt limit is sufficiently
protecting the protocol from debt loss.

there were some bugs identified
during this process, but they have been fixed now.

### DIFF
--- a/packages/synthetix-main/contracts/storage/Pool.sol
+++ b/packages/synthetix-main/contracts/storage/Pool.sol
@@ -70,12 +70,10 @@ library Pool {
          * Reciprocally, this pro-rata share also determines how much the pool is exposed to each market's debt.
          */
         uint128 totalWeightsD18;
-
         /**
          * @dev Accumulated cache value of all vault collateral debts
          */
         int128 totalVaultDebtsD18;
-
         /**
          * @dev Array of markets connected to this pool, and their configurations. I.e. weight, etc.
          *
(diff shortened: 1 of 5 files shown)


packages/synthetix-main/contracts/storage/Vault.sol
```diff
@@ -47,7 +47,7 @@ library Vault {
         /**
          * @dev The previous credit capacity of the vault (collateral - debt), when the system was last interacted with.
```
leomassazza, line 48: Comment outdated with old wording

packages/synthetix-main/contracts/storage/Pool.sol
```diff
@@ -192,24 +198,31 @@ library Pool {
...
+        uint creditCapacityD18,
+        int debtD18
     ) internal view returns (int) {
+        Market.Data storage marketData = Market.load(marketId);
         uint minLiquidityRatioD18 = SystemPoolConfiguration.load().minLiquidityRatioD18;
 
         // Calculate the margin of debt that the market could incur to hit the system wide limit.
-        uint marginD18;
+        uint totalSharesD18 = self.vaultsDebtDistribution.totalSharesD18;
+        int valuePerShare = marketData.poolsDebtDistribution.getValuePerShare();
```
leomassazza, line 210: Add Decimal precision to variable name (valuePerShare)

leomassazza (changes requested):
Check variable name nomenclature, and outdated comment. The rest looks great

packages/synthetix-main/contracts/storage/Pool.sol
dbeal-eth: fixed

packages/synthetix-main/contracts/storage/Vault.sol
dbeal-eth: thanks

leomassazza (approved):

PR #1231, opened 2022-12-02
Distribution tests

sunnyvempati:


packages/synthetix-main/contracts/storage/Distribution.sol
```diff
@@ -30,15 +30,6 @@ library Distribution {
...
     error EmptyDistribution();
-    /**
-     * @dev Thrown when an attempt is made to add value to a distribution
-     * whose valuePerShare is zero.
-     */
-    error ZeroValuePerShare();
-    /**
-     * @dev Thrown when a single distribution is used in the two modes mentioned above.
-     */
-    error InconsistentDistribution();
```
sunnyvempati, line 41: not being used

dbeal-eth (approved):
probably could have more tests but this is super helpful, thanks for impl!

packages/synthetix-main/contracts/storage/Distribution.sol
dbeal-eth: thanks for cleaning up :pray:

PR #1238, opened 2022-12-02
Assorted module cleanup

noahlitvin:


packages/synthetix-main/contracts/modules/core/AccountModule.sol
```diff
@@ -53,6 +62,9 @@ contract AccountModule is IAccountModule {
         emit AccountCreated(msg.sender, requestedAccountId);
     }
 
+    /**
+     * @dev Updates storage with the new owner when the account token is transferred
```
dbeal-eth, line 66: called by `AccountTokenModule` system to...

dbeal-eth (approved):
some of the new comments in the `IssueUSDModule` seem unnecessary

PR #1240, opened 2022-12-02
Docs for VaultModule

ajsantander:


packages/synthetix-main/contracts/modules/core/VaultModule.sol
```diff
@@ -167,7 +169,7 @@ contract VaultModule is IVaultModule {
     }
 
     /**
-     * @dev See {IVaultModule-getPositionDebt}.
+     * @dev Returns the debt for the given account, pool, and collateral type.
```
dbeal-eth, line 172: should clarify that this function will also trigger a debt distribution chain update, which is why this is a writable function

dbeal-eth: possible include reccomendation to use `callStatic` if in ethers (is the function users are looking for)

packages/synthetix-main/contracts/modules/core/VaultModule.sol
```diff
@@ -190,43 +192,56 @@ contract VaultModule is IVaultModule {
     }
 
     /**
-     * @dev See {IVaultModule-getVaultDebt}.
+     * @dev Returns the total debt for the given pool, and collateral type.
```
dbeal-eth, line 195: same here, function is mutable becuase it triggers a debt distribution change in order to get the debt

dbeal-eth (approved):

packages/synthetix-main/contracts/modules/core/VaultModule.sol
ajsantander: Implemented

packages/synthetix-main/contracts/modules/core/VaultModule.sol
ajsantander: Implemented

PR #1241, opened 2022-12-02
Rewards manager module clean up, etc.

noahlitvin:


packages/synthetix-main/contracts/modules/core/AssociateDebtModule.sol
```diff
@@ -11,6 +11,11 @@ import "../../storage/Distribution.sol";
 import "../../storage/Pool.sol";
 import "../../storage/Market.sol";
 
+import "@synthetixio/core-modules/contracts/storage/FeatureFlag.sol";
+
+/**
+ * @title System module that allows a market to assign debt to specific positions
```
ajsantander, line 17:
```suggestion
 * @title System module that allows a market to associate debt to a user's position.
 *
 * E.g. when migrating a position from v2 into v3's legacy market, the market first scales up everyone's debt, and then associates it to a position using this module.
```

ajsantander (approved):
Approved with a suggestion

PR #1251, opened 2022-12-03
Events clean up

noahlitvin:


packages/synthetix-main/contracts/interfaces/ILiquidationModule.sol
```diff
@@ -3,13 +3,19 @@ pragma solidity ^0.8.0;
...
+
     event Liquidation(
         uint128 indexed accountId,
         uint128 indexed poolId,
         address indexed collateralType,
-        uint debtLiquidated,
-        uint collateralLiquidated,
-        uint amountRewarded
+        LiquidationData liquidationData,
+        uint liquidateAsAccountId,
```
dbeal-eth, line 17: supposed to be uint128

packages/synthetix-main/contracts/interfaces/ILiquidationModule.sol
```diff
@@ -3,13 +3,19 @@ pragma solidity ^0.8.0;
...
+        address sender
     );
 
     event VaultLiquidation(
```
dbeal-eth, line 21: this structure should be updated to use the same `LiquidationData` structure as well

packages/synthetix-main/contracts/interfaces/IMarketManagerModule.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity ^0.8.0;
 
 /// @title Market Manager Module
 interface IMarketManagerModule {
-    event MarketRegistered(address indexed market, uint128 indexed marketId);
+    event MarketRegistered(address indexed market, uint128 indexed marketId, address indexed sender);
```
dbeal-eth, line 6: feels like sender might make more sense as the first property here

packages/synthetix-main/contracts/interfaces/IPoolModule.sol
```diff
@@ -6,9 +6,9 @@ import "../storage/MarketConfiguration.sol";
 /// @title Module for managing pool token and pools positions distribution
 interface IPoolModule {
     /// @notice gets fired when pool will be created
-    event PoolCreated(uint128 indexed poolId, address indexed owner);
+    event PoolCreated(uint128 indexed poolId, address indexed owner, address indexed sender);
```
dbeal-eth, line 9: same thing, sender probably makes more sense in the front here

dbeal-eth (changes requested):

packages/synthetix-main/contracts/interfaces/IMarketManagerModule.sol
noahlitvin: I think we should keep this consistent with the other events.

leomassazza (approved):

dbeal-eth (approved):

PR #1271, opened 2022-12-07
Core modules docs

ajsantander:


utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
```diff
@@ -1,9 +1,42 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-/// @title Allows for the registration and tracking of auxillery contracts which also follow the proxy architecture
+/**
+ * @title Module for connecting a system with other associated systems.
+ *
+ * Associated systems can me managed or unmanaged.
```
noahlitvin, line 7:
```suggestion
 * Associated systems can be managed or unmanaged.
```

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
```diff
@@ -1,9 +1,42 @@
...
+ *
+ * Managed systems are connected via a proxy, which means that their implementation can be updated, and the system controls the execution context of the associated system. Example, an snxUSD token connected to the system, and controlled by the system.
+ *
+ * Unmanaged systems are just addresses tracked by the system, for which it has no control whatsoever. Example, Uniswap v3, Curve, etc.
+ *
+ * Associated systems, managed or unmanaged, become available to all system modules for communication and interaction, but as opposed to inter-modular communications, associated system interactions will require the use of `CALL`.
+ */
```
noahlitvin, line 14: Maybe explain the concepts of 'kinds' of systems up here? (Token, Nft, maybe we add a 'generic' one?)

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
```diff
@@ -1,9 +1,42 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-/// @title Allows for the registration and tracking of auxillery contracts which also follow the proxy architecture
+/**
+ * @title Module for connecting a system with other associated systems.
+ *
+ * Associated systems can me managed or unmanaged.
+ *
+ * Managed systems are connected via a proxy, which means that their implementation can be updated, and the system controls the execution context of the associated system. Example, an snxUSD token connected to the system, and controlled by the system.
```
noahlitvin, line 9: It specifically has to be this kind of proxy, right?

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
```diff
@@ -1,9 +1,42 @@
...
+
+    /**
+     * @notice Creates or initializes a managed associated ERC20 token.
+     * @param id The bytes32 identifier of the associated system.
+     * @param name The token name that will be used to initialize the proxy.
+     * @param symbol The token symbol that will be used to initialize the proxy.
+     * @param decimals The token decimals that will be used to initialize the proxy.
+     * @param impl The first/new ERC20 implementation of the proxy.
+     *
+     * Note: If the id is new, it will create a new proxy and set its first implementation. Otherwise, it will upgrade the implementation of the existing proxy.
```
noahlitvin, line 38: This is a little confusing because above you say the "first/new" implementation, but here it reads like it could be an existing one?

noahlitvin: Also, these always assume 18 decimals I think? Probably worth mentioning somewhere

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
```diff
@@ -12,6 +45,16 @@ interface IAssociatedSystemsModule {
...
+    /**
+     * @notice Creates or initializes a managed associated ERC721 token.
+     * @param id The bytes32 identifier of the associated system.
+     * @param name The token name that will be used to initialize the proxy.
+     * @param symbol The token symbol that will be used to initialize the proxy.
+     * @param uri The token uri that will be used to initialize the proxy.
+     * @param impl The first/new ERC721 implementation of the proxy.
+     *
+     * Note: If the id is new, it will create a new proxy and set its first implementation. Otherwise, it will upgrade the implementation of the existing proxy.
+     */
```
noahlitvin, line 57: Same as above.

utils/core-modules/contracts/interfaces/IFeatureFlagModule.sol
```diff
@@ -1,23 +1,78 @@
...
+ * If such a line is not present in a function, then it is not controlled by a feature flag.
+ *
+ * If a feature flag is set and then removed forever, consider deleting the line mentioned above from the function's body.
+ */
 interface IFeatureFlagModule {
-    /// @notice Set a feature flag to either allow all or not
+    /**
+     * @notice Emitted when general access has been given or removed for a feature.
+     * @param feature The bytes32 id of the feature.
+     * @param allowAll True if the feature was allowed for everyone and false if it was disallowed for everyone.
```
noahlitvin, line 18: If false it's not really disallowed for everything I think? It's 'only allowed for those in this list'

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
ajsantander: Updated the docs, please review.
I don't think we need a generic new type, `KIND_UNMANAGED` takes care of that.

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
ajsantander: > This is a little confusing because above you say the "first/new" implementation, but here it reads like it could be an existing one?
Made a change, please review.

> Also, these always assume 18 decimals I think? Probably worth mentioning somewhere
This module does not make that assumption, but the system does. Which means I don't think this clarification needs to be made here.

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
ajsantander: Done

utils/core-modules/contracts/interfaces/IAssociatedSystemsModule.sol
ajsantander: The system creates the proxy for you, and it is UUPS yes.
Unmanaged systems can be behind whatever, a proxy of any type, no proxy, etc

utils/core-modules/contracts/interfaces/IFeatureFlagModule.sol
```diff
@@ -1,23 +1,78 @@
...
```
noahlitvin, line 18:
```suggestion
     * @param allowAll True if the feature was allowed for everyone and false if it is only allowed for those included in the allowlist.
```

noahlitvin (approved):

PR #1280, opened 2022-12-08
Fix error in IUSDTokenModule

ajsantander:


protocol/synthetix/contracts/modules/usd/USDTokenModule.sol
```diff
@@ -83,7 +83,7 @@ contract USDTokenModule is ERC20, InitializableMixin, IUSDTokenModule {
 
         ERC20Storage.Data storage store = ERC20Storage.load();
 
-        if (amount < store.allowance[from][spender]) {
+        if (amount > store.allowance[from][spender]) {
```
ajsantander, line 86: This bug in the contract was found by looking at the Anvil DEBUG console, and confirming that the code was indeed failing because of an overflow and not a custom error.

After changing this code, shows a custom error and its signature matches the custom error.

We can't detect custom errors _from another contract_ in tests now, so we have to do it manually for now.

protocol/synthetix/test/integration/modules/core/MarketManagerModule.test.ts
```diff
@@ -74,21 +74,22 @@ describe('MarketManagerModule', function () {
...
+      // TODO (Anvil Custom Errors)
+      // Anvil is failing to detect custom errors when the error is emitted
+      // from a contract which is not the one that the user is interacting with.
+      // We can see the error when running tests with `DEBUG=cannon:cli:rpc npm t`,
+      // but we cannot detect it in these tests.
+      // Error debugged: 'Custom Error 2a1b2dd8:(0x0000000000000000000000000de0b6B3a7640000, false)'
+      // 0x2a1b2dd8 = keccak256("InsufficientAllowance(uint256,uint256)")
+      // Replace this with the commented check above when fixed.
+      // For now we check that it simply reverts.
+      await assertRevert(MockMarket().connect(user1).buySynth(One));
```
ajsantander, line 87: I suggest that we leave these tagged TODOs and don't check for revert reason for now. We do it manually.

noahlitvin (approved):

protocol/synthetix/contracts/modules/usd/USDTokenModule.sol
dbeal-eth: whoops

PR #1295, opened 2022-12-14
Reviewed all storage namespaces

ajsantander:


utils/hardhat-storage/src/internal/dump.ts
```diff
@@ -30,14 +30,24 @@ export async function dumpStorage(
 
     const resultNode = clone(contractNode);
 
+    const constDeclarations = findAll(
+      contractNode,
+      'VariableDeclaration',
+      (node) => node.mutability === 'constant'
+    );
```
ajsantander, line 37: @mjlescano please be aware of this change

PR #1300, opened 2022-12-15
Strict revert checks

ajsantander:


protocol/synthetix/contracts/storage/ScalableMapping.sol
```diff
@@ -45,7 +45,7 @@ library ScalableMapping {
     /**
      * @dev Thrown when attempting to scale a mapping with an amount that is lower than its resolution.
      */
-    error InsufficientMappedAmount(int256 scaleModifier);
```
ajsantander, line 48: This is probably the only significant change to the contracts of this PR.

I removed this value from the error because it is not very informative beyond the logical reason that triggers it, and it is very hard to calculate and produce a value to check against from the tests without adding a bunch of view functions on ScalableMapping.sol without bloating its interface to a point that might be more confusing for devs using it.

protocol/synthetix/test/integration/modules/core/VaultModule.test.ts
```diff
@@ -337,7 +338,7 @@ describe('VaultModule', function () {
             await systems().Core.connect(user2).delegateCollateral(
               user2AccountId,
               poolId,
-              collateralAddress,
+              collateralAddress(),
```
ajsantander, line 341: Caught this one on the go...

utils/core-modules/storage.dump.sol
```diff
@@ -31,139 +31,3 @@ contract ProxyStorage {
         }
     }
 }
-
-// @custom:artifact @synthetixio/core-contracts/contracts/token/ERC20Storage.sol:ERC20Storage
-library ERC20Storage {
-    bytes32 private constant _slotERC20Storage = keccak256(abi.encode("io.synthetix.core-contracts.ERC20"));
```
ajsantander, line 37: Not super sure why core-modules' storage dump was updated. @mjlescano ?

noahlitvin (approved):

PR #1322, opened 2022-12-21
Custom Fee Calculator

noahlitvin:


markets/spot-market/contracts/interfaces/external/ICustomFeeCalculator.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@synthetixio/core-contracts/contracts/interfaces/IERC165.sol";
+import "../../storage/SpotMarketFactory.sol";
+
+interface ICustomFeeCalculator is IERC165 {
```
sunnyvempati, line 7: hmm this is a bit confusing.  i didn't realize you meant we wanted to have a separate system for calculating fees.  i feel like we should just have a contract that we send the fees to and keep the fee calculation logic the same for all markets.

PR #1325, opened 2022-12-23
Introduce fee collector

sunnyvempati:


markets/spot-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -106,15 +106,6 @@ contract AsyncOrderModule is IAsyncOrderModule {
...
         );
 
-        // Calculate fees
-        // TODO: use `amountUsable` instead of `synthAmountEscrowedUsd`?
-        (, int256 feesQuoted) = Fee.calculateFees(
-            marketId,
-            msg.sender,
-            synthAmountEscrowedUsd,
-            SpotMarketFactory.TransactionType.ASYNC_SELL
-        );
```
sunnyvempati, line 116: the `feesQuoted` wasn't really being used here.  but lets revisit async orders in general later.  its a bit unclear what we should escrow and which price to honor (commitment vs settlement)

noahlitvin: In a sell commitment, I think we burn all of the escrowed synths, withdraw their value in usd, and run it through the fee collector to potentially deposit it all back.

We could probably add a conditional here to avoid the gas from the round trip as an optimization.

markets/spot-market/contracts/storage/Fee.sol
```diff
@@ -213,10 +216,23 @@ library Fee {
...
+        if (address(feeCollector) != address(0)) {
+            SpotMarketFactory.Data storage store = SpotMarketFactory.load();
+
+            store.usdToken.approve(address(feeCollector), totalFees);
+            feesCollected = feeCollector.collectFees(totalFees);
+            store.usdToken.approve(address(feeCollector), 0);
+
+            store.depositToMarketManager(marketId, msg.sender, totalFees - feesCollected);
+        }
+    }
```
noahlitvin, line 230: I imagined this function would always end with `depositToMarketManager` (or this functionality would exist in the `depositToMarketManager` function). i.e. we have a 'wrapper' around `depositToMarketManager` (essentially "improve market PnL"), where a custom contract has an opportunity to intercept some of profit before it goes back into the system.

markets/spot-market/contracts/interfaces/external/IFeeCollector.sol
```diff
@@ -0,0 +1,7 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @title Spot Market Interface
+interface IFeeCollector {
```
noahlitvin, line 5: Nitpick: could be good to add ERC165 support here.

markets/spot-market/contracts/modules/AsyncOrderModule.sol
sunnyvempati: let's chat tomorrow about this.  

as a reminder for tomorrow:  the usd amount that is escrowed could be more or less than the amount we'd need based on price during settlement.  lets go through this together; might be missing something.

markets/spot-market/contracts/storage/Fee.sol
sunnyvempati: yea makes sense.  i added `processFees` that can be called, and it will calculate fees, run through fee collector then deposit the rest into market manager all in that call.

noahlitvin (approved):

PR #1326, opened 2022-12-24
ReceiveModule

ajsantander:


ajsantander:
It seems that receive() is not part of the ABI of a compiled contract, so this ReceiveModule approach will probably not work.

PR #1328, opened 2023-01-01
Validate epoch 3

ajsantander:


packages/scripts/data/councils-epoch-3.json
```diff
@@ -0,0 +1,88 @@
+[
+  {
+    "name": "ambassador-council",
+    "owner": "0x6cd3f878852769e04A723A5f66CA7DD4d9E38A6C",
+    "nominatedOwner": "0x0000000000000000000000000000000000000000",
+    "nominationPeriodStartDate": "2023-03-19T02:22:29.000Z",
```
fritzschoff, line 6: isn't that longer now? Remember a request from Matt that periods should be longer now?

PR #1329, opened 2023-01-02
Docstring fixes

noahlitvin:


protocol/oracle-manager/storage.dump.sol
```diff
@@ -64,18 +61,15 @@ library NodeDefinition {
...
+        PYTH
     }
     struct Data {
         bytes32[] parents;
         NodeType nodeType;
         bytes parameters;
     }
     function load(bytes32 id) internal pure returns (Data storage data) {
-        bytes32 s = keccak256(abi.encode("io.synthetix.oracle-manager.Node", id));
+        bytes32 s = keccak256(abi.encode("Node", id));
```
dbeal-eth, line 72: might want to be careful with this name here, its so generic that it could easily clash with another module thats added at a later date that may happen to have the same name. maybe going further to name this `OracleManagerNode` would be safer?

protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
```diff
@@ -139,7 +139,8 @@ contract MarketManagerModule is IMarketManagerModule {
         if (msg.sender != marketData.marketAddress) revert AccessError.Unauthorized(msg.sender);
 
         // Ensure that the market's balance allows for this withdrawal.
-        if (amount > getWithdrawableUsd(marketId)) revert NotEnoughLiquidity(marketId, amount);
+        if (amount > getWithdrawableMarketUsd(marketId))
```
dbeal-eth, line 142: in every case in the codebase tmk we put curly braces around all if statements even if they are only 1 line long

dbeal-eth (approved):

PR #1330, opened 2023-01-02
Adds getMarketCollateralValue()

noahlitvin:


protocol/synthetix/storage.dump.sol
```diff
@@ -67,15 +67,15 @@ library ERC20Storage {
 
 // @custom:artifact @synthetixio/core-contracts/contracts/token/ERC721EnumerableStorage.sol:ERC721EnumerableStorage
 library ERC721EnumerableStorage {
-    bytes32 private constant _slotERC721EnumerableStorage = keccak256(abi.encode("io.synthetix.core-contracts.ERC721Enumerable"));
+    bytes32 private constant _SLOT_ERC721_ENUMERABLE_STORAGE = keccak256(abi.encode("io.synthetix.core-contracts.ERC721Enumerable"));
```
dbeal-eth, line 70: this pattern of having the slot defined as a outside constant is not consistant with what we have been doing elsewhere in the codebase

dbeal-eth (approved):

PR #1338, opened 2023-01-10
👷🏻‍♂️ Collateral lock withdraw fix

JaredBorders:
Current collateral locks do not prevent withdrawal. This PR follows auditor's recommended fix.

## Description
* Add further validation when checking `CollateralModule.withdraw(...) `
* Test validation

## Motivation and Context
I was reading through v3 today and noticed this issue mentioned by the auditor. I decided to take a few minutes to implement a fix to help out 👍

### DIFF
--- a/protocol/synthetix/contracts/modules/core/CollateralModule.sol
+++ b/protocol/synthetix/contracts/modules/core/CollateralModule.sol
@@ -72,7 +72,12 @@ contract CollateralModule is ICollateralModule {
             .load(collateralType)
             .convertTokenToSystemAmount(tokenAmount);
 
-        if (account.collaterals[collateralType].availableAmountD18 < systemAmount) {
+        (uint256 totalDeposited, uint256 totalAssigned, uint256 totalLocked) = account
+            .getCollateralTotals(collateralType);
+        if (
+            totalDeposited - totalAssigned < systemAmount ||
+            totalDeposited - totalLocked < systemAmount
+        ) {
             revert InsufficientAccountCollateral(systemAmount);
         }
 
@@ -86,14 +91,15 @@ contract CollateralModule is ICollateralModule {
     /**
      * @inheritdoc ICollateralModule
      */
-    function getAccountCollateral(
-        uint128 accountId,
-        address collateralType
-    )
+    function getAccountCollateral(uint128 accountId, address collateralType)
         external
         view
         override
-        returns (uint256 totalDeposited, uint256 totalAssigned, uint256 totalLocked)
+        returns (
+            uint256 totalDeposited,
+            uint256 totalAssigned,
+            uint256 totalLocked
+        )
     {
         return Account.load(accountId).getCollateralTotals(collateralType);
     }
@@ -101,10 +107,12 @@ contract CollateralModule is ICollateralModule {
     /**
      * @inheritdoc ICollateralModule
      */
-    function getAccountAvailableCollateral(
-        uint128 accountId,
-        address collateralType
-    ) public view override returns (uint256) {
+    function getAccountAvailableCollateral(uint128 accountId, address collateralType)
+        public
+        view
+        override
+        returns (uint256)
+    {
         return Account.load(accountId).collaterals[collateralType].availableAmountD18;
     }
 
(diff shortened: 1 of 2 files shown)


protocol/synthetix/contracts/modules/core/CollateralModule.sol
```diff
@@ -72,7 +72,12 @@ contract CollateralModule is ICollateralModule {
             .load(collateralType)
             .convertTokenToSystemAmount(tokenAmount);
 
-        if (account.collaterals[collateralType].availableAmountD18 < systemAmount) {
+        (uint256 totalDeposited, uint256 totalAssigned, uint256 totalLocked) = account
+            .getCollateralTotals(collateralType);
+        if (
+            totalDeposited - totalAssigned < systemAmount ||
```
dbeal-eth, line 78: this condition should probably just be `if (totalDeposited - totalAssigned - totalLocked)` because `totalDeposited` is literally `totalAssignedD18 + self.collaterals[collateralType].availableAmountD18`

protocol/synthetix/test/integration/modules/core/CollateralModule/CollateralModule.deposit.test.ts
```diff
@@ -147,6 +148,32 @@ describe('CollateralModule', function () {
...
             });
 
+            describe('when attempting to withdraw locked collateral', () => {
+              var lockedTill;
+              it('get current time', async () => {
+                const currentTime = await getTime(provider());
+                lockedTill = currentTime + 60 * 60 * 24 * 30;
+              });
+              it('reverts', async () => {
+                await systems()
```
dbeal-eth, line 158: in accordance with our testing patterns, please move this call into a separate `before` section

protocol/synthetix/test/integration/modules/core/CollateralModule/CollateralModule.deposit.test.ts
```diff
@@ -147,6 +148,32 @@ describe('CollateralModule', function () {
...
+                // so scale up the expected amount.
+                const errorAmount = depositAmount.mul(ethers.BigNumber.from(10).pow(12));
+
+                await assertRevert(
+                  systems().Core.connect(user1).withdraw(1, Collateral.address, depositAmount),
+                  `InsufficientAccountCollateral("${errorAmount}")`,
+                  systems().Core
+                );
+              });
+              it('unlock collateral', async () => {
```
dbeal-eth, line 172: this test has no conditions validating that the collateral was actually unlocked

also this should go in a nested `describe` with a `before` section running the fast forward

dbeal-eth (changes requested):
Hi! Thanks for contributing to Synthetix v3! We greatly appreciate quality community contributions such as this.

A few minor fixes necessary.

Also, I noticed that there are a lot of extra changes in the `CollateralModule` file for moving parts of function definitions into newlines etc. Did you need to do that for a linter error or something?

dbeal-eth:
looks like CI pipeline is mostly not executing, will need to verify what is going on with that before this can be merged. sorry!

protocol/synthetix/contracts/modules/core/CollateralModule.sol
JaredBorders: I see. Will update.

dbeal-eth: Actually, loading `availableAmount` from storage a second time is more expensive than reading memory variables and extra math op

dbeal-eth: I just realized that locked collateral can also be assigned. So really, it should be the greater of the two `totalLocked` and `totalAssigned` is what we subtract by

JaredBorders: Is all locked collateral not assigned?
ref:
```
@return totalAssigned The amount of collateral in the account that is delegated to pools, denominated with 18 decimals of precision.
@return totalLocked The amount of collateral in the account that cannot currently be undelegated from a pool, denominated with 18 decimals of precision.
```

JaredBorders: if so:
```
if (totalDeposited - totalLocked < systemAmount) {
            revert InsufficientAccountCollateral(systemAmount);
        }
```

protocol/synthetix/test/integration/modules/core/CollateralModule/CollateralModule.deposit.test.ts
JaredBorders: (If you did not fast-forward, then the following withdraw would have failed)

JaredBorders:
Should be good to go now 👍

ajsantander (approved):
Thanks for the contribution @JaredBorders !!

dbeal-eth (approved):

PR #1344, opened 2023-01-12
Handle plain ETH transfers

ajsantander:
As pointed out by auditors, plain ETH transfers hit the receive function and then execution gets forwarded to the routing table, where no selector match is found, thus always causing reverts in plain ETH transfers.

This PR removes the receive function from our routers and explicitly declares that they cannot receive ETH transfers.

However, this PR leaves a half finished implementation in the router to enable them eventually, which simply adds an empty receive function. Note that this receive does not forward execution to the routing table.

Finally, the body of the routing table is moved into the fallback function to save a bit of gas, since it will not be reached from other entry points with receive enabled or disabled.

### DIFF
--- a/utils/core-modules/contracts/interfaces/ISampleOwnedModule.sol
+++ b/utils/core-modules/contracts/interfaces/ISampleOwnedModule.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.0;
 
 interface ISampleOwnedModule {
-    function setProtectedValue(uint newProtectedValue) external;
+    function setProtectedValue(uint newProtectedValue) external payable;
 
     function getProtectedValue() external view returns (uint);
 }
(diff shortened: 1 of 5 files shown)


utils/core-modules/contracts/interfaces/ISampleOwnedModule.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.0;
 
 interface ISampleOwnedModule {
-    function setProtectedValue(uint newProtectedValue) external;
+    function setProtectedValue(uint newProtectedValue) external payable;
```
ajsantander, line 5: Used for testing transactions with `msg.value` > 0.

dbeal-eth (approved):
there are probably better ways this could be done but :+1:

PR #1347, opened 2023-01-16
Locks review

ajsantander:


protocol/synthetix/contracts/modules/core/CollateralModule.sol
```diff
@@ -68,20 +68,23 @@ contract CollateralModule is ICollateralModule {
...
         (uint256 totalDeposited, uint256 totalAssigned, uint256 totalLocked) = account
             .getCollateralTotals(collateralType);
 
-        // neither totalAssigned collateral nor totalLocked is available to be withdrawn
-        uint256 unavailableCollateral = totalAssigned > totalLocked ? totalAssigned : totalLocked;
-        if (totalDeposited - unavailableCollateral < systemAmount) {
-            revert InsufficientAccountCollateral(systemAmount);
+        // The amount that cannot be withdrawn from the protocol is the max of either
+        // locked collateral or delegated collateral.
+        uint256 unavailableCollateral = totalLocked > totalAssigned ? totalLocked : totalAssigned;
```
dbeal-eth, line 80: inconsistant usage of D18 for local variables

dbeal-eth (approved):

protocol/synthetix/contracts/modules/core/CollateralModule.sol
ajsantander: Yeah, it was previously called `systemAmount` which was a bit confusing because it evoked some kind of system quantity as opposed to the token amount that the user wants to withdraw.

And `tokenAmountSystemDenominated` sounds horrible. That's why I went with `tokenAmountD18`.

PR #1351, opened 2023-01-17
removed unsued contract imports

fritzschoff:


protocol/synthetix/contracts/storage/AccountRBAC.sol
```diff
@@ -23,6 +21,11 @@ library AccountRBAC {
     bytes32 internal constant _MINT_PERMISSION = "MINT";
     bytes32 internal constant _REWARDS_PERMISSION = "REWARDS";
 
+    /**
+     * @dev Thrown when a permission specified by a user does not exist or is invalid.
+     */
+    error InvalidPermission(bytes32 permission);
+
```
fritzschoff, line 28: moved error down because otherwise AccountRBAC would need to import Account.sol and that would make it a circular dependency

ajsantander: I think that the circular dependency is ok in this case. I would rather not duplicate code (an error definition in this case).

But I think you can leave it like this, since another PR is moving the error from Account to AccountRBAC anyways. The important thing is that we never duplicate any code.

PR #1359, opened 2023-01-18
Return Account obj in onlyWithPermission

ajsantander:


protocol/synthetix/contracts/storage/Account.sol
```diff
@@ -137,11 +137,19 @@ library Account {
...
+        uint128 accountId,
+        bytes32 permission
+    ) internal view returns (Data storage account) {
+        account = Account.load(accountId);
+
+        if (!account.rbac.authorized(permission, msg.sender)) {
             revert PermissionDenied(accountId, permission, msg.sender);
         }
+
+        return account;
```
sunnyvempati, line 152: don't think you need this since you're assigning `account` on line 146

sunnyvempati (approved):
one small optimization but 👍🏽

noahlitvin (approved):

noahlitvin (approved):

PR #1360, opened 2023-01-18
Renaming usages of self

ajsantander:


dbeal-eth (approved):
using `self` everywhere would be better since that is closer to the object oriented approach seen in other programming languages (ex. with `this`)

In newer programming languages, such as `Rust` `self` is literally the default name and the way that structures are designed to be object oriented is with a pattern very similar to this, so that is the inspiration for the `self` naming (see https://doc.rust-lang.org/rust-by-example/fn/methods.html)

ajsantander:
Decided with team to go with a different approach, so closing PR.

PR #1369, opened 2023-01-20
Wrapper changes / tests

sunnyvempati:
- wrapper fees changed to int's as they can be negative.
- removed `wrappingEnabled` and setup a local supply cap that can be set to 0 if we need to disable.
- unwrapping is still allowed and the supply cap doesn't affect it.
- lots of tests; including negative fee tests.

### DIFF
--- a/markets/spot-market/contracts/storage/Wrapper.sol
+++ b/markets/spot-market/contracts/storage/Wrapper.sol
@@ -7,7 +7,7 @@ import "@synthetixio/core-modules/contracts/interfaces/ITokenModule.sol";
  * @title Wrapper library servicing the wrapper module
  */
 library Wrapper {
-    error WrappingNotInitialized();
+    error InvalidCollateralType();
 
     struct Data {
         /**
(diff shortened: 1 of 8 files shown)


markets/spot-market/contracts/storage/Wrapper.sol
```diff
@@ -16,9 +16,9 @@ library Wrapper {
          */
         address collateralType;
```
noahlitvin, line 17: Maybe we rename to `wrapCollateralType` or something like that?

markets/spot-market/contracts/storage/Wrapper.sol
```diff
@@ -16,9 +16,9 @@ library Wrapper {
...
         /**
-         * @dev sets if wrapping is enabled for given market id
+         * @dev amount of collateral that can be wrapped
          */
-        bool wrappingEnabled;
+        uint256 supplyCap;
```
noahlitvin, line 21: Unsure if this is right name for the variable. I think the market collateral module calls it "max depositable", Maybe this is like `maxWrappableAmount`?

JvRLmV (approved):

noahlitvin (approved):

PR #1374, opened 2023-01-24
audit feedback: fix IssueUSDModule debt bug

dbeal-eth:
in audit feedback submitted by iosiro, it is possible to

it turns out this issue was related to insufficient testing of the system collateralization ratio check, which should have been deducting the balance of the system from the available credit capacity. thankfully, this was easily resolved from a solidity perspective, it was just a missing calculation in one branch.

unfortunately the fix uncovered a few other faulty tests which also had to be resolved, which is part of the reason why there are changes all over the tests for various reasons.

iosiro's provided test has been altered slightly and brought into the codebase to ensure we have strong coverage of this particular bug in the future.

cc @iosiro-security

### DIFF
--- a/protocol/synthetix/contracts/storage/Pool.sol
+++ b/protocol/synthetix/contracts/storage/Pool.sol
@@ -213,6 +213,7 @@ library Pool {
         uint256 totalSharesD18 = self.vaultsDebtDistribution.totalSharesD18;
         int256 valuePerShareD18 = marketData.poolsDebtDistribution.getValuePerShare();
 
+        // solhint-disable-next-line numcast/safe-cast
         int256 debtPerShareD18 = debtD18 > 0 ? debtD18.divDecimal(totalSharesD18.toInt()) : int(0);
 
         if (minLiquidityRatioD18 == 0) {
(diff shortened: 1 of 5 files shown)


protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -414,11 +417,10 @@ library Pool {
...
         int256 getPositionDebtD18 = updateAccountDebt(self, collateralType, accountId);
 
-        // if they have a credit, just treat their debt as 0
         return
-            getPositionCollateralValueD18.divDecimal(
-                getPositionDebtD18 < 0 ? 0 : getPositionDebtD18.toUint()
-            );
+            getPositionDebtD18 > 0
+                ? getPositionCollateralValueD18.divDecimal(getPositionDebtD18.toUint())
+                : 1e20;
```
ajsantander, line 423: Just FYI, I am working on another PR that returns Max uint 256 here in the case of infinite cratio, and does so earlier in this code block to avoid getting the collateral value. I think we can merge this and I'll update the other PR.

dbeal-eth: :+1:

PR #1379, opened 2023-01-24
Fixes rewards permission

ajsantander:


protocol/synthetix/contracts/storage/AccountRBAC.sol
```diff
@@ -38,6 +43,23 @@ library AccountRBAC {
         SetUtil.AddressSet permissionAddresses;
     }
 
+    /**
+     * @dev Reverts if the specified permission is unknown to the account RBAC system.
+     */
+    // Note: Disabling Solidity warning, not sure why it suggests pure mutability.
```
sunnyvempati, line 49: i think this _can_ be pure.  as long as it doesn't read/modify state, reverting can happen.  we're only checking against constants not storage

sunnyvempati (approved):
one small nit but pre-approving

protocol/synthetix/contracts/storage/AccountRBAC.sol
ajsantander: Yeah, I think the tests were having trouble with `pure` here. Will make the change nonetheless and see what happens.

PR #1380, opened 2023-01-24
Fix post transfer hook in ERC721

ajsantander:


sunnyvempati (approved):
why does this matter exactly?  (mostly for my knowledge)

kylriley:
> why does this matter exactly? (mostly for my knowledge)

The ERC721.sol contract has been ported from other generic implementations to be compatible with this proxy pattern (i.e. it might be used for other ERC721 applications than the Accounts Module in the future). As the implementation includes pre- and post-transfer hooks, as a developer reusing it, it's a fairly safe assumption that these hooks should be called during mint and burn operations as they inherently transfer the token. Another option would be to assume that neither hook is called, but it would be quite unintuitive for the pre-transfer hook to be called and not the post-transfer (as it currently does), as that's only a specific requirement of the Accounts Module implementation. 

See OZ [ERC721.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L285) implementation for reference.

PR #1383, opened 2023-01-25
Rename self v2

ajsantander:


sunnyvempati (approved):
this is great, thanks for doing that for spot market as well.

aren't there other modules that load from storage that are probably using variable names like `store`?  I noticed only the storage files were fixed and not the modules.  

only noting because you did fix the storage names in modules for spot market.

PR #1384, opened 2023-01-25
Document and optimize getSystemMaxValuePerShare

ajsantander:


protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -194,44 +194,54 @@ library Pool {
     }
 
     /**
-     * @dev Implements a system-wide fail safe to prevent a market from taking too much debt.
+     * @dev Determines the resulting maximum value per share for a market, according to a system-wide minimum liquidity ratio. This prevents markets from taking too much debt.
```
dbeal-eth, line 197: this prevents markets from assigning more debt to pools than they have collateral to cover

would be a better way to say it

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -194,44 +194,54 @@ library Pool {
...
      *
-     * Note: There is a non-system-wide fail safe for each market at `MarketConfiguration.maxDebtShareValue`.
+     * Note: There is a market-wide fail safe for each market at `MarketConfiguration.maxDebtShareValue`. The lower of the two values should be used.
      *
      * See `SystemPoolConfiguration.minLiquidityRatio`.
      */
     function getSystemMaxValuePerShare(
         Data storage self,
-        uint128 marketId, // we are keeping this value as a uint128 for now for testability reasons
+        uint128 marketId, // Note: We are keeping this value as a uint128 for now for testability reasons.
```
dbeal-eth, line 205: im not sure this comment is relevant any more?

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -194,44 +194,54 @@ library Pool {
...
-        // Calculate the margin of debt that the market could incur to hit the system wide limit.
+        // Retrieve the total shares of the pool's debt distribution.
         uint256 totalSharesD18 = self.vaultsDebtDistribution.totalSharesD18;
+
+        // Retrieve the current value per share of the market.
+        Market.Data storage marketData = Market.load(marketId);
         int256 valuePerShareD18 = marketData.poolsDebtDistribution.getValuePerShare();
 
+        // If there are no shares in the pool's debt distribution,
+        // then the maximum value per share is the market's current value per share.
```
dbeal-eth, line 220: it should probably actually be negative infinity because if the pool has no shares at all, that means it can never, ever, provide liquidity to the pool

dbeal-eth (approved):
minor updates to comments and one more thing to look at on the code side

protocol/synthetix/contracts/storage/Pool.sol
ajsantander: Ok, will track this as a separate issue if it implies significant code changes, and not just docs and clean ups as this PR does.

PR #1385, opened 2023-01-25
Macro Audit Oracle Manager

Rickk137:
https://www.notion.so/0xmacro/Synthetix-2-Preliminary-Audit-Report-external-private-fd8eeb903e8446b2a8f53767802ebd14#1b4b19e94c1e4d7ba4c939dd41233ae7

### DIFF
--- a/protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
+++ b/protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
@@ -13,7 +13,7 @@ library StalenessCircuitBreakerNode {
     ) internal view returns (NodeOutput.Data memory) {
         uint256 stalenessTolerance = abi.decode(parameters, (uint256));
 
-        if (block.timestamp - parentNodeOutputs[0].timestamp < stalenessTolerance) {
+        if (block.timestamp - parentNodeOutputs[0].timestamp <= stalenessTolerance) {
             return parentNodeOutputs[0];
         } else if (parentNodeOutputs.length == 1 || parentNodeOutputs[1].price == 0) {
             revert StalenessToleranceExceeded();


protocol/oracle-manager/README.md
```diff
@@ -42,6 +42,8 @@ The Uniswap Node retrieves data from a [Uniswap Oracle](https://docs.uniswap.org
 - Parameters:
   - `address tokenAddress` - The address of the token
   - `address stablecoinAddress` - The address of the stablecoin
+  - `uint8 tokenDecimals` - The number of decimals places used by the token contract. _This must match what is provided by the token contract's `decimals()` function_.
+  - `uint8 stablecoinDecimals` - The number of decimals places used by the stablecoin contract. _This must match what is provided by the stablecoin contract's `decimals()` function_.
```
dbeal-eth, line 46: is it always the case that this node has to use a `stablecoin` address? What if you want to get a uniswap price between two non-stablecoins?

in uniswap, they use a different strategy: they call it `token0` and `token1`. Perhaps we should follow the same pattern here?

noahlitvin (approved):

protocol/oracle-manager/contracts/nodes/UniswapNode.sol
```diff
@@ -16,12 +18,21 @@ library UniswapNode {
     using SafeCastU56 for uint56;
     using SafeCastU32 for uint32;
     using SafeCastI56 for int56;
+    using SafeCastI256 for int256;
+
+    using DecimalMath for int256;
+
+    uint8 public constant PRECISION = 18;
```
dbeal-eth, line 25: I believedata for what precision we are targeting is also stored in the `DecimalMath` library

dbeal-eth (changes requested):

protocol/oracle-manager/README.md
noahlitvin: But we need to resolve it to a denominated value with 18 decimal precision?

noahlitvin (approved):

PR #1386, opened 2023-01-26
Upgrade election module

dbeal-eth:
Its time to bring the election module back to the present!

This adds the `election-module` project to the `protocol` folder, and modernizes the storage, modules, and cannon build to the latest.

The cannon build is verified to produce a valid initial state of the election module (which is all I *personally* need right now)

The tests currently do not work. I have not testsed them. Very likely there are stupid mistakes in a few p laces in the migrated code because I was going fast.


protocol/election-module/contracts/storage/DebtShare.sol
```diff
@@ -0,0 +1,25 @@
...
+        // Array of debt share snapshot id's for each epoch
+        uint128[] debtShareIds;
+        // Array of CrossChainDebtShareData's for each epoch
+        CrossChainDebtShare.Data[] crossChainDebtShareData;
+    }
+
+    function load() internal pure returns (Data storage debtShare) {
+        assembly {
+            // bytes32(uint(keccak256("io.synthetix.debtshare")) - 1)
+            debtShare.slot := 0x24dbf425c80a2b812a860ebf3bf1d082b94299e66be3feb971f862ad0811d2b8
```
ajsantander, line 22: This should probably be calculated as a constant, as it is done elsewhere in the current v3 code.

protocol/election-module/cannonfile.toml
```diff
@@ -0,0 +1,125 @@
+name = "synthetix-election-module"
```
ajsantander, line 1: How does this account to the fact that there are 4 councils: spartan, treasury, grants, and ambassador

protocol/election-module/cannonfile.toml
```diff
@@ -0,0 +1,125 @@
...
+defaultValue = "90" # days
+
+[setting.voting_period_duration]
+defaultValue = "7" # days
+
+[contract.InitialModuleBundle]
+artifact = "InitialModuleBundle"
+create2 = true
+
+[contract.InitialProxy]
```
ajsantander, line 33: Does this mean that the previous proxies are deprecated (and thus the history of past elections)?

protocol/election-module/contracts/submodules/election/ElectionBase.sol
```diff
@@ -0,0 +1,61 @@
...
+
+import "../../storage/Council.sol";
+import "../../storage/Election.sol";
+
+/// @dev Common utils, errors, and events to be used by any contracts that conform the ElectionModule
+contract ElectionBase {
+    // ---------------------------------------
+    // Enums
+    // ---------------------------------------
+
```
ajsantander, line 12: Probably should delete.

ajsantander (approved):

ajsantander:
We also need to bring back the scripts folder, which we use to verify a council once its deployed.

protocol/election-module/cannonfile.toml
dbeal-eth: you deploy one instance of this cannonfile for every council with a different salt.

If one council requires additional modules or etc. then an extension cannonfile should be created.

protocol/election-module/cannonfile.toml
dbeal-eth: becuaes we are using `create2` unless we change the proxy code in the future, we should not need to migrate again.

protocol/election-module/cannonfile.toml
ajsantander: Sweet

protocol/election-module/cannonfile.toml
ajsantander: Cool, thanks guys

PR #1388, opened 2023-01-26
Iosiro 5.4.4 - remove rewardId from storage

fritzschoff:


protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -235,11 +235,12 @@ contract RewardsManagerModule is IRewardsManagerModule {
             revert ParameterError.InvalidParameter("distributor", "is not registered");
         }
 
-        rewardIds.remove(rewardId);
+        pool.vaults[collateralType].rewardIds.remove(rewardId);
```
ajsantander, line 238: This has the same effect of the line that was removed, and actually costs a little more gas because it does not use the pointer declared in line 226.

I think what the auditors meant was to identify the RewardDistribution object to which the rewardId in question refers to, and resetting that. Since `_getRewardId` produces an id that is determined by `poolId, collateralType, msg.sender`, a future reward id could use the same id of one that was removed, and thus be initialized with pre-existing data.

My recommendation would be to introduce a `resetEntry` function in RewardDistribution, which would set all of this object's properties to zero.

ajsantander (approved):
Please see my comment. I think that the auditors meant a different change.

ajsantander (changes requested):
Please see my comment, I think that the auditors mean a different change.

protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -242,6 +242,10 @@ contract RewardsManagerModule is IRewardsManagerModule {
         }
         pool.vaults[collateralType].rewards[rewardId].distributor = IRewardDistributor(address(0));
 
+        RewardDistribution.Data storage reward = pool.vaults[collateralType].rewards[rewardId];
+
+        reward.resetEntry();
```
fritzschoff, line 247: wondering why I don't have to specify an argument here?

protocol/synthetix/contracts/storage/RewardDistribution.sol
```diff
@@ -172,4 +172,14 @@ library RewardDistribution {
...
+
+    function resetEntry(Data storage entry) internal {
+        entry.distributor = IRewardDistributor(address(0));
+        entry.__slotAvailableForFutureUse = 0;
+        entry.rewardPerShareD18 = 0;
+        entry.scheduledValueD18 = 0;
+        entry.start = 0;
+        entry.duration = 0;
+        entry.lastUpdate = 0;
+    }
```
fritzschoff, line 184: one down side is I can't reset the mapping because I don't know the keys

protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
ajsantander: It's because we're using it as a library: https://www.tutorialspoint.com/solidity/solidity_libraries.htm

Basically, once you use the `using` keyword, all library functions in the file don't need the first argument.

protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -242,6 +242,10 @@ contract RewardsManagerModule is IRewardsManagerModule {
         }
         pool.vaults[collateralType].rewards[rewardId].distributor = IRewardDistributor(address(0));
 
+        RewardDistribution.Data storage reward = pool.vaults[collateralType].rewards[rewardId];
```
ajsantander, line 245: I think that this can be put above line 243, and we can use the storage pointer in that line too. That way we can save a storage read, which are really gas expensive.

protocol/synthetix/contracts/storage/RewardDistribution.sol
```diff
@@ -172,4 +172,14 @@ library RewardDistribution {
 
         return valuePerShareChangeD18;
     }
+
+    function resetEntry(Data storage entry) internal {
+        entry.distributor = IRewardDistributor(address(0));
+        entry.__slotAvailableForFutureUse = 0;
```
ajsantander, line 178: We can probably not reset this one since it will most likely be zero anyway.

protocol/synthetix/contracts/storage/RewardDistribution.sol
ajsantander: Yeah, this is a very good point! Let's discuss off-git and decide.

ajsantander:
Deprecated in favor of https://github.com/Synthetixio/synthetix-v3/pull/1439

PR #1389, opened 2023-01-26
(H-5) - Fix storage slot value check

mjlescano:


ajsantander (changes requested):
Looking good! However, I think this needs two things before we can merge:
1) Ensure this validation is properly connected to the router plugin.
2) If you `npx hardhat storage:verify` in protocol/synthetix, you will see errors due to some storage slots being dynamic. E.g. load in Pool, Market, etc.

ajsantander (approved):

PR #1401, opened 2023-01-31
Simplify pool removal code

ajsantander:


protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -240,11 +238,12 @@ contract VaultModule is IVaultModule {
             collateral.increaseAvailableCollateral(oldCollateralAmount - newCollateralAmount);
         }
 
-        // If the collateral amount is positive, make sure that the pool exists
+        // If the collateral amount not negative, make sure that the pool exists
```
fritzschoff, line 241: `is` is missing

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -240,11 +238,12 @@ contract VaultModule is IVaultModule {
             collateral.increaseAvailableCollateral(oldCollateralAmount - newCollateralAmount);
         }
 
-        // If the collateral amount is positive, make sure that the pool exists
+        // If the collateral amount is not negative, make sure that the pool exists
         // in the collateral entry's pool array. Otherwise remove it.
-        if (newCollateralAmount > 0 && !collateral.pools.contains(poolId)) {
+        bool containsPool = collateral.pools.contains(poolId);
+        if (newCollateralAmount >= 0 && !containsPool) {
```
dbeal-eth, line 244: im confused as to how this is a "simplify" or improves from the code that we had before. can you explain the reasoning behind this change?

dbeal-eth (changes requested):

protocol/synthetix/contracts/modules/core/VaultModule.sol
ajsantander: Sure.

One simplification comes from reducing storage reads. Instead of using `collateral.pools.contains` twice, we do it once and save it on a local var in the stack.

Another simplification comes from the removal of an entire function (`_ensureAccountCollateralsContainsPool`)

The change was suggested by our auditors, and avoids removing and re-adding the pool in certain edge cases.

dbeal-eth (approved):

PR #1406, opened 2023-02-01
🧮 Improve system token scaling 

JaredBorders:
`CollateralConfiguration.ConvertTokenToSystemAmount()` needs to account for ERC20 tokens that do not have the decimals() function or tokens with zero decimals. 

The `CollateralConfigurationModule.configureCollateral()` function should validate the decimals property when adding a new collateral type.

## Description
* Implement try/catch to handle when the ERC20 `decimals()` extension is absent
* Implement validation when configuring collateral (i.e. when adding a new collateral type)
* Test `ConvertTokenToSystemAmount()` with an ERC20 token where `decimals()` returns `0`
* Test `ConvertTokenToSystemAmount()` with an ERC20 token with absent `decimals()` extension
* Test `ConvertTokenToSystemAmount()` with an ERC20 token with absent `decimals()` extension and a fallback function that returns *incompatible* data (i.e. not uint8)
* Test `ConvertTokenToSystemAmount()` with an ERC20 token with absent `decimals()` extension and a fallback function that returns *compatible* data (i.e. uint8)
* Test `configureCollateral()` change

## Related issue(s)
N/A

## Motivation and Context
Addresses Iosiro audit feedback item 5.3.1


ajsantander (approved):

dbeal-eth (changes requested):
@JaredBorders could you please add tests to verify this functionality? you can do so in the `CollateralConfiguration.test.ts` (currently doesnt exist in `main`, here is an example [protocol/synthetix/test/integration/storage/CollateralConfiguration.test.ts](https://github.com/Synthetixio/synthetix-v3/pull/1421/files#diff-8e330e92400382a2d90b74509a6f0975b862f657c65ba9504723b8da6ffaa769R1) ). As this is a view function, this should be a relatively easy test impl. Let us know if you are having trouble running the tests or setting up your tests with this.

Also, we recently received audit feedback informing us that this function does not correctly handle tokens with `decimals() > 18`. Could you verify this and add the necessary test there as well?

Thanks!

JaredBorders:
> 
Well, do we want to allow the system to support decimals > 18?
ex: 
say `decimals = 20` and `amount = x`
then `(x * 10^18) / 10^20 = x  / 10^2`
given above, if `x` is less than `10^2` then it just rounds to zero.
-
Is the above behavior expected? I'd recommend throwing error if decimals > 18

dbeal-eth:
yes, that behavior is expected. in this case, the system would accumulate a very small amount of dust, but there is a possibility we want to support tokens with a higher number of decimals in the future, and if that happens, there isn't any reason to block them from being supported just because we loose an extremely small amount of precision

dbeal-eth:
oh and btw, I guess for ultimate implementation reasons, we could probably have the function revert if precision would be lost. So for example, if the system precision is 2 decimals but the token's is 4, specifying a converted amount of 1.234 would revert because the 0.004 is a precision loss. This would prevent people from depositing a collateral amount that would result in precision loss in the system. Then the problem would go away entirely right?

JaredBorders:
Yes, I think that would work. 

So, in the example I provided above, if `x < 10^2` OR if there are any non-zero decimals at position 19 and 20 the call reverts.
i.e.
amount = 111115555511111555551 would revert bc post conversion 1.111155555111115555(51) 5 and 1 are non-zero?

JaredBorders:
Also, regarding creating a new test specifically for that library: Could you provide some code to help me set that up? Something to expose functions defined in that library anyway

JaredBorders:
Okay @dbeal-eth I just push a logic change to handle when precision is lost

dbeal-eth:
> Also, regarding creating a new test specifically for that library: Could you provide some code to help me set that up? Something to expose functions defined in that library anyway

please check discord friend requests

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
```diff
@@ -215,17 +220,33 @@ library CollateralConfiguration {
...
+        try IERC20(self.tokenAddress).decimals() returns (uint8 decimals) {
+            /// @dev if decimals() is not defined, but an existing fallback function returns a
+            /// bytes representation of uint8, this logic may have unexpected results
+            if (decimals == 0) {
+                amountD18 = tokenAmount;
+            } else if (decimals <= 18) {
+                amountD18 = (tokenAmount * DecimalMath.UNIT) / (10 ** decimals);
+            } else {
+                // ensure no precision is lost when converting to 18 decimals
+                if (tokenAmount % (10 ** DecimalMath.UNIT) != 0) {
```
dbeal-eth, line 243: if i'm not mistaken should be `tokenAmount % (10 ** (decimals - DecimalMath.UNIT))`

i'm sure you will pick this up in the tests tho

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
```diff
@@ -215,17 +220,33 @@ library CollateralConfiguration {
...
         // this extra condition is to prevent potentially malicious untrusted code from being executed on the next statement
         if (self.tokenAddress == address(0)) {
             revert CollateralNotFound();
         }
 
-        return (tokenAmount * DecimalMath.UNIT) / (10 ** IERC20(self.tokenAddress).decimals());
+        try IERC20(self.tokenAddress).decimals() returns (uint8 decimals) {
+            /// @dev if decimals() is not defined, but an existing fallback function returns a
+            /// bytes representation of uint8, this logic may have unexpected results
+            if (decimals == 0) {
```
dbeal-eth, line 237: actually its technically possible for a token to have decimals = 0. For example, if the token exists as only whole units and ex 1 token = `1` unit. So I think it would be best to handle this the same way as any other upscale

also `decimals == 0` doesn't mean `undefined`

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
```diff
@@ -215,17 +220,33 @@ library CollateralConfiguration {
...
+                amountD18 = tokenAmount;
+            } else if (decimals <= 18) {
```
dbeal-eth, line 239: technially this is not the most gas efficient if `decimals == 18`. might be best to just see if `decimals == 18` at the beginning of the function and return early if so

dbeal-eth (changes requested):

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
JaredBorders: `tokenAmount % (10 ** (decimals - 18))` but yeah, that is correct

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
JaredBorders: good idea! early exit is always better

JaredBorders:
@dbeal-eth @ajsantander ready for review 👍

protocol/synthetix/cannonfile.toml
```diff
@@ -12,7 +12,7 @@ description = "Admin user for the system"
 defaultValue = "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266"
 
 [setting.oracle_manager_source]
-defaultValue = "oracle-manager:1.2.6"
+defaultValue = "oracle-manager:1.2.7"
```
dbeal-eth, line 15: @noahlitvin FYI somehow this version wasn't properly bumped

protocol/synthetix/contracts/mocks/CollateralMockWithoutDecimals.sol
```diff
@@ -0,0 +1,180 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.11 <0.9.0;
+
+contract CollateralMockWithoutDecimals {
```
dbeal-eth, line 4: probably could have just modified the other mock tocken to optionally revert on the `decimals` call

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
```diff
@@ -212,20 +217,39 @@ library CollateralConfiguration {
...
+        /// @dev this try-catch block assumes there is no malicious code in the token's fallback function
+        try IERC20(self.tokenAddress).decimals() returns (uint8 decimals) {
+            if (decimals == 18) {
+                amountD18 = tokenAmount;
+            } else if (decimals < 18) {
+                amountD18 = (tokenAmount * DecimalMath.UNIT) / (10 ** decimals);
+            } else {
+                // ensure no precision is lost when converting to 18 decimals
+                if (tokenAmount % (10 ** (decimals - 18)) != 0) {
+                    revert PrecisionLost();
```
dbeal-eth, line 244: this is a very minor thing but its usually a good idea to include as much data as possible in revert reasons. so for `PrecisionLost`, including the `tokenAmount` and the number of decimals reported by the token would be potentially useful in case a confused developer is debugging this code and not understanding why it is erroring exactly

protocol/synthetix/contracts/storage/CollateralConfiguration.sol
```diff
@@ -212,20 +217,39 @@ library CollateralConfiguration {
...
+                } else {
```
dbeal-eth, line 245: unnecessary `else`

protocol/synthetix/test/integration/storage/CollateralConfiguration.test.ts
```diff
@@ -0,0 +1,167 @@
...
+          minDelegationD18: '20000000000000000000',
+          depositingEnabled: true,
+        })
+    ).wait();
+  };
+
+  describe('convertTokenToSystemAmount()', async () => {
+    it('scales token with 0 decimals to system amount', async () => {
+      const DECIMALS = 0;
+      await initFakeCollateralConfig(DECIMALS);
```
dbeal-eth, line 41: for all the tests that use `initFakeCollateralConfig` would be best to enclose in a `describe` block that has a `before`

I know this is probably normal for foundry but in our hardhat project tests are highly heiarchal.

dbeal-eth (approved):
just smaller notes from me now

protocol/synthetix/contracts/mocks/CollateralMockWithoutDecimals.sol
dbeal-eth: rather than creating a whole new mock

PR #1408, opened 2023-02-02
Pyth Url; reorg NFT

sunnyvempati:
- fixing pyth url based on feedback
- removed custom mint/burn


markets/spot-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -74,15 +74,20 @@ contract AsyncOrderModule is IAsyncOrderModule {
...
+            .settlementDelay;
 
         asyncOrderClaim = AsyncOrderClaim.create(
             marketId,
             asyncOrderId,
             orderType,
             amountEscrowed,
             settlementStrategyId,
-            block.timestamp,
+            block.timestamp + settlementDelay,
```
sunnyvempati, line 90: storing settlement time instead of commitment time

markets/spot-market/contracts/modules/token/AsyncOrderClaimTokenModule.sol
```diff
@@ -2,33 +2,11 @@
...
-     */
-    function mint(address owner) external returns (uint256 id) {
-        OwnableStorage.onlyOwner();
+contract AsyncOrderClaimTokenModule is NftModule {
 
-        id = totalSupply();
-        _mint(owner, id);
-
-        emit Mint(owner, id);
-    }
```
sunnyvempati, line 22: this was throwing a compiler error; we can't override `mint` since `NFTModule` already is.  we would have to inherit directly from `erc721` then re-add all the initialization code.

instead, i opted to remove the special mint/burn functions here and just use the `totalSupply` as the id, the one place we mint in `AsyncOrderModule`.  burn didn't change so it was redundant.

noahlitvin (approved):

PR #1409, opened 2023-02-03
Async fixed fee

sunnyvempati:
after chatting with afif, looks like we need this back.  generally, this will be set to lower than the atomic fixed fee so users are incentivized to use async orders more.

### DIFF
--- a/markets/spot-market/contracts/interfaces/IFeeConfigurationModule.sol
+++ b/markets/spot-market/contracts/interfaces/IFeeConfigurationModule.sol
@@ -34,7 +34,8 @@ interface IFeeConfigurationModule {
     event AsyncFixedFeeSet(uint indexed synthMarketId, uint asyncFixedFee);
 
     /**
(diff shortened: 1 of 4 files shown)


sunnyvempati:
tests are coming in a separate branch.

markets/spot-market/contracts/interfaces/IFeeConfigurationModule.sol
```diff
@@ -26,13 +26,20 @@ interface IFeeConfigurationModule {
...
 
+    /**
+     * @notice emitted when the fixed fee for async orders is set.
+     * @param synthMarketId Id of the market to set the fees for.
+     * @param asyncFixedFee the fixed fee for the corresponding market
+     */
+    event AsyncFixedFeeSet(uint indexed synthMarketId, uint asyncFixedFee);
+
     /**
      * @notice emitted when the fixed fee for atomic orders is set for a given transactor
```
noahlitvin, line 37: Add async order here

noahlitvin (approved):

PR #1412, opened 2023-02-03
Add councils cannon deployments

mjlescano:


protocol/governance/cannonfile.ambassador-council.toml
```diff
@@ -0,0 +1,65 @@
+name = "synthetix-ambassador-council"
+description = "On-chain voting for synthetix Ambassador Council"
+version = "<%= package.version %>"
+
+[setting.salt]
```
dbeal-eth, line 5: why aren't you `include`ing from the other `cannonfile.toml`? Doesn't that seem like it would result in signficantly less code duplication?

dbeal-eth (changes requested):

PR #1421, opened 2023-02-08
fix `mintUSD` with stronger validations

dbeal-eth:
The purpose of disabling a collateral is to begin the offloading/phase-out of a collateral token on the system. Therefore, minting of additional collateral for disabled collateral types should not be allowed.

reported in OZ audit


sunnyvempati (approved):
🚢

PR #1423, opened 2023-02-08
remove owner module initialize functions

dbeal-eth:
now that we have the `UUPSProxyWithOwner`, there is no need to have any initialization in the `OwnerModule`. Therefore, the module has been stripped and now is just literally module that inherits directly from Ownable with nothing else.

this will leave uninitialized contracts which utilize owner functions or anything sensitive with `owner = 0x0000000...`, preventing any selfdestruct() potential.

this issue is a resolution for an issue raised by OZ audit.

**NOTE:** this change affects `Proxy` code and will therefore trigger a redeploy cc @noahlitvin


ajsantander (approved):
LGTM

PR #1427, opened 2023-02-09
fixes for collateral locks

dbeal-eth:
fixes a number of issues identified during the audit related to collateral locks:

* `cleanExpiredLocks` was not correctly limiting its iteration
* `createLock` was not validating inputs enough (ex. creating lock in the past)
* lack of emitted events

found as part of the OZ audit


sunnyvempati (approved):
🚢

PR #1431, opened 2023-02-09
pools not removed after delegate collateral exit

dbeal-eth:
previous code did not remove pools from the set when the account exited the pool position.

found as part of OZ audit


protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -276,4 +278,25 @@ contract VaultModule is IVaultModule {
...
+        Collateral.Data storage depositedCollateral = Account.load(accountId).collaterals[
+            collateralType
+        ];
+
+        bool containsPool = depositedCollateral.pools.contains(poolId);
+        if (added && !containsPool) {
+            depositedCollateral.pools.add(poolId);
+        } else if (!added && containsPool) {
+            depositedCollateral.pools.remove(poolId);
+        }
```
sunnyvempati, line 300: is this tested specifically, or implicitly somewhere?

sunnyvempati (approved):
add test if it doesn't exist otherwise 🚢

PR #1440, opened 2023-02-12
fix credit capacity handling with market deposited collateral

dbeal-eth:
when there is market deposited collateral, it should be possible for creditCapacity to go negative in order to protect against overruns when more collateral is withdrawn from the market than capacity provided by the pools.

when there is no deposited collateral, then creditcapacity functions same as before.

found as part of oz audit


sunnyvempati (approved):
makes sense; nice

PR #1445, opened 2023-02-13
N-15: gas optimizations

sunnyvempati:
- [x] When depositing market collateral, authorized access is checked after converting the
tokenAmount into systemAmount (18 decimals). This performs an external call to
check the token decimals which can be avoided if authorized access is checked first.
- [x] Line 159 in the LiquidationModule is redundant, since rawVaultDebt can never
be negative at this point because isLiquidatable would have returned false and reverted
earlier. It can be safely casted to uint without checking whether it is a non-negative
value.
- [x] In line 77 of RewardDistribution , the contract is first defining the variable
totalSharesD18 but then is not using it in line 89.
- [x] When calculating the total deposited collateral value in USD for a Market , it loops
through all DepositedCollateral entries and queries the current USD price through
the Oracle. Consider first checking whether the amount of collateral is larger than zero, in
order to avoid querying the Oracle in that case.
- [x] The accumulateActor function is called as a ticker from several parts of the system,
with the goal of updating a specific actor's last value per share and recovering the value
change since the last interaction. In order to update this value in storage, the actor's number of shares is read from storage and re-updated to the same value in order to
leverage the setActorShares logic which includes updating this inner value
( lastValuePerShareD27 ). Consider isolating this piece of logic in a separate internal
function so that it can be reused without the need for two unnecessary SSTORE
opcodes (updating actor's shares and total shares within the Distribution since
those values have not changed).
- [x] In the ScalableMapping storage at the get function, the self.totalSharesD18
is first read, but then it is read again internally in the totalAmount function called
inside.
- [x] In the createLock function of the CollateralModule contract, the account is first
loaded internally in the onlyWithPermission function, and then again in the next line.
In the RewardDistribution storage, the block.timestamp is first casted to
int256 , then to int128 to be stored in the curTime variable, and then it is
converted back again to uint256 in the if statement.
- [x] When calling distribute within RewardDistribution , the first step is to read total
shares from storage and keep the value on the local totalSharesD18 variable.
However, it is then read again from storage instead of using the cached value.
- [x] When migrating a position from v2x within LegacyMarket , consider creating the
account on V3 before attempting to gather all assets from v2x, including vesting rewards,
since the selected accountId might already be used, saving the user some gas by
reverting early if this is the case.

### DIFF
--- a/protocol/synthetix/contracts/modules/core/CollateralModule.sol
+++ b/protocol/synthetix/contracts/modules/core/CollateralModule.sol
@@ -49,11 +49,6 @@ contract CollateralModule is ICollateralModule {
 
         address self = address(this);
 
-        uint256 allowance = IERC20(collateralType).allowance(depositFrom, self);
-        if (allowance < tokenAmount) {
-            revert IERC20.InsufficientAllowance(tokenAmount, allowance);
-        }
-
         collateralType.safeTransferFrom(depositFrom, self, tokenAmount);
 
         account.collaterals[collateralType].increaseAvailableCollateral(
--- a/protocol/synthetix/contracts/modules/core/CollateralModule.sol
+++ b/protocol/synthetix/contracts/modules/core/CollateralModule.sol
@@ -49,6 +49,11 @@ contract CollateralModule is ICollateralModule {
 
         address self = address(this);
 
+        uint256 allowance = IERC20(collateralType).allowance(depositFrom, self);
+        if (allowance < tokenAmount) {
+            revert IERC20.InsufficientAllowance(tokenAmount, allowance);
+        }
+
         collateralType.safeTransferFrom(depositFrom, self, tokenAmount);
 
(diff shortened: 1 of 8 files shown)


protocol/synthetix/contracts/modules/core/CollateralModule.sol
```diff
@@ -49,11 +49,6 @@ contract CollateralModule is ICollateralModule {
 
         address self = address(this);
 
-        uint256 allowance = IERC20(collateralType).allowance(depositFrom, self);
-        if (allowance < tokenAmount) {
-            revert IERC20.InsufficientAllowance(tokenAmount, allowance);
-        }
```
sunnyvempati, line 55: this is already checked in erc20.sol

dbeal-eth (approved):
why does it look like the feature flag security changes are being re-implemented?

PR #1449, opened 2023-02-14
oracle manager audit fix

Rickk137:
[H-1] PriceDeviationCircuitBreakerNode does not handle negative prices correctly
[H-2] ReducerNode’s max() operation incorrect on only negative prices
[M-1] UniswapNode does not validate that input tokens belong to the pool
[M-2] PriceDeviationCircuitBreaker does not scale deviation parameter to 18 decimal places
[M-3] IExternalNode is missing logic to validate parameters
[L-1] ExternalNodes lack the ability to have additional parameters
[L-2] UniswapNode cannot handle ERC20s with larger than 18 decimal places
[L-3] PriceDeviationCircuitBreakerNode does not allow for prices to be zero
[Q-1] UniswapNode’s token0 and token1 naming is not user friendly
[Q-2] ReducerNode.sol doesn’t mention that Operation.MEDIAN returns the average when there are an equal amount of Parents
[Q-3] ReducerNode’s div() function will revert if a price is zero
[Q-4] PriceDeviationCircuitBreakerNode's naming of fallbackPrice is misleading


protocol/oracle-manager/contracts/nodes/PriceDeviationCircuitBreakerNode.sol
```diff
@@ -18,17 +18,16 @@ library PriceDeviationCircuitBreakerNode {
...
-            if (deviationTolerance.toInt() < ((difference * 100) / primaryPrice)) {
-                if (parentNodeOutputs.length > 2 && parentNodeOutputs[2].price != 0) {
+        int256 primaryPrice = abs(parentNodeOutputs[0].price);
+        int256 comparisonPrice = abs(parentNodeOutputs[1].price);
+
+        if (primaryPrice != comparisonPrice) {
+            int256 difference = abs(primaryPrice - comparisonPrice);
+            if (
+                primaryPrice == 0 ||
+                deviationTolerance.toInt() < ((difference * 1e18) / primaryPrice)
```
dbeal-eth, line 28: `difference.upscale` would be preferred over `difference * 1e18`

protocol/oracle-manager/contracts/nodes/ReducerNode.sol
```diff
@@ -92,7 +92,10 @@ library ReducerNode {
     function max(
         NodeOutput.Data[] memory parentNodeOutputs
     ) internal pure returns (NodeOutput.Data memory maxPrice) {
-        for (uint256 i = 0; i < parentNodeOutputs.length; i++) {
+        if (parentNodeOutputs.length > 0) {
```
dbeal-eth, line 95: would be better to just verify that the number of parent nodes for the reducer node is >0 in the validate function (I believe that is possible right?)

protocol/oracle-manager/contracts/nodes/ReducerNode.sol
```diff
@@ -128,6 +131,9 @@ library ReducerNode {
         divPrice.price = parentNodeOutputs[0].price;
         divPrice.timestamp = parentNodeOutputs[0].timestamp;
         for (uint256 i = 1; i < parentNodeOutputs.length; i++) {
+            if (parentNodeOutputs[i].price == 0) {
+                revert("parent price is zero");
```
dbeal-eth, line 135: please create a parameterized revert rather than a string revert here

protocol/oracle-manager/contracts/nodes/UniswapNode.sol
```diff
@@ -104,11 +104,25 @@ library UniswapNode {
...
+
+        if (
+            !(poolToken0 == token && poolToken1 == stablecoin) &&
+            !(poolToken0 == stablecoin && poolToken1 == token)
+        ) {
             return false;
         }
 
-        if (IERC20(token1).decimals() != decimals1) {
+        if (decimals0 > 18 || decimals1 > 18) {
```
dbeal-eth, line 125: this could cause problems in the future for supporting tokens that we want to support. we have a function in v3 which effectively handles the issue of what happens if a token has > 18 decimals. are we sure this is the path we want to take (at least for right now?)

dbeal-eth (approved):
tests may still be a little weak

PR #1451, opened 2023-02-16
(M-4) Add storage slot validation on all contracts

mjlescano:
This PR fixes the following:
1. Adds all the contracts from `contracts/` folders to the storage check, not only the ones in `contracts/modules/**`
2. Adds a github action to check that the developer has updated the storage dump file (meaning that it was checked)
3. Updates the old ci package-lock check to not use an unnecessary github action dependency

### DIFF
--- a/.github/workflows/check-packages.yml
+++ b/.github/workflows/check-packages.yml
@@ -23,10 +23,5 @@ jobs:
       # forgot to update it.
       - name: Check package-lock.json was correctly submmitted by the user
         run: |
-          cp package-lock.json package-lock.original.json
           npm i --package-lock-only
-      - uses: GuillaumeFalourd/diff-action@v1
-        with:
-          first_file_path: package-lock.json
-          second_file_path: package-lock.original.json
-          expected_result: PASSED
+          git diff --exit-code -- package-lock.json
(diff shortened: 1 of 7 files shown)


.github/workflows/check-packages.yml
```diff
@@ -23,10 +23,5 @@ jobs:
       # forgot to update it.
       - name: Check package-lock.json was correctly submmitted by the user
         run: |
-          cp package-lock.json package-lock.original.json
           npm i --package-lock-only
-      - uses: GuillaumeFalourd/diff-action@v1
```
dbeal-eth, line 28: why this change?

.github/workflows/test.yml
```diff
@@ -60,9 +60,11 @@ jobs:
           run_daemon: true
       - run: npm ci
       - run: npm run build
-      - if: ${{ matrix.size-contracts }}
+      - if: matrix.size-contracts
         name: Execute size command
         run: npm run --workspace=${{ matrix.workspace }} size-contracts
+      - name: Check storage.dump.sol is up-to-date
+        run: npm explore ${{ matrix.workspace }} -- git diff --exit-code -- storage.dump.sol
```
dbeal-eth, line 67: :pray:

dbeal-eth (approved):

PR #1463, opened 2023-02-17
Address OZ out of scope issues

ajsantander:


dbeal-eth (approved):
still need to impl @Rickk137 suggested change

noahlitvin (approved):

PR #1496, opened 2023-03-10
feat: add quote buy function to spot market

Rickk137:


dbeal-eth (changes requested):
i may be missing something but I think we also need `quoteBuyIn` and `quoteBuyOut`. Otherwise, you would have no way to accurately quote if you are wanting the purchase cost for an exact amount out with the current implementation.

also `quoteBuy` and `quoteSell` need docstrings confirming these things if only having the quote for In is intended.

PR #1497, opened 2023-03-12
Add openPosition

bachstatter:


markets/perps-market/contracts/modules/AccountModule.sol
```diff
@@ -14,4 +16,17 @@ contract AccountModule is IAccountModule {
     function totalAccountOpenInterest(uint128 accountId) external view override returns (int) {
         return PerpsAccount.load(accountId).getTotalAccountOpenInterest(accountId);
     }
+
+    function openPosition(
```
sunnyvempati, line 20: should we rename this to `getAccountPosition` or `getOpenPosition`?  `openPosition` sounds like we're opening a position.

sunnyvempati (commented):
one quick note.  but lgtm.

i have some changes in the testing branch which changes what `calculateExpectedPosition` does, but pnl and accruedFunding are the 2nd and 3rd params.  `pnl` also includes the `accruedFunding`, but yeah prolly no changes needed here really but just fyi.

markets/perps-market/contracts/modules/AccountModule.sol
bachstatter: yea I wasn't sure what we want for views.. I went with no `get` prefix but we could add it for all view functions

sunnyvempati (approved):

PR #1498, opened 2023-03-13
Yarn

noisekit:
- Deps checker tool
- Add yarn
- npm -> yarn in all scripts
- Update all internal dependencies to have explicit "workspace:*" version
- Add missing dependencies and remove unused
- Fix mismatched versions
- Remove package checks as yarn does it on any run
- Skip building perps market as it is WIP
- Cleanup eslint config and disable prettier-handled rules
- Fix lint errors
- Only build what each project depends on
- Clean cache and rebuild to get storage regenerated in `spot-market`
- Fix invalid bash script syntax (comment) and few unimplemented scripts
- Update outdated snapshot in `hardhat-storage`

NOTE: Filter out `zip` files from review if not using `.` editor 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "deps:circular": "deps-circular"
   },
   "devDependencies": {
-    "@snx-private/deps": "workspace:*",
+    "@synthetixio/deps": "workspace:*",
     "@types/mocha": "10.0.1",
     "@types/node": "18.11.10",
     "@typescript-eslint/eslint-plugin": "5.45.0",
--- a/utils/deps/package.json
+++ b/utils/deps/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@snx-private/deps",
+  "name": "@synthetixio/deps",
   "private": true,
   "version": "1.0.0",
   "author": "Nikita <noisekit@btnk.au>",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1907,20 +1907,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@snx-private/deps@workspace:*, @snx-private/deps@workspace:utils/deps":
-  version: 0.0.0-use.local
-  resolution: "@snx-private/deps@workspace:utils/deps"
-  dependencies:
-    depcheck: 1.4.3
-    prettier: 2.8.0
-    typescript: 4.9.3
-  bin:
-    deps: ./deps.js
-    deps-circular: ./circular.js
-    deps-mismatched: ./mismatched.js
-  languageName: unknown
-  linkType: soft
-
 "@solidity-parser/parser@npm:^0.14.0, @solidity-parser/parser@npm:^0.14.1, @solidity-parser/parser@npm:^0.14.5":
   version: 0.14.5
   resolution: "@solidity-parser/parser@npm:0.14.5"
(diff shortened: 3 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -57,19 +54,39 @@ jobs:
...
+
+      - run: yarn workspaces foreach --topological-dev --recursive --verbose --from ${{ matrix.workspace }} run build
+
+      - name: Execute size-contracts command
+        run: |
+          if yarn workspace ${{ matrix.workspace }} run | grep size-contracts; then
+            yarn workspace ${{ matrix.workspace }} run size-contracts
+          else
+            echo 'SKIP. No "size-contracts" script'
+          fi
```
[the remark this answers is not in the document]
dbeal-eth: in the future `size-contracts` should probably be moved to a separate CI job and run `yarn lerna run size-contracts` instead

.yarnrc.yml
```diff
@@ -0,0 +1,27 @@
+checksumBehavior: ignore
+
+defaultSemverRangePrefix: ""
+
+enableTelemetry: false
```
dbeal-eth, line 5: i personally think its always good to enable telemetry since it helps the devs who build the tool but just my opinion :)

dbeal-eth: if the devs were evil and were collecting private data from us they wouldn't have done it through the opt-outable telemetry option

.yarnrc.yml
```diff
@@ -0,0 +1,27 @@
...
+
+supportedArchitectures:
+  os:
+    - "darwin"
+    - "linux"
+  cpu:
+    - "x64"
+    - "arm64"
+  libc:
+    - "glibc"
```
dbeal-eth, line 27: lol we don't support musl? :laughing:

markets/perps-market/package.json
```diff
@@ -4,28 +4,36 @@
...
-    "@synthetixio/router": "*",
-    "node-fetch": "2.6.9",
-    "solidity-docgen": "0.6.0-beta.22",
+    "@ethersproject/abi": "5.7.0",
+    "@ethersproject/providers": "5.7.2",
+    "@nomiclabs/hardhat-ethers": "2.2.1",
+    "@synthetixio/common-config": "workspace:*",
+    "@synthetixio/core-contracts": "workspace:*",
+    "@synthetixio/core-modules": "workspace:*",
+    "@synthetixio/main": "workspace:*",
```
dbeal-eth, line 27: so this probably isnt relevant to this specific PR, but the plan here was that cannon would read the version of `synthetix:<version>` to depend on from this variable. In order to do this, the version would be `@synthetixio/main: 3.0.1` for example or whatever the version is, with no `workspace:` or any other addition. Then, cannonfile imports the correct version with `source = "synthetix:<%= package.devDependencies['@synthetixio/main'] %>"` (as an example). Finally, lerna automatically bumps the version any time the dependency is updated, and cannon remains happy. guess what I am saying is, LMK if you have an alternative plan at this point.

dbeal-eth: we have a lot of these types of dependencies within the project. They are basically dependencies only insofar as that we are using the solidity sources and cannon package matching, so it makes sense to consider that these packages may have their version resolved differently.

dbeal-eth: oh one more thing, wouldn't `workspace:*` resolve to `*` for the versions when the pacakge is actually published? Is this intended behavior for our internal javascript packages? its not a big deal for us, but we might as well try to do things reasonably when possible.

package.json
```diff
@@ -14,20 +14,24 @@
   "license": "MIT",
   "homepage": "https://github.com/Synthetixio/synthetix-v3#readme",
   "scripts": {
-    "clean": "npm run --workspaces --if-present clean",
-    "build": "npm run -w @synthetixio/core-utils -w @synthetixio/hardhat-storage -w @synthetixio/router -w @synthetixio/sample-project -w @synthetixio/oracle-manager -w @synthetixio/main build",
-    "test": "npm run --workspaces --if-present test",
-    "coverage": "npm run --workspaces --if-present coverage",
+    "clean": "yarn workspaces foreach --parallel --verbose run clean",
+    "build": "yarn workspaces foreach --topological-dev --verbose run build",
```
dbeal-eth, line 18: these commands will probably become their lerna equivalents

dbeal-eth (approved):
just notes from me, if you already are aware/have everything accounted we can probably immediately merge this

PR #1503, opened 2023-03-16
fix: wrapper module decimal precision

Rickk137:
https://gist.github.com/sipox11/8da4d03df0ff2abf2cae3c9d0b8ba0c6


sunnyvempati (commented):
some variable name changes but otherwise lgtm

PR #1504, opened 2023-03-17
Lerna ❤️ Cannon publish

noisekit:
- Reference exact `package.version` in cannon configs
- Make dependency packages public (when we publish contract lib, we need all of it's deps to be public too)
- Fix incorrect `solhint-plugin-numcast` version
- Add `publish-contracts` script for each publishable contract package (we'll need to add more later)
- Update README with new publishing process
- Upgrade cannon to `2.1.9`


dbeal-eth:
@noisekit thank you for this PR, the failure we are currently observing on CI pipeline is a lack of interpolation on cannon side which will be resolved shortly

dbeal-eth (approved):

dbeal-eth:
merging because CI is now passing after I made minimal pre-discussed changes, multiple reviews have had a chance to review, and we all want this change in time for next week would be nice.

PR #1519, opened 2023-03-27
implement SIP-320: market enforced min delegation time

dbeal-eth:
implements [SIP-320](https://sips.synthetix.io/sips/sip-320/)

notes:
* we decided to go with a setter function on the market manager rather than extending the `IMarket` interface because it would impact existing markets. Also, we had this built as an isolated function to reduce the chances that this would break markets in the future in case we were to ex. change the data structure in the future.


protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
```diff
@@ -164,4 +169,33 @@ contract MarketManagerModule is IMarketManagerModule {
...
+     * @inheritdoc IMarketManagerModule
+     */
+    function setMarketMinDelegateTime(uint128 marketId, uint32 minDelegateTime) external override {
+        Market.Data storage market = Market.load(marketId);
+
+        if (msg.sender != market.marketAddress) revert AccessError.Unauthorized(msg.sender);
+
+        // min delegate time should not be unreasonably long
+        // solhint-disable-next-line numcast/safe-cast
+        uint maxMinDelegateTime = uint(Config.read(_CONFIG_SET_MARKET_MIN_DELEGATE_MAX));
```
sunnyvempati, line 183: isn't there a `readUint` now that you can use?

sunnyvempati (approved):
couple non blocking comments. send it

noahlitvin (approved):

PR #1523, opened 2023-03-30
C-01: inflation attack fixes

sunnyvempati:
- introduces a `minimumUsdExchangeAmount` & `maxRoundingLoss` variables. (see comments for descriptions).
- adds tests against inflation attacks described in this gist: https://gist.github.com/sipox11/522bcd08115742990a85dca2909f18f2


markets/spot-market/contracts/storage/AsyncOrder.sol
```diff
@@ -45,6 +49,13 @@ library AsyncOrder {
 
         token.transferFrom(from, address(this), synthAmount);
         asyncOrderData.totalEscrowedSynthShares += sharesAmount;
+
+        // sanity check to ensure the right shares amount is calculated
+        uint acceptableSynthAmount = convertSharesToSynth(asyncOrderData, marketId, sharesAmount) +
+            maxRoundingLoss;
```
dbeal-eth, line 55: shouldn't we be subtracting maxRoundingLoss here?

markets/spot-market/contracts/storage/AsyncOrder.sol
```diff
@@ -2,12 +2,15 @@
...
 
 import "@synthetixio/core-modules/contracts/interfaces/ITokenModule.sol";
+import "@synthetixio/core-contracts/contracts/utils/DecimalMath.sol";
 import "../utils/SynthUtil.sol";
 
 /**
  * @title Async order top level data storage
  */
 library AsyncOrder {
+    error InvalidSharesCalculation(uint256 expected, uint256 actual);
```
dbeal-eth, line 12: this error name could be more descriptive to an innocent user. for example something like "InsufficientAmountDeposited" or something like that, could perhaps be more helpful

dbeal-eth (approved):
nice tests

minor notes

markets/spot-market/contracts/storage/AsyncOrder.sol
sunnyvempati: no, the `convertSharesToSynth` will never return _more_ synth than the synth amount sent in.  due to division, it loses some micro precision so adding the maxRoundingLoss to the conversion back

markets/spot-market/contracts/storage/AsyncOrder.sol
sunnyvempati: makes sense. changing

PR #1526, opened 2023-03-30
L-02/N-01: fix transactor

sunnyvempati:
- transactor is now the trader and not the keeper (msg.sender)
- N-01 also fixed in the event


markets/spot-market/contracts/modules/AsyncOrderSettlementModule.sol
```diff
@@ -282,7 +282,7 @@ contract AsyncOrderSettlementModule is IAsyncOrderSettlementModule {
         collectedFees = config.collectFees(
             marketId,
             fees,
-            msg.sender,
+            trader,
```
dbeal-eth, line 285: i just want to fully understand the reprecussions of this... does this mean that the fees will be collected from the person who created the trade, rather than the person who triggered a liquidation?

sunnyvempati: noo; we're just communicating to the fee collector who the trader was in case there's some special rules on how much to collect or w.e.  just a composability thing.  instead of sending the trader, prior to this change, it was sending the keeper address which was wrong

noahlitvin (approved):

PR #1528, opened 2023-04-02
L-10: Input validation

sunnyvempati:
- Check synth owner is not zero address

### DIFF
--- a/markets/spot-market/contracts/interfaces/ISpotMarketFactoryModule.sol
+++ b/markets/spot-market/contracts/interfaces/ISpotMarketFactoryModule.sol
@@ -16,6 +16,11 @@ interface ISpotMarketFactoryModule is IMarket {
     error NotNominated(address addr);
 
     /**
+     * @notice Thrown when createSynth is called with zero-address synth owner
+     */
+    error InvalidMarketOwner();
+
+    /**
      * @notice Gets fired when the synth is registered as a market.
      * @param synthMarketId Id of the synth market that was created
      */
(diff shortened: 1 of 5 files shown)


markets/spot-market/contracts/interfaces/ISpotMarketFactoryModule.sol
```diff
@@ -15,6 +15,11 @@ interface ISpotMarketFactoryModule is IMarket {
      */
     error NotNominated(address addr);
 
+    /**
+     * @notice Thrown when createSynth is called with zero-address synth owner
+     */
+    error InvalidMarketOwner();
```
dbeal-eth, line 21: you know we also have `ParameterError.InvalidParameter("owner", "cannot be 0")` for this as well, rigth?

markets/spot-market/contracts/storage/SettlementStrategy.sol
```diff
@@ -29,7 +29,7 @@ library SettlementStrategy {
          * @dev the address of the contract that will verify the result data blob.
          * @dev used for pyth and chainlink offchain strategies.
          */
-        address priceVerificationContract; // For Chainlink and Pyth settlement strategies
```
dbeal-eth, line 32: why remove this comment all of a sudden?

dbeal-eth (approved):

markets/spot-market/contracts/interfaces/ISpotMarketFactoryModule.sol
sunnyvempati: wow this is very useful.  i did not know that;  i'll make a note to come back and fix some of these.  could definitely reuse this

PR #1529, opened 2023-04-03
L-17: safe transfers

sunnyvempati:
- for wrappers and escrowing synths, use `safeTransfer` and `safeTransferFrom`.

for other tokens (such as USD token) and when transferring into the spot market system, using the normal `transfer` function to save gas.

### DIFF
--- a/markets/spot-market/contracts/storage/AsyncOrder.sol
+++ b/markets/spot-market/contracts/storage/AsyncOrder.sol
@@ -1,7 +1,6 @@
 //SPDX-License-Identifier: MIT
 pragma solidity >=0.8.11 <0.9.0;
 
-import "@synthetixio/core-contracts/contracts/token/ERC20Helper.sol";
 import "@synthetixio/core-modules/contracts/interfaces/ITokenModule.sol";
 import "../utils/SynthUtil.sol";
 
@@ -9,8 +8,6 @@ import "../utils/SynthUtil.sol";
  * @title Async order top level data storage
  */
 library AsyncOrder {
-    using ERC20Helper for address;
-
     struct Data {
         /**
          * @dev tracking total shares for share calculation of synths escrowed.  instead of storing direct synth amounts, we store shares in case of token decay.
@@ -47,7 +44,7 @@ library AsyncOrder {
                 token.balanceOf(address(this));
 
         asyncOrderData.totalEscrowedSynthShares += sharesAmount;
-        address(token).safeTransferFrom(from, address(this), synthAmount);
+        token.transferFrom(from, address(this), synthAmount);
     }
 
     function burnFromEscrow(uint128 marketId, uint256 sharesAmount) internal {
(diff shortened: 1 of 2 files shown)


markets/spot-market/contracts/storage/AsyncOrder.sol
```diff
@@ -43,8 +46,8 @@ library AsyncOrder {
             : (synthAmount * asyncOrderData.totalEscrowedSynthShares) /
                 token.balanceOf(address(this));
 
-        token.transferFrom(from, address(this), synthAmount);
         asyncOrderData.totalEscrowedSynthShares += sharesAmount;
+        address(token).safeTransferFrom(from, address(this), synthAmount);
```
dbeal-eth, line 50: why did you put the `safeTransferFrom` call after `asyncOrderData.totalEscrowedSynthShares`? Seems like this would open the door for potential reentrancy attack

sunnyvempati: wouldn't it be possible for an attack if i put it _after_?

we should make sure `totalEscrowedSynthShares` accounts for the new shares _before_ we call `safeTransferFrom` which could re-enter

dbeal-eth: in general when it comes to reentrancy, you always want to decrement tokens or balances before send a users "benefit" from the operation. In this case, `totalEscrowedsynthShares` is a value used to determine how much a user could hypothetically *withdraw* upon `burnFromEscrow`. They are able to call `burnFromEscrow` as long as `totalEscrowedSynthshares > synthAmount`.

now, lets just look at this function in isolation, assuming no safety guard rails anywhere else. Hypothetically, an attacker could call `transferIntoEscrow` and specify regular parameters, but have the `from` address be a contract. Upon calling `safeTransferFrom`, safeTranferFrom calls `from` and verifies that the transaction is valid or whatever. However, we ahve already incremented `totalescrowedSynthShares`, so the attacker can now call `burnFromEscrow` and withdraw the funds before depositing them.

If `totalescrowedSynthshares += sharesAmount` line was *after` the `safeTransferFrom` call, this simply wouldn't be possible, since you aren't giving the attacker the property they would be able to exploit in a call to `burnFromEscrow` (aka `totalEscrowedSynthShares` being higher.

hopefully you can see how its safer to first pull the tokens from the user before incrementing totalescrowedSynthShares

sunnyvempati: i see your point, but an attacker can't actually call `burnFromEscrow` in the same transaction.  the only time `burnFromEscrow` gets called is when the transaction is cancalled which cannot happen in the same transaction (due to the delay required).

if `totalescrowedSynthshares += sharesAmount` was after the `safeTransferFrom`, don't we have an issue with the following scenario via another commit sell:
- attacker commits sell, `totalEscrowedSynthShares` is incremented and we call `safeTransferFrom`.
- the attacker re-enters and commits another sell.  no synths transferred yet, but `totalEscrowedSynthShares` is higher.
- the second sell commit now the calculation of shares returns a larger amt of shares to the attacker because the `totalEscrowedSynthShares` to synth total balance ratio is off.

the good news is that, the sanity check that checks that the shares that the trader is receiving == the synth amt they specified would fail causing this attack not to be performed; but the threat still exists.

i'm also not opposed to just removing `safeTransferFrom` here.  since this is a synth token, we know the implementation will always revert and not return bool.  and we don't care if its a contract or an EOA or w.e as long as we get the funds.  

hope i'm making sense, let me know your thoughts

noahlitvin (approved):

PR #1550, opened 2023-04-05
N-24: Fix Transaction Type default

sunnyvempati:
- Add `NULL` to Transaction enum which sets the value 0 to "null".


noahlitvin:
@sunnyvempati Should there be a check somewhere to make sure you can't set it to zero when adding a settlement strategy?

noahlitvin (approved):

PR #1556, opened 2023-04-06
L-07 / L-09 / N-07 

Rickk137:


utils/core-modules/contracts/modules/NftModule.sol
```diff
@@ -32,7 +32,7 @@ contract NftModule is INftModule, ERC721Enumerable, InitializableMixin {
         string memory tokenName,
         string memory tokenSymbol,
         string memory uri
-    ) public {
+    ) external onlyIfNotInitialized {
```
dbeal-eth, line 35: why are we adding `onlyIfNotInitialized` here? Similar to TokenModule, should be reinitializable

dbeal-eth (approved):

PR #1561, opened 2023-04-06
c-6

noahlitvin:


markets/spot-market/contracts/modules/AtomicOrderModule.sol
```diff
@@ -271,6 +286,11 @@ contract AtomicOrderModule is IAtomicOrderModule {
             revert ExceedsMaxSynthAmount(maxSynthAmount, synthToBurn);
         }
 
+        (uint buySynths, ) = quoteBuyExactIn(marketId, usdAmount);
+        if (buySynths > synthToBurn) {
+            revert InvalidPrices();
```
dbeal-eth, line 291: would be better if `InvalidPrices()` returned parameters helping to indicate the invalidness of the prices. for example, the `buySynths > synthAmount` (or analagous for sell) seems to be the root of the problem, so it would make sense to include those as part of the parameters for the revert

dbeal-eth (approved):

PR #1566, opened 2023-04-12
Update router interface for new plugins

mjlescano:
This PR makes sure to update the cannon plugin interface being updated here: https://github.com/usecannon/cannon/pull/158


dbeal-eth (approved):
:pray:

utils/router/src/utils/cannon.ts
```diff
@@ -37,16 +39,22 @@ export type Config = JTDDataType<typeof config>;
...
+    ctx: ChainBuilderContextWithHelpers,
+    config: Config
+  ) {
+    if (!(runtime as unknown as { baseDir: string }).baseDir) {
       return null; // skip consistency check
       // todo: might want to do consistency check for config but not files, will see
     }
 
-    const newConfig = this.configInject(ctx, config);
+    const newConfig = this.configInject(ctx, config, null as unknown as PackageState);
```
dbeal-eth, line 57: btw, `packageSTate` is actually passed into the `getState` function now, shuoldn'th ave to use `null`

PR #1584, opened 2023-04-24
SIP-318-321 fix iosiro audit feedback

dbeal-eth:
lot of misc

### DIFF
--- a/protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
+++ b/protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
@@ -158,8 +158,13 @@ contract MarketManagerModule is IMarketManagerModule {
         // which doesn't affect `totalSupply`, thus simplifying accounting.
         IUSDTokenModule(address(usdToken)).burnWithAllowance(target, msg.sender, amount);
 
-        if (feeAmount > 0) {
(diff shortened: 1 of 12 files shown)


protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
```diff
@@ -237,7 +241,10 @@ contract MarketManagerModule is IMarketManagerModule {
         if (msg.sender != market.marketAddress) revert AccessError.Unauthorized(msg.sender);
 
         // min delegate time should not be unreasonably long
-        uint maxMinDelegateTime = Config.readUint(_CONFIG_SET_MARKET_MIN_DELEGATE_MAX, 86400 * 30);
+        uint256 maxMinDelegateTime = Config.readUint(
+            _CONFIG_SET_MARKET_MIN_DELEGATE_MAX,
+            86400 * 30
```
sunnyvempati, line 246: save this as a named constant instead?

sunnyvempati (approved):
small suggestion otherwise lg

protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
dbeal-eth: yea I need to figure out a better wya to do this in general, not just this isntance. tbd.

PR #1585, opened 2023-04-24
fix spot market warnings

sunnyvempati:


dbeal-eth (changes requested):
rather than just ignoring a bunch of next lines for unused param, can we ignore the whole file, like for example in the `FeeCollectorMock` and the `OracleVerifierMock` especially?

also does this fix *all* of the warnings or just a subset of them? could just be my imagination. either way woudl be nice to clarify.

noahlitvin (approved):
Thanks for doing this.

sunnyvempati:
> rather than just ignoring a bunch of next lines for unused param, can we ignore the whole file, like for example in the `FeeCollectorMock` and the `OracleVerifierMock` especially?
> 
> also does this fix _all_ of the warnings or just a subset of them? could just be my imagination. either way woudl be nice to clarify.

fixes all warnings in spot-market.
and i ignored the warnings for all contracts in "mocks" now.  thanks for feedback 👍🏽

PR #1589, opened 2023-04-26
cleanups for spot market cannonfile

dbeal-eth:
* as is standard now, all `invoke` steps should use camel case
* not necessary to have `coryProxyOwner` setting
* can now inject `formatBytes32String` directly into the command rather than just precalculating it and adding a comment


sunnyvempati (approved):
lgtm ty!

PR #1593, opened 2023-04-30
Fix owner sm ff

sunnyvempati:


protocol/oracle-manager/test/integration/bootstrap.ts
```diff
@@ -10,7 +10,7 @@ const abi = ethers.utils.defaultAbiCoder;
 interface Contracts {
   NodeModule: NodeModule;
 }
-
+console.log('ORACLE CORE BOOTSTRAP');
```
leomassazza, line 13: rogue console.log or intended?

protocol/synthetix/test/integration/bootstrap.ts
```diff
@@ -1,64 +1,52 @@
...
+export interface Systems {
   Account: AccountProxy;
   Core: CoreProxy;
   USD: USDProxy;
   CollateralMock: CollateralMock;
-  OracleManager: Oracle_managerProxy;
+  OracleManager: OracleManagerProxy;
 }
 
+console.log('SNX BOOTSTRAP');
```
leomassazza, line 28: another one

markets/perps-market/test/integration/bootstrap/bootstrapTraders.ts
```diff
@@ -0,0 +1,68 @@
...
+  needed for testing
+*/
+export function bootstrapTraders(data: Data) {
+  const { systems, signers, provider, accountIds } = data;
+  bootstrapStakers(systems, signers);
+
+  let trader1: ethers.Signer, trader2: ethers.Signer;
+
+  before('provide access to create account', async () => {
+    const [owner, , , trader1, trader2] = signers();
```
leomassazza, line 24: this pattern might hide trader1 and trader2 from definition 3 lines above. Won't be a problem in this `before` but if used in another place can hide a difficult to fix issue

sunnyvempati: well, it returns `trader1` and `trader2` as top level keys in this function so in a test, you would use those functions instead of using the 4th and 5th index in `signers`.  similar pattern to how we source `owner` and `marketOwner`

leomassazza (commented):
LGTM. As mentioned in DM, just some np comments added

leomassazza (approved):

PR #1598, opened 2023-05-05
Spot Market View Functions

noahlitvin:


sunnyvempati (commented):
could you just have one function that returns the entire `MarketConfiguration.Data`?

the FE would only need to call one function to get all configuration data then.  not sure whats best practice 🤷🏽

sunnyvempati (approved):

PR #1608, opened 2023-05-19
feat: add fill price in spot market events

Rickk137:


markets/spot-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -89,7 +89,6 @@ contract AsyncOrderModule is IAsyncOrderModule {
             referrer
         );
 
-        // Emit event
```
sunnyvempati, line 92: lol

sunnyvempati (approved):

PR #1609, opened 2023-05-20
Readme updates

sunnyvempati:


bachstatter:
Some notes:
No need to install cannon globally,`yarn cannon` works fine for me.
Mention that you configure cannon by running: `yarn cannon setup`

I don’t think yarn clean should be part of getting started, since I shouldn’t have any built file when setting up the project.

I my case I had some old generate files for test, (eg. `TestableAccount`). These caused yarn build to fail. I had to manually delete those. Looks like these files now get created in a generated folder, so probably not worth mention this issue

Suggestion change:

## Developer Get Started

1. Ensure you have [IPFS Desktop](https://docs.ipfs.tech/install/ipfs-desktop/) running and your local cannon settings configured.
2. Run `yarn` at the root directory: `/synthetix-v3`.
3. Run `yarn cannon setup`
- Run `cat ~/.local/share/cannon/settings.json` and ensure your `ipfsUrl` is configured to `http://localhost:5001`.
4. Run `yarn build` at the root directory: `/synthetix-v3`.
5. Run `yarn test` in the synthetix protocol directory: `/protocol/synthetix`.

- This will create `synthetix-testable` version and throw it into your local IPFS cluster.

6. Run `yarn test` in the spot market directory: `/markets/spot-market`.

- This will create a `spot-market-testable` version and throw it into your local IPFS cluster.

7. Run `yarn test` in the perps market directory: `/markets/perps-market`.

- The perps market is composed of both the synthetix core and spot market contracts and uses the testable versions to create a testable perps market environment.

If you have passing tests at this point, **Congrats**, you are all set for development.


## Trouble shooting
If you experience any unexpected errors try running:
- `yarn clean`

noahlitvin (approved):

PR #1581, opened 2023-04-21
Allow to set merkle root during voting period

mjlescano:
With this fix we are going to be able to execute the commands `setDebtShareSnapshotId` and `setCrossChainDebtShareMerkleRoot` during voting period.

PR #1611, opened 2023-05-22
Remove permission to set merkle tree during voting period

mjlescano:
This fix was added here, this PR removes it as it is not needed anymore: https://github.com/Synthetixio/synthetix-v3/pull/1581


fritzschoff (approved):
Screenshot 2023-05-25 at 16 34 43

PR #1614, opened 2023-05-22
Create global config params

sunnyvempati:
- we had a lot of global parameters in `storage/PerpsMarketFactory.sol`.  Moved these into global storages to improve where data exists.


markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -18,28 +18,22 @@ contract LiquidationModule is ILiquidationModule {
...
+            .liquidatableAccounts;
 
-        for (uint i = 0; i < factory.liquidatableAccounts.length(); i++) {
+        for (uint i = 0; i < liquidatableAccounts.length(); i++) {
             _liquidateAccount(liquidatableAccounts.valueAt(i).to128());
         }
     }
 
-    function _liquidateAccount(uint128 accountId) private {
+    function _liquidateAccount(uint128 accountId) internal {
```
leomassazza, line 30: Is there a way to just flag and not liquidate? 
If I understand the behavior correctly, it will check if the account can be flagged for liquidation, mark it (flag) and liquidate. or revert. So flag and liquidate is in fact a single step.

sunnyvempati: we will probably want to add another external function to flag account

markets/perps-market/contracts/storage/PerpsAccount.sol
kingofclubstroyDev, line 1: Line 188 should be `for (uint i = 0; i < runtime.profitableMarketsLength; i++) {`
and line 209 should be `for (uint i = 0; i < runtime.losingMarketsLength; i++) {` 
Currently they are looping through the length of all `openPositionMarketIds`, as both arrays are initialized as that length, which will leave at least one array (if openPositionMarketIds.length > 0) to execute the loop with an empty element (marketId of zero).

sunnyvempati: great catch; added a TODO For myself to clean this up

PR #1615, opened 2023-05-23
Added view func getPerpsMarketIds 

fritzschoff:


fritzschoff:
takled with Sunny and decided that we dont need that

PR #1616, opened 2023-05-23
Add assertion

fritzschoff:


markets/perps-market/test/integration/Account/ModifyCollateral.test.ts
```diff
@@ -1,5 +1,6 @@
 import { ethers } from 'ethers';
 import { bn, bootstrapMarkets } from '../bootstrap';
+import assertBignumber from '@synthetixio/core-utils/src/utils/assertions/assert-bignumber';
```
dbeal-eth, line 3: normally we call it `assertBn`

dbeal-eth (approved):

PR #1622, opened 2023-05-25
Implement burn from sender

barrasso:
This was requested by CCIP for testing purposes with their token pools.


dbeal-eth (changes requested):
* does not check access for owner or CCIP as sender
* needs tests
* does not pass lint checks

dbeal-eth (approved):

PR #1625, opened 2023-05-29
Test  perps market creation

leomassazza:


markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -64,6 +66,10 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
     ) external override returns (uint128) {
         FeatureFlag.ensureAccessToFeature(_CREATE_MARKET_FEATURE_FLAG);
 
+        if (marketOwner == address(0)) {
```
sunnyvempati, line 69: doesn't solidity automatically check that you're not sending address(0)?

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -59,6 +64,22 @@ library PerpsMarket {
...
+        if (a.owner == address(0)) {
+            revert InvalidMarket(marketId);
+        }
+
+        if (PerpsPrice.load(marketId).feedId == "") {
+            revert PriceFeedNotSet(marketId);
+        }
+
+        return a;
+    }
```
sunnyvempati, line 81: nicee

sunnyvempati: i'll remove my change on my branch

sunnyvempati (approved):

markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
leomassazza: no, if I remove that if, it will allow the function to execute.

PR #1627, opened 2023-05-30
add coverage to spot and perps markets

leomassazza:


sunnyvempati (commented):
we are actually moving away from coverage.  we are into issues with contract sizes only when running coverage; can't rmr the exact reason.

leomassazza:
deprecated

PR #1629, opened 2023-05-31
Liquidation rewards perps v3

sunnyvempati:
- added `liquidationRewardRatioD18`
- calculated liquidationReward based on the above ratio * notional value.  
- checked against global min/max liquidation reward values.

- applied to maintenance margin to determine liquidations and withdrawal of margin.
- refactored `liquidate` on account to take into account liquidation rewards and pay them out properly.

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -66,7 +66,7 @@ library PerpsAccount {
         if (self.openPositionMarketIds.length() == 0) {
             return (false, availableMargin, 0);
         }
-        // TODO: add in liquidation rewards
+
         requiredMaintenanceMargin = getAccountMaintenanceMargin(self, accountId);
         isEligible = requiredMaintenanceMargin > availableMargin;
     }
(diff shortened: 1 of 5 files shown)


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -114,120 +114,6 @@ library PerpsAccount {
...
-
-            runtime.amountToDeposit -= liquidationReward;
-            runtime.totalLiquidationRewards += liquidationReward;
-
-            factory.depositToMarketManager(positionMarketId, runtime.amountToDeposit);
-            self.collateralAmounts[SNX_USD_MARKET_ID] -= runtime.amountToDeposit;
-        }
-
-        factory.usdToken.transfer(msg.sender, runtime.totalLiquidationRewards);
-    }
```
sunnyvempati, line 229: just moved them to the bottom closer to the private functions

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -243,15 +129,15 @@ library PerpsAccount {
             uint256 requiredMaintenanceMargin
         ) = isEligibleForLiquidation(self, accountId);
 
-        availableWithdrawableCollateralUsd = availableMargin - requiredMaintenanceMargin;
-
         if (isEligible) {
             revert AccountLiquidatable(accountId);
         } else if (amountToWithdraw > availableWithdrawableCollateralUsd) {
```
leomassazza, line 134: we can remove the else here (and below) since we are reverting inside.

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -315,21 +201,28 @@ library PerpsAccount {
...
             );
             uint256 notionalValue = position.getNotionalValue(PerpsPrice.getCurrentPrice(marketId));
-            (, , , uint256 marketMaintenanceMargin) = marketConfig.calculateMarginRatios(
-                notionalValue
-            );
+            (, , , uint256 marketMaintenanceMargin, uint256 liquidationMargin) = marketConfig
+                .calculateRequiredMargins(notionalValue);
 
+            marketLiquidationRewards += liquidationMargin;
             totalMaintenanceMargin += marketMaintenanceMargin;
```
leomassazza, line 220: please, rename the intermediate value to prevent confusion with returned one.

leomassazza (approved):

PR #1630, opened 2023-06-01
More collateral tests

bachstatter:
I noticed we didn't have a successful withdraw test. And stumble on a bug. 

In general it would be nice to hear how we plan to tackle integration vs unit tests.  This bug was in `checkAvailableWithdrawableValue` which is an internal function. Do we plan write unit tests for this function or should the integration test cover the specifics of it?


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -134,14 +134,14 @@ library PerpsAccount {
             revert AccountLiquidatable(accountId);
         }
 
+        availableWithdrawableCollateralUsd = availableMargin - requiredMaintenanceMargin;
+
```
sunnyvempati, line 138: why is this a bug?  it can be after the revert.  makes no real difference since the middle if statement is a revert.  

essentially its:
- if any of these if statements are true, the call will revert.
- otherwise set the `availableWithdrawableCollateralUsd`.

markets/perps-market/test/integration/Account/ModifyCollateral.test.ts
```diff
@@ -187,6 +188,22 @@ describe('ModifyCollateral', () => {
...
+          .synth()
+          .connect(trader1())
+          .balanceOf(await trader1().getAddress());
+        const perpsBalanceAfter = await synthMarkets()[0]
+          .synth()
+          .connect(trader1())
+          .balanceOf(systems().PerpsMarket.address);
+        assertBn.equal(perpsBalanceAfter, wei(0.9).toBN());
+        assertBn.equal(spotBalanceAfter, wei(0.1).toBN());
+      });
```
sunnyvempati, line 205: i would move this into a separate test file since we will be testing a lot more withdraw calls.

and in general, move the action into a `before` and in the `it` blocks do the assertions.  and if you need to contain the logic, move all of it into its own describe, otherwise, it can potentially affect the subsequent `it` blocks.

for example, all of these `it` blocks are only adding assertions for the transaction that occured in the `before` block thats adding collateral.

sunnyvempati (changes requested):

markets/perps-market/contracts/storage/PerpsAccount.sol
bachstatter: Hmm are you sure, the new withdraw test is failing with out this change?

markets/perps-market/test/integration/Account/ModifyCollateral.test.ts
bachstatter: Yep that make sense. I've not split it out to failures, deposit and withdrawl

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        name: 'Ether',
+        token: 'snxETH',
+        buyPrice: bn(1_000),
+        sellPrice: bn(1_000),
+      },
+    ],
+    perpsMarkets: [
+      { name: 'Bitcoin', token: 'snxBTC', price: bn(10_000) },
+      { name: 'Ether', token: 'snxETH', price: bn(1_000) },
+    ],
```
sunnyvempati, line 33: can remove creating these markets

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
```diff
@@ -0,0 +1,126 @@
+import { ethers } from 'ethers';
+import { bn, bootstrapMarkets } from '../bootstrap';
+import assertBn from '@synthetixio/core-utils/src/utils/assertions/assert-bignumber';
+import assertEvent from '@synthetixio/core-utils/utils/assertions/assert-event';
+import { wei } from '@synthetixio/wei';
+
+describe('ModifyCollateral', () => {
```
sunnyvempati, line 7:
```suggestion
describe('ModifyCollateral withdraw', () => {
```

sunnyvempati (approved):
it jsut so happened that the perpsBtcMarketId was the same as the synth one, but we should fix that on the deposit tests.  i don't think we need to create perps markets at all for these modify collateral tests.

pre-approving, good stuff, thanks for organizing these

markets/perps-market/test/integration/Account/ModifyCollateral.failures.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    synthBTCMarketId = synthMarkets()[0].marketId(); // 3
+    synthETHMarketId = synthMarkets()[1].marketId(); // 4
+  });
+
+  describe('failure cases', async () => {
+    it('reverts when the account does not exist', async () => {
+      await assertRevert(
+        systems()
+          .PerpsMarket.connect(trader2())
+          .modifyCollateral(invalidAccountId, perpBTCMarketId, oneBTC),
```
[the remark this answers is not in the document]
bachstatter: Nice catch, this make sense.. I think I was a little quick on just copying

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
bachstatter: 👍

PR #1631, opened 2023-06-01
PV3 Offchain test helpers

leomassazza:
This PR introduces commit and settle order helpers. Also splits tests in minimal functionality for each file to prevent supersized and unmanageable tests as we had in PerpsV1/V2 (more than 7k lines).

- `OffchainAsyncOrder.commit.test.ts` will focus on commit's common test cases for an async order. It uses `settleHelper` 
- `OffchainAsyncOrder.settle.test.ts` will focus on settle's common test cases for an async order. It uses `commitHelper`


markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -85,4 +87,20 @@ contract PerpsAccountModule is IAccountModule {
         );
         return (pnl, accruedFunding, position.size);
     }
+
+    function submittedAsyncOrder(
+        uint128 accountId,
+        uint128 marketId
+    ) external view override returns (int) {
+        PerpsMarket.Data storage perpsMarket = PerpsMarket.loadValid(marketId);
```
sunnyvempati, line 95: optional check; could remove

markets/perps-market/test/integration/helpers/settleHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+    tx = await chainState
+      .systems()
+      .PerpsMarket.connect(data.keeper())
+      .settlePythOrder(pythPriceData, extraData, { value: updateFee });
+  });
+
+  return {
+    // startTime: () => startTime,
+    // initialCollateral: () => initialCollateral,
+    // totalCollateral: () => totalCollateral,
```
sunnyvempati, line 75: .

sunnyvempati (approved):
left some comments

markets/perps-market/contracts/modules/PerpsAccountModule.sol
leomassazza: need to load it either (at least until we change the location of asynchOrderClaims

PR #1633, opened 2023-06-02
Fix eslint running out of memory

0xjocke:
We loose `@typescript-eslint/no-floating-promises` but linting is now super fast and doesn't run out of memory


sunnyvempati (approved):

sunnyvempati:
a fair tradeoff

dbeal-eth (commented):
checking for floating promises is a critical lint imo as they can be extremely difficult to track down and are a common source of bugs. at least set it up so that floating promises can be checked with the addition of an environment variable?

dbeal-eth:
also I dont think the benefits of having "super fast linting" are super useful because we also have to wait for tests and etc which are always going to take longer during CI

btw I would like to point out on my computer(s) I have never encountered this out of memory issue. not sure why.

bachstatter:
I do agree that it's a good rule when it works.. 
I was running out of memory consistently, I think this has happened for Sunny and Leo too.
With the floating promises rule it took 30s to lint this file:
`protocol/governance/hardhat.config.ts` 
And after that it would always run out of memory.

Apparently it works when doing `yarn clean`. I'm happy to keep the rule if you rely on it, but I would need to disable it locally  since it consistently hangs in vscode

bachstatter:
Trying to find other workaround...https://github.com/Synthetixio/synthetix-v3/pull/1638

PR #1638, opened 2023-06-05
Use tsconfig.eslint.json for ts files

bachstatter:
We configure eslint to use:
```
  parserOptions: {
    ecmaVersion: 12,
    tsconfigRootDir: __dirname,
    project: ['./tsconfig.eslint.json'],
  },
```
But then we override it to 
```
   parserOptions: {
        project: ['./tsconfig.json'],
      },
```

Removing this override helped with my running out of memory issue..  I think previosly it didn't lint properly since it now finds two instances of lint errors locally..

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,9 +33,7 @@ module.exports = {
     {
       files: ['**/*.ts'],
       extends: ['plugin:@typescript-eslint/recommended'],
-      parserOptions: {
-        project: ['./tsconfig.json'],
-      },
+
       rules: {
         '@typescript-eslint/no-floating-promises': 'error',
         '@typescript-eslint/no-var-requires': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,11 +12,8 @@ module.exports = {
     Promise: 'readonly',
   },
   plugins: ['progress', 'no-only-tests'],
-  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 12,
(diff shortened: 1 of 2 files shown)


bachstatter:
Hmm it still doesn't find the errors in CI..
`GlobalPerpsMarket.test.ts` have two floating promises

bachstatter:
image
image

Errors shows up locally, both in editior and when running `yarn lint`.. But on CI `yarn lint` does not pick up the errors :(

bachstatter:
Sent this is discord too, but adding here for the protocol... 

GlobalPerpsMarket.test.ts have two @typescript-eslint/no-floating-promises issues.
These show up locally but not in CI.

The reason for this is that @synthetixio/core-utils/utils/assertions/assert-revert would be used from src locally, but in the GitHub action docker container it would use the published version. And the published version seems to be missing types.

I would expect that issue to be detected by running typescript. That’s when I realised we do not run and typeschecks  (tsc ) in CI. 
Currently if I run yarn tsc —noEmit we have 303 issues.  To have @typescript-eslint/no-floating-promises plugin working I think we need to resolve our type issues..

So I think we have a few options:
Make sure we run typechecks in CI and fix our issues. Note that a “fix” can still be explicitly adding any or @ts-ignore.
Or
Realise that typescript is causing more problem that is solves and use JS for tests
Or 
I guess we keep things as is, but then we can’t rely on any lint rules requiring typescript

bachstatter:
#1644

PR #1642, opened 2023-06-06
Collateral Helper and More commit test

leomassazza:


markets/perps-market/test/integration/helpers/collateralHelper.ts
```diff
@@ -0,0 +1,121 @@
+import { ethers } from 'ethers';
+import { Systems } from '../bootstrap';
+
+type IncomingChainState = {
+  systems: () => Systems;
+  provider: () => ethers.providers.JsonRpcProvider;
+};
+
+type TradeSynthData = {
+  synthName: () => string;
```
sunnyvempati, line 10: dont think we need this. also we should just pass the entire synth that gets returned from `bootstrapSynthMarkets` so that way you have the id and the synth itself.  instead of destructuring

markets/perps-market/test/integration/helpers/collateralHelper.ts
```diff
@@ -0,0 +1,121 @@
...
+      perpsFinalBalance = await trade
+        .synthMarket()
+        .balanceOf(chainState.systems().PerpsMarket.address);
+    }
+
+    stats.push({
+      spotInitialBalance: () => spotInitialBalance,
+      perpsInitialBalance: () => perpsInitialBalance,
+      spotFinalBalance: () => spotFinalBalance,
+      perpsFinalBalance: () => perpsFinalBalance,
```
sunnyvempati, line 114: these are helpful 👍🏽

markets/perps-market/test/integration/helpers/commitHelper.ts
```diff
@@ -42,6 +42,7 @@ export const commitOrder: CommitOrderType = (data, chainState) => {
       acceptablePrice: data.acceptablePrice(),
       trackingCode: data.trackingCode,
     });
+    await tx.wait(); // force immediate confirmation to prevent flaky tests due to block timestamp
```
sunnyvempati, line 45: this is good to know, saw some flakiness in spot market, could be related to this in case they pop up next time

sunnyvempati (approved):
one optimization.  can be in a follow up

PR #1644, opened 2023-06-07
Type issues

bachstatter:
This fixes 200+ type issues..
Most of them are fixed properly. To avoid sinking to much time into this,  some issues where I struggled to find quick fix is ignored with @ts-ignore


dbeal-eth:
how do I reproduce these type issues on my machine? I don't understand the need for most of these changes I guess.

bachstatter:
> how do I reproduce these type issues on my machine? I don't understand the need for most of these changes I guess.

you can run `yarn tsc --noEmit`.. I wouldn't care too much but we've been running into some issues that are hard to explain, and I think all these type errors are the reason for it. 

The main thing that led me down this rabbithole is that some lint rules relies on typescript, and it seems like there's some memory leaks when the project doesn't compile properly. 

also, both me an sunny got this very weird compilation error that only goes a way with a clean:
image

In general all "weird" errors get hard to fix when you have 200+ type errors. But after this PR everything compiles fine

bachstatter:
> how do I reproduce these type issues on my machine? I don't understand the need for most of these changes I guess.

It is only tests so I dont think we should force anyone to write perfect types if the tests pass. All I'm asking is that issues gets marked with `//@ts-ignore` so we can compile the project without issues

bachstatter:
I did also asked for feedback, describing the issues and suggested three different options to move forward. I got no replies so went with option 1
https://discord.com/channels/413890591840272394/880428609800650772/1115131157172977744

dbeal-eth (approved):
im sure this change is super helpful but please fix ci fails before merge :+1:

PR #1645, opened 2023-06-07
Reorg and add open position

sunnyvempati:
- cleaned up interfaces for commit/settle
- auto create settlement strategy for each perps market based on defaults (overridable)
- added openPosition helper that calls commit and settle


markets/perps-market/test/integration/Orders/OffchainAsyncOrder.commit.test.ts
```diff
@@ -1,73 +1,39 @@
...
-      strategyType: ASYNC_OFFCHAIN_ORDER_TYPE, // OFFCHAIN
-      settlementDelay,
-      settlementWindowDuration,
-      priceVerificationContract,
-      feedId,
-      url: ASYNC_OFFCHAIN_URL,
-      disabled: false,
-      settlementReward,
-      priceDeviationTolerance,
-    });
```
sunnyvempati, line 70: all this gets auto created now

sunnyvempati: you can always override the defaults and get the defaults as needed

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.commit.test.ts
```diff
@@ -209,9 +175,9 @@ describe('Commit Offchain Async Order test', () => {
       it('emit event', async () => {
         await assertEvent(
           tx,
-          `OrderCommitted(${ethMarketId}, 2, ${ASYNC_OFFCHAIN_ORDER_TYPE}, ${bn(1)}, ${bn(1000)}, ${
-            startTime + 5
-          }, ${startTime + 5 + 120}, "${
+          `OrderCommitted(${ethMarketId}, 2, ${DEFAULT_SETTLEMENT_STRATEGY.strategyType}, ${bn(
```
sunnyvempati, line 178: get the default value if you need it so you dont have to store it in the test

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.commit.test.ts
```diff
@@ -246,21 +212,19 @@ describe('Commit Offchain Async Order test', () => {
...
-            accountId: () => 2,
-            feedId: () => feedId,
-            startTime: () => startTime,
-            settlementDelay: () => settlementDelay,
-            offChainPrice: () => 1000,
-          },
-          {
+        before('settle', async () => {
+          const settlementTime = startTime + DEFAULT_SETTLEMENT_STRATEGY.settlementDelay + 1;
+          await fastForwardTo(settlementTime, provider());
```
sunnyvempati, line 217: this fast forward orchestration happens in the openPosition but when we're testing the actual commit and settle, we should do the time fast forward orchestration in the test so its clear imo. thoughts?

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.settle.test.ts
```diff
@@ -1,102 +1,77 @@
...
-      disabled: false,
-      settlementReward,
-      priceDeviationTolerance,
-    });
-  });
-
-  before('set skew scale', async () => {
-    await systems()
-      .PerpsMarket.connect(marketOwner())
-      .setFundingParameters(marketId, bn(100_000), bn(0));
```
sunnyvempati, line 51: setting it in the bootstrap call above

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.settle.test.ts
```diff
@@ -1,102 +1,77 @@
...
-      settlementDelay,
-      settlementWindowDuration,
-      priceVerificationContract,
-      feedId,
-      url: ASYNC_OFFCHAIN_URL,
-      disabled: false,
-      settlementReward,
-      priceDeviationTolerance,
-    });
-  });
```
sunnyvempati, line 46: auto created on perps market creation

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.settle.test.ts
```diff
@@ -1,102 +1,77 @@
...
-      accountId: () => 2,
-      sizeDelta: () => bn(1),
-      settlementStrategyId: () => 0,
-      settlementDelay: () => settlementDelay,
-      settlementWindowDuration: () => settlementWindowDuration,
-      acceptablePrice: () => bn(1_000),
-      trackingCode: ethers.constants.HashZero,
-    },
-    { systems, provider }
-  );
```
sunnyvempati, line 71: commitOrder now runs inside a before block.  with this change, we don't have to make all the variables functions

markets/perps-market/test/integration/bootstrap/bootstrapPerpsMarkets.ts
```diff
@@ -152,9 +175,26 @@ export const bootstrapPerpsMarkets: BootstrapPerpsMarketType = (data, chainState
         });
       }
 
+      let strategyId: ethers.BigNumber;
+      // create default settlement strategy
+      before('create default settlement strategy', async () => {
+        const strategy = {
+          ...DEFAULT_SETTLEMENT_STRATEGY,
+          ...(settlementStrategy ?? {}),
```
sunnyvempati, line 183: this is where you can pass in a settlementStrategy and it'll override the defaults

markets/perps-market/test/integration/helpers/commitHelper.ts
```diff
@@ -1,65 +1,40 @@
...
-    await assertEvent(
-      tx,
-      `OrderCommitted(${data.marketId()}, ${data.accountId()}, ${ASYNC_OFFCHAIN_ORDER_TYPE}, ${data.sizeDelta()}, ${data.acceptablePrice()}, ${
-        startTime + data.settlementDelay()
-      }, ${startTime + data.settlementDelay() + data.settlementWindowDuration()}, "${
-        ethers.constants.HashZero
-      }", "${await data.trader().getAddress()}"`,
-      chainState.systems().PerpsMarket
-    );
-  });
```
sunnyvempati, line 59: seems weird to check for an event in a helper to bootstrap system.  assuming this will be tested in the test for committing order?  

we could keep it if it provides value.  the other thing i noticed as i was thinning out the commitOrder is that we probably dont really need this helper, we can directly call commitOrder as it will only ever be used when testing commitOrder.  the rest of the time we would use the openPosition helper to create a position.

markets/perps-market/test/integration/helpers/settleHelper.ts
```diff
@@ -1,75 +1,50 @@
...
 
-type SettleOrderType = (data: SettleOrderData, chainState: IncomingChainState) => SettleOrderReturn;
-
-export const settleOrder: SettleOrderType = (data, chainState) => {
-  let tx: ethers.ContractTransaction;
-  let updateFee: ethers.BigNumber;
-
-  let pythPriceData: string, extraData: string;
-
-  before('fast forward to settlement time', async () => {
```
sunnyvempati, line 32: mostly removed the before blocks and made all of it top level and so it runs inside a before block

bachstatter (approved):

markets/perps-market/test/integration/helpers/commitHelper.ts
```diff
@@ -1,65 +1,40 @@
...
+  await commitTx.wait(); // force immediate confirmation to prevent flaky tests due to block timestamp
```
leomassazza, line 37: You should either get the tx time here and return it, or return the transactionExecutionResult (`const res = await commitTx.wait()`) too have the exact time of commitment. Any operation between this commit and when the user gets the timestamp to calculate the settlementTime can be delayed by 1 sec (at least)

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.commit.test.ts
leomassazza: This test is focusing on the commit, that was originally used the settleHelper to help confirm the order can be settled. Settlement is tested fully (wip) in another test

leomassazza (commented):
Love the cleanup you made. I added just a couple of comments

markets/perps-market/test/integration/helpers/commitHelper.ts
sunnyvempati: im thinking of just getting rid of this commit helper since all it does is call commit directly.

leomassazza (approved):

PR #1646, opened 2023-06-07
Feature/more market views

davidvuong:
This PR adds the following:

- A `getMarketSummary` view
- Exposes `currentFundingVelocity`
- Initial unit tests

### DIFF
--- a/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
+++ b/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
@@ -1,6 +1,8 @@
 //SPDX-License-Identifier: MIT
 pragma solidity >=0.8.11 <0.9.0;
 
+import {AsyncOrder} from "../storage/AsyncOrder.sol";
+
 /**
  * @title Perps market module
  */
@@ -35,4 +37,13 @@ interface IPerpsMarketModule {
     function getMarketSummary(
         uint128 marketId
     ) external view returns (MarketSummary memory summary);
+
+    /**
+     * @dev A pagination view starting from `cursor` to retrieve up to `amount` for a given `marketId`.
+     */
+    function getAsyncOrdersPaginated(
+        uint128 marketId,
+        uint256 cursor,
+        uint256 amount
+    ) external view returns (AsyncOrder.Data[] memory orders, uint256 nextCursor, uint256 pageSize);
 }
--- a/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
+++ b/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
@@ -37,13 +37,4 @@ interface IPerpsMarketModule {
     function getMarketSummary(
         uint128 marketId
     ) external view returns (MarketSummary memory summary);
-
-    /**
-     * @dev A pagination view starting from `cursor` to retrieve up to `amount` for a given `marketId`.
-     */
-    function getAsyncOrdersPaginated(
-        uint128 marketId,
-        uint256 cursor,
-        uint256 amount
-    ) external view returns (AsyncOrder.Data[] memory orders, uint256 nextCursor);
 }
--- a/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
+++ b/markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
@@ -5,6 +5,16 @@ pragma solidity >=0.8.11 <0.9.0;
  * @title Perps market module
  */
 interface IPerpsMarketModule {
+    struct MarketSummary {
+        int256 skew;
+        uint256 size;
+        uint256 maxOpenInterest;
+        int currentFundingRate;
+        int currentFundingVelocity;
+        uint indexPrice;
+        uint fillPrice;
+    }
+
     function skew(uint128 marketId) external view returns (int256);
 
     function size(uint128 marketId) external view returns (uint256);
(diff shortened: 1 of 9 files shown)


markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -47,6 +49,7 @@ library PerpsMarket {
         uint128 lastUtilizedLiquidationCapacity;
         // accountId => asyncOrder
         mapping(uint => AsyncOrder.Data) asyncOrders;
+        SetUtil.UintSet asyncOrdersSet;
```
sunnyvempati, line 52: using a `UintSet` is very expensive as it creates an array and a mapping.  Generally speaking, we shouldn't be adding additional storage variables to contracts if its not serving the contract itself.  View functions are ok, but adding storage is not a negligible cost unfortunately.

markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
```diff
@@ -1,10 +1,22 @@
...
 interface IPerpsMarketModule {
+    struct MarketSummary {
+        int256 skew;
+        uint256 size;
+        uint256 maxOpenInterest;
+        int currentFundingRate;
+        int currentFundingVelocity;
+        uint indexPrice;
+        uint fillPrice;
+    }
```
sunnyvempati, line 18: nicee

sunnyvempati (changes requested):
we should chat so i can understand the severity of the problem for keepers.  

adding storage is a pretty expensive solution (esp uint set).  wondering if subgraphs are enough to solve this, even though they're not as reliant.  otherwise, we'll have to figure out a more cost effective solution (maybe a linked list for cheaper insertion/deletion?)

davidvuong:
> we should chat so i can understand the severity of the problem for keepers.

Yeah sure. A keeper on startup and throughout its lifetime needs know what orders are pending for execution. So on initial startup, a keeper can:

1. Spam `eth_getLogs` from block `n` until they reach `latestBlock`
2. Do (1) but cached and track the last block number to avoid reindexing on restart
3. Use an OTS indexing solution like The Graph
4. Pagination on the contract

Contract pagination is by far the fastest, easiest to use, and cheapest from the perspective of a keeper. Knowing the size, you can page out calls in a single `eth_call` to a multicall contract. With a large enough page size, you can probably fit everything you need in a single call.

We have something like this in PerpsV2 and it's quite useful. 

> we'll have to figure out a more cost effective solution (maybe a linked list for cheaper insertion/deletion?)

I'm easy. This PR is still in draft and WIP. There are more changes to be introduced. That said, just looking at `UintSet`, both `.add` and `.remove` operations are constant time? `.add` is just a `.push` and `.remove` is an index lookup, swap with tail and then `.pop`. I haven't looked at gas cost yet though.

sunnyvempati:
@davidvuong 

> Contract pagination is by far the fastest, easiest to use, and cheapest from the perspective of a keeper. Knowing the size, you can page out calls in a single eth_call to a multicall contract. With a large enough page size, you can probably fit everything you need in a single call.

Cheap for keepers, but not for traders committing orders.  My hope is we have much better eventing than v2x and let the graph / other solutions being discussed fill in the gaps for keepers.  

Also, I think we can layer this solution for keepers in later once we get the core functionality completed.  This isn't really necessary for getting perps v3 into audit.  Hopefully it will give us an easier way to assess the gas costs once we're code complete and can have baseline gas fee results.

leomassazza:
After some thinking, I agree we should (if possible) improve on-chain for users (traders) and off-chain for keepers/monitoring.

davidvuong:
Pagination has been removed as per discussion above.

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -3,6 +3,8 @@ pragma solidity >=0.8.11 <0.9.0;
 
 import {DecimalMath} from "@synthetixio/core-contracts/contracts/utils/DecimalMath.sol";
 import {SafeCastU256, SafeCastI256, SafeCastU128} from "@synthetixio/core-contracts/contracts/utils/SafeCast.sol";
+import {AccessError} from "@synthetixio/core-contracts/contracts/errors/AccessError.sol";
```
sunnyvempati, line 6: can omit

sunnyvempati (approved):
🚢 it

markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -31,6 +32,7 @@ contract AsyncOrderModule is IAsyncOrderModule {
     using GlobalPerpsMarket for GlobalPerpsMarket.Data;
     using SafeCastU256 for uint256;
     using SafeCastI256 for int256;
+    using SetUtil for SetUtil.UintSet;
```
sunnyvempati, line 35: .

markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity >=0.8.11 <0.9.0;
 import {SafeCastU256, SafeCastI256} from "@synthetixio/core-contracts/contracts/utils/SafeCast.sol";
 import {DecimalMath} from "@synthetixio/core-contracts/contracts/utils/DecimalMath.sol";
 import {Account} from "@synthetixio/main/contracts/storage/Account.sol";
+import {SetUtil} from "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
```
sunnyvempati, line 7: .

PR #1647, opened 2023-06-07
updated orderFees

fritzschoff:


markets/perps-market/test/integration/Market/MarketConfigration.test.ts
```diff
@@ -61,34 +62,110 @@ describe('MarketConfiguration', async () => {
...
-        marketId,
-        fixture.initialMarginFraction,
-        fixture.maintenanceMarginFraction,
-        fixture.liquidationRewardRatioD18,
-        fixture.maxLiquidationLimitAccumulationMultiplier
-      );
-    await systems()
-      .PerpsMarket.connect(marketOwner)
-      .setLockedOiPercent(marketId, fixture.lockedOiPercent);
+  it('owner can set all market configurations properties and events are emitted', async () => {
```
sunnyvempati, line 65: in general, better to split these into separate `it` blocks but since these are events,  if even one of these events fails, it'll bubble it up so not a huge deal here.

sunnyvempati (approved):

PR #1648, opened 2023-06-07
docs: fix decimal math function natspec comments

tommyrharper:


noahlitvin:
I think the comments are correct as-is? It's pointing the reader to reference the analogous functions for unsigned ints?

PR #1652, opened 2023-06-08
Liquidation margin tests

sunnyvempati:
- fixed a few skew scale issues
- margin test tracked here: https://docs.google.com/spreadsheets/d/1eiR9riut62MrTvUR-qlHCfSJw0lGZ9Sg0pPhqOBPEsU/edit#gid=440773928

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -134,15 +134,15 @@ library PerpsAccount {
             revert AccountLiquidatable(accountId);
         }
 
+        // availableMargin can be assumed to be positive since we check for isEligible for liquidation prior
+        availableWithdrawableCollateralUsd = availableMargin.toUint() - requiredMaintenanceMargin;
+
         if (amountToWithdraw > availableWithdrawableCollateralUsd) {
             revert InsufficientCollateralAvailableForWithdraw(
                 availableWithdrawableCollateralUsd,
                 amountToWithdraw
             );
         }
-
-        // availableMargin can be assumed to be positive since we check for isEligible for liquidation prior
-        availableWithdrawableCollateralUsd = availableMargin.toUint() - requiredMaintenanceMargin;
     }
 
     function getTotalCollateralValue(Data storage self) internal view returns (uint) {
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -14,6 +14,8 @@ import {GlobalPerpsMarket} from "./GlobalPerpsMarket.sol";
 import {GlobalPerpsMarketConfiguration} from "./GlobalPerpsMarketConfiguration.sol";
 import {PerpsMarketConfiguration} from "./PerpsMarketConfiguration.sol";
 
+import "hardhat/console.sol";
+
 uint128 constant SNX_USD_MARKET_ID = 0;
 
 /**
@@ -60,7 +62,7 @@ library PerpsAccount {
     )
         internal
         view
-        returns (bool isEligible, uint256 availableMargin, uint256 requiredMaintenanceMargin)
+        returns (bool isEligible, int256 availableMargin, uint256 requiredMaintenanceMargin)
     {
         availableMargin = getAvailableMargin(self, accountId);
 
@@ -69,7 +71,7 @@ library PerpsAccount {
         }
 
         requiredMaintenanceMargin = getAccountMaintenanceMargin(self, accountId);
-        isEligible = requiredMaintenanceMargin > availableMargin;
+        isEligible = requiredMaintenanceMargin.toInt() > availableMargin;
     }
 
     function flagForLiquidation(Data storage self, uint128 accountId) internal {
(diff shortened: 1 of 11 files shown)


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -126,7 +126,7 @@ library PerpsAccount {
     ) internal view returns (uint256 availableWithdrawableCollateralUsd) {
         (
             bool isEligible,
-            uint256 availableMargin,
+            int256 availableMargin,
```
leomassazza, line 129: can it be negative at some point?

leomassazza (approved):

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -141,7 +141,8 @@ library PerpsAccount {
             );
         }
 
-        availableWithdrawableCollateralUsd = availableMargin - requiredMaintenanceMargin;
+        // availableMargin can be assumed to be positive since we check for isEligible for liquidation prior
+        availableWithdrawableCollateralUsd = availableMargin.toUint() - requiredMaintenanceMargin;
```
bachstatter, line 145: This is fixed in my PR, but this statement needs to be above the if statement on line 137

markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
```diff
@@ -1,38 +1,243 @@
...
+        const [positionPnl] = await systems().PerpsMarket.getOpenPosition(
+          2,
+          perpsMarkets()[i].marketId()
+        );
+        assertBn.equal(positionPnl, pnl);
+      });
+    });
+
+    it('has correct available margin', async () => {
+      assertBn.equal(await systems().PerpsMarket.getAvailableMargin(2), bn(12_550));
```
bachstatter, line 186: I was gonna comment saying that a comment explaining how we arrive to 12_550 might make sense. 
Then I looked up getAvailableMargin, so it is `totalCollateralValue` + `getAccountPnl`. 

Total collateral value is using a very real life price by looking up what price we would get if we were to sell the collateral. Account pnl on the other hand is using oracle rather than skew adjusted price. I might be wrong but it feels like using fillPrice for pnl would be more accurate represent the true collateral value

bachstatter (approved):
Left some comments, but dont think anything is blocking this PR

markets/perps-market/contracts/storage/PerpsAccount.sol
sunnyvempati: oh yeah; we should get that merged.  i'll fix it here so maybe you can rebase

markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
sunnyvempati: well account pnl has to do with the perps market price, whereas total collateral value looks at the spot market price (which is related to its own risk mechanism).  we wouldn't necessarily look at spot synth price for a perps market.  hope this makes sense.

markets/perps-market/contracts/storage/PerpsAccount.sol
sunnyvempati: it can, yes.  account pnl could continue to get very negative, even past the liquidation margin until all your margin is gone + more.  hopefully liquidation bots liquidate the account before this happens, but still possible.

good test case actually for liquidation of an account with negative margin!

markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
bachstatter: Yea I didn't mean that we should use the spot market price. I meant the perps skew adjusted price

sunnyvempati: oh yeah, very good point.  in my opinion, even if we use fill price, it's not really an accurate price since when the trader ends up modifying their position, would be average of the current skew price and the skew after the amt they're modifying.  and the price delta gets better/worse depending on how large the size is vs the fill price calculated with skew 0.  

i think i rmr afif mentioning to use on chain price; but it might be good to bring this up during our meetings with afif this week.  i'll take a note and double check 👍🏽 .  thanks for pointing it out.

bachstatter: Cool, and yea fill price based on the size of the position would be the way I expect it to work. 

It does ring a bell that we asked afif about this before. I just cant remember any argument for why we wouldn't use fill price.

sunnyvempati: the argument i'd make is i wouldn't think the trader should have a fluctuating liquidation based on the market's skew (fill price is based on market skew).  imagine if you had a long open, and there's a massive long skew.  and i get liquidated even when price didn't change, but skew did when some other trader closed a really large short position; that doesn't seem fair.

bachstatter: Hmm yea, that is an argument. Using fill price still seems better to me, that is the current price of the asset in our platform. Just as changing market skew affect the price if I want to close a position I would expect changing market skew to affect liquidation. 

But as long as it gets discussed and it's a conscious decision I'm happy

PR #1653, opened 2023-06-09
added cancelOrder to AsyncOrderModule

fritzschoff:


bachstatter (approved):
Looks good to me! Probably want another review though

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      accountId: () => 2,
+      collaterals: [
+        {
+          snxUSDAmount: () => bn(10_000),
+        },
+      ],
+    });
+  });
+
+  it('commit can not be canceled when settlement window is withing range', async () => {
```
bachstatter, line 47: Maybe add a test case to assert revert when order doesn't exists

fritzschoff:
> Looks good to me! Probably want another review though

yep. Waiting on sunny if we actually want that feature, cause there is no incentive to do so if you could just override your oder. Waiting on Sunny and Leo

markets/perps-market/contracts/interfaces/IAsyncOrderModule.sol
```diff
@@ -52,7 +52,12 @@ interface IAsyncOrderModule {
...
-    // function cancelOrder(uint128 marketId, uint128 asyncOrderId) external;
+    function cancelOrder(uint128 marketId, uint128 accountId) external;
+
+    function getOrder(
+        uint128 marketId,
+        uint128 accountId
+    ) external returns (AsyncOrder.Data memory);
 
     // function getAsyncOrderClaim(
     //     uint128 marketId,
```
sunnyvempati, line 63: can remove this

sunnyvempati (commented):
hold on this.  we have to add a new configuration parameter called `executionTime`, and we can only cancel after this time instead of settlement time.  they will have different meanings now

fritzschoff:
> hold on this. we have to add a new configuration parameter called `executionTime`, and we can only cancel after this time instead of settlement time. they will have different meanings now

alright. I will wait until I get the green light from you

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  it('commit can be canceled when settlement window is outside of range', async () => {
+    await fastForwardTo((await getTime(provider())) + 9000000000, provider());
+    const orderBeforeCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
+    assertBn.equal(orderBeforeCancelation.sizeDelta, bn(1));
+    await systems().PerpsMarket.cancelOrder(ethMarketId, 2);
+    const orderAfterCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
+    assertBn.equal(orderAfterCancelation.sizeDelta, bn(0));
+  });
```
sunnyvempati, line 71: the pattern we're using for tests is where the transactions are in before blocks and only hte assertions live in the it blocks.  this makes it easier to figure out where errors are occuring if there are any in the prior transactions.  so in this scenario, maybe something like this would be good:
```
before('commit order', () => {
  await systems().PerpsMarket.commitOrder({
    marketId: ethMarketId,
    accountId: 2,
    sizeDelta: bn(1),
    settlementStrategyId: 0,
    acceptablePrice: bn(1000),
    trackingCode: ethers.constants.HashZero,
  });
});

describe('error cases', () => {
  it('reverts when settlement window is not expired', async () => {
    await assertRevert(
      systems().PerpsMarket.cancelOrder(ethMarketId, 2),
      `SettlementWindowNotExpired`,
      systems().PerpsMarket
    );
  });
});

describe('success', () => {
  before('fast forward past settlement window', () => {
    await fastForwardTo((await getTime(provider())) + 9000000000, provider());
  });

  let cancelTx: TransactionResponse;
  before('cancel order', async () => {
    cancelTx = await systems().PerpsMarket.cancelOrder(ethMarketId, 2);
  });

  it('resets order correctly', async () => {
    // check order is reset
  });

  it('emits event', async () => {
    // check event is emitted
  })
});
```

Note: check for event emitted on cancellation as well

markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -121,6 +121,33 @@ contract AsyncOrderModule is IAsyncOrderModule {
...
+    }
+
+    function cancelOrder(uint128 marketId, uint128 accountId) external override {
+        AsyncOrder.Data storage order = PerpsMarket.loadValid(marketId).asyncOrders[accountId];
+        SettlementStrategy.Data storage settlementStrategy = PerpsMarketConfiguration
+            .load(marketId)
+            .settlementStrategies[order.settlementStrategyId];
+
+        order.checkOutsideSettlementWindow(settlementStrategy);
+        order.reset();
```
sunnyvempati, line 138: emit cancellation event

sunnyvempati (changes requested):

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -90,6 +96,21 @@ library AsyncOrder {
         }
     }
 
+    function checkOutsideSettlementWindow(
```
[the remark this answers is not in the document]
fritzschoff: yep sounds good!

markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -121,6 +121,26 @@ contract AsyncOrderModule is IAsyncOrderModule {
...
+        uint128 accountId
+    ) public view override returns (AsyncOrder.Data memory) {
+        return PerpsMarket.loadValid(marketId).asyncOrders[accountId];
+    }
+
+    function cancelOrder(uint128 marketId, uint128 accountId) external override {
+        AsyncOrder.Data storage order = PerpsMarket.loadValid(marketId).asyncOrders[accountId];
+        if (order.sizeDelta == 0) {
+            revert OrderDoesNotExists(marketId, accountId);
+        }
```
sunnyvempati, line 135: there is a function on order called `checkValidity` which does this check for ya

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -90,6 +96,21 @@ library AsyncOrder {
...
+        Data storage self,
+        SettlementStrategy.Data storage settlementStrategy
+    ) internal view {
+        uint settlementExpiration = self.settlementTime +
+            settlementStrategy.settlementWindowDuration;
+        if (block.timestamp < self.settlementTime || block.timestamp < settlementExpiration) {
```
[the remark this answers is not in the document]
sunnyvempati: check suggestion

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        systems().PerpsMarket
+      );
+    });
+  });
+
+  describe('success', async () => {
+    it('commit can be canceled when settlement window is outside of range', async () => {
+      await fastForwardTo((await getTime(provider())) + 9000000000, provider());
+      const orderBeforeCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
+      assertBn.equal(orderBeforeCancelation.sizeDelta, bn(1));
```
sunnyvempati, line 73: move this check to right after `commit order`.  this assert is a nice sanity check as the consequence of committing an order.

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        systems().PerpsMarket.cancelOrder(ethMarketId, 3),
+        'OrderDoesNotExists("1", "3")',
+        systems().PerpsMarket
+      );
+    });
+  });
+
+  describe('success', async () => {
+    it('commit can be canceled when settlement window is outside of range', async () => {
+      await fastForwardTo((await getTime(provider())) + 9000000000, provider());
```
sunnyvempati, line 71: move to before block.  in case you add other tests after this describe, its easier to understand state of the chain via before blocks.  if we limit assertions only to `it` blocks, we dont have to worry about state changing within it blocks

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      const orderBeforeCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
+      assertBn.equal(orderBeforeCancelation.sizeDelta, bn(1));
+      const tx = await systems().PerpsMarket.cancelOrder(ethMarketId, 2);
+      // Ignore settlement time and price for the event
+      await assertEvent(tx, `OrderCanceled(1, 2,`, systems().PerpsMarket);
+      const orderAfterCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
+      assertBn.equal(orderAfterCancelation.sizeDelta, bn(0));
```
sunnyvempati, line 78: could be better to move these into different it blocks:
```
it('emits event when order is cancelled');
it('resets the order');
```

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
fritzschoff: created a new before. Cant put it into existing before block because this will destroy the `error`assumptions

sunnyvempati: yeah thats what i meant was, create new `before`.  this looks good 👍🏽

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+    });
+  });
+
+  describe('success', async () => {
+    before('fast forward outside of settlement window', async () => {
+      await fastForwardTo((await getTime(provider())) + 9000000000, provider());
+    });
+    it('emits event when order is canceled', async () => {
+      const tx = await systems().PerpsMarket.cancelOrder(ethMarketId, 2);
```
sunnyvempati, line 75: would move this to a before as well.  the only reason to keep it in the `it` block is if you're asserting the call will revert.  moving it to before makes it clear whats all happening before any of the it blocks are executed.

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.cancel.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      // Ignore settlement time and price for the event
+      await assertEvent(tx, `OrderCanceled(1, 2,`, systems().PerpsMarket);
+    });
+    it('sizeDelta is 0 when order is canceled', async () => {
+      const orderAfterCancelation = await systems().PerpsMarket.getOrder(ethMarketId, 2);
```
sunnyvempati, line 80: can keep this here since its just a view to assert data, and doesn't affect chain 👍🏽

sunnyvempati (approved):
pre-approving but a couple small notes. 

good work on this! 🙌🏽

markets/perps-market/contracts/interfaces/IAsyncOrderModule.sol
```diff
@@ -32,6 +32,13 @@ interface IAsyncOrderModule {
         address indexed settler
     );
 
+    event OrderCanceled(
+        uint128 indexed marketId,
+        uint128 indexed accountId,
+        uint256 settlementTime,
+        uint256 acceptablePrice
```
leomassazza, line 39: why do we need the acceptablePrice? (is ok if it was discussed offline, just to know if there's a reason or just because we have the data available)

fritzschoff: we dont need it necessarily. I added it because it was nice to have

PR #1655, opened 2023-06-12
Update synth wrapp/unwrapp event

Rickk137:


sunnyvempati (approved):

bachstatter (commented):
The description is a little sparse, so i'm not sure the thinking behind this. But just wanted to add that if this is for a subgraph we can already get the msg.sender by `event.transaction.from`

PR #1657, opened 2023-06-13
Withdraw tests

bachstatter:
- Add `InsufficientCollateral` error to avoid panic underflow weh users try to withdraw more than avail collateral
- Add some more params to the `MaxCollateralExceeded`
- Add test case for `InsufficientCollateral`
- Add test case for  `InsufficientCollateralAvailableForWithdraw`
- Add test case for  `AccountLiquidatable` 
- Add test case to make sure we can withdraw with open positions

For now the tests for `InsufficientCollateralAvailableForWithdraw` and `AccountLiquidatable` are fairly simple. There's lots of permutations that could be tested, but I think more scenarios are covered in the liquidation tests. And I think it make sense to keep it like this.. 

There is one more case we could write a tests for, and that is making sure we cant withdraw when an account is flagged, will tackle that in a separate PR. 

Let me know if there's good reason to keep errors lean with less params.


markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -47,16 +53,28 @@ library GlobalPerpsMarket {
...
+                    synthAmount.toUint()
+                );
             } else {
                 self.collateralAmounts[synthMarketId] += synthAmount.toUint();
             }
         } else {
+            uint synthAmountAbs = MathUtil.abs(synthAmount);
+            if (collateralAmount < synthAmountAbs) {
+                revert InsufficientCollateral(synthMarketId, collateralAmount, synthAmountAbs);
+            }
```
sunnyvempati, line 76: makes sense

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
```diff
@@ -114,4 +117,155 @@ describe('ModifyCollateral Withdraw', () => {
...
+        );
+      });
+    });
+    describe('withdraw with open position', () => {
+      it('allow withdraw when its less than "collateral available for withdraw"', async () => {
+        before('withdraw allowed amount', async () => {
+          await systems()
+            .PerpsMarket.connect(trader1())
+            .modifyCollateral(trader1AccountId, sUSDSynthId, bn(-17000));
+        });
```
sunnyvempati, line 227: this is odd; you have the `before` inside the `it` block.  surprised the test passed

sunnyvempati (commented):
have one change but pre approving.

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
bachstatter: Opsie, will fix

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
bachstatter: I noticed we have `InsufficientMargin` in the AsyncOrder file. Do we have a definition on the differences between margin and collateral? And do we want to refuse errors as much as possible?

markets/perps-market/test/integration/Account/ModifyCollateral.withdraw.test.ts
bachstatter: Fixed!

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
bachstatter: Hmm we also have a `error InsufficientMarginError(uint leftover);` in PerpsAccount.

sunnyvempati: we should distinguish them in this way imo:

- `InvalidWithdrawalAmount(amountAvailableToWithdraw, amountRequested)`: this one for the above scenario where trader is attempting to withdraw more than the entire market has.
- `InsufficientMargin()`: currently this is thrown when validating orders.  I have a TODO to create separate errors for which this is being reused during validation.
- `InsufficientCollateral()`: this should occur very rarely cuz by the time you're deducting, you've already checked you have enough to cover fees.  so yeah lets rename `InsufficientMarginError` to this or something similar?

all these are suggestions; feel free to suggest other names and ways to distinguish 👍🏽

sunnyvempati (approved):

PR #1660, opened 2023-06-14
Fill price size & price params

bachstatter:
Previously it wasn't really a fill price, but a skew adjusted oracle price.
After chatting a bit about it we also thought it can be removed from summaries. We could add a skew adjusted price, but given that frontends will want to use a more up to date offchain price we opted to remove it from the summaries view.


markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
```diff
@@ -29,7 +28,7 @@ interface IPerpsMarketModule {
 
     function indexPrice(uint128 marketId) external view returns (uint);
 
-    function fillPrice(uint128 marketId) external returns (uint);
+    function fillPrice(uint128 marketId, int orderSize, uint price) external returns (uint);
```
sunnyvempati, line 31: i'm not really sure this function is needed.   liek if you have orderSize and price, you prolly have the skewscale and skew to back into this price.  non blocking

sunnyvempati (approved):

markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
bachstatter: Not necessarily, price could be fetched from Pyth and size from an input field.. But sure you can always fetch the skew + skewScale and calculate it yourself. I think it's an reasonable view to have. But yea, lets keep track on it if it's not useful I'm all for removing

PR #1661, opened 2023-06-14
Account id refactor

bachstatter:
This PR adds `id` to the data struct in `PerpsAccount`. This means we dont have to pass around `accountId` everywhere.
Given that we inherit the `createAccount` function from the core `AccountModule` we cant change the `createAccount` to set the id to the `PerpsAccount`. Instead we set the id (if it's missing) in the `modifyCollateral` method


fritzschoff (approved):

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -44,6 +44,9 @@ contract PerpsAccountModule is IAccountModule {
         globalPerpsMarket.checkLiquidation(accountId);
 
         PerpsAccount.Data storage account = PerpsAccount.load(accountId);
+        if (account.id == 0) {
+            account.addAccountId(accountId);
```
sunnyvempati, line 48: can just do `account.id = accountId` and nix the `addAccountId` function

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -103,6 +102,10 @@ library PerpsAccount {
         self.collateralAmounts[synthMarketId] += amountToAdd;
     }
 
+    function addAccountId(Data storage self, uint128 accountId) internal {
+        self.id = accountId;
+    }
```
sunnyvempati, line 107: x

sunnyvempati (approved):
one suggestion; pre approving

markets/perps-market/contracts/modules/PerpsAccountModule.sol
bachstatter: I see, will do

PR #1665, opened 2023-06-15
Node 18.16.0

noisekit:
It's deprecated now.

This is what Vercel tells us for other projects:

> Error: Node.js version 16.x has reached End-of-Life. Deployments created on or after 2023-08-15 will fail to build. Please set "engines": { "node": "18.x" } in your `package.json` file to use Node.js 18.


bachstatter:
Looks like the tests didn't like the upgrade :(

bachstatter (approved):
Looks good to me

PR #1666, opened 2023-06-16
Codecov and tests coverage

noisekit:
- Codecov must wait for CI to pass harder
- Manually list codecov flags to report on
- Fix coverage scripts to actually run coverage reports

Coverage SKIPPED for perps-market and spot-market for now as there is some difference in how hardhat networks behaves comparing to anvil

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
             oracle-manager,
             spot-market,
             perps-market,
-            main-subgraph,
+            core-subgraph,
           ]
         include:
           - package: core-contracts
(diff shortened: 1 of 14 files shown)


codecov.yml
```diff
@@ -1,9 +1,60 @@
...
+    paths:
+      - utils/core-utils/src
+    carryforward: true
+
+  core-modules:
+    paths:
+      - utils/core-modules/contracts
+    carryforward: true
+
+  router:
```
bachstatter, line 32: This got moved out right?

bachstatter (approved):
One comment about

PR #1668, opened 2023-06-19
Add price validity window param

leomassazza:


markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -22,7 +22,6 @@ contract AsyncOrderModule is IAsyncOrderModule {
     using DecimalMath for uint256;
     using DecimalMath for int64;
     using PerpsPrice for PerpsPrice.Data;
-    using AsyncOrder for AsyncOrder.Data;
```
leomassazza, line 25: remove duplicate

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -21,7 +21,6 @@ contract PerpsAccountModule is IAccountModule {
     using AsyncOrder for AsyncOrder.Data;
     using SafeCastU256 for uint256;
     using SafeCastI256 for int256;
-    using PerpsAccount for PerpsAccount.Data;
```
leomassazza, line 24: remove duplicate

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -7,7 +7,6 @@ import {IAsyncOrderModule} from "../interfaces/IAsyncOrderModule.sol";
 import {SettlementStrategy} from "./SettlementStrategy.sol";
 import {Position} from "./Position.sol";
 import {PerpsMarketConfiguration} from "./PerpsMarketConfiguration.sol";
-import {SettlementStrategy} from "./SettlementStrategy.sol";
```
leomassazza, line 10: duplicate import

markets/perps-market/contracts/storage/SettlementStrategy.sol
```diff
@@ -26,6 +26,10 @@ library SettlementStrategy {
          * @dev the duration of the settlement window, after which committed orders can be cancelled.
          */
         uint256 settlementWindowDuration;
+        /**
+         * @dev the duration of the price window, after which price is not valid.
+         */
+        uint256 priceWindowDuration;
```
bachstatter, line 32: We're not using setters for these? Is the idea that a new Strategy would be created if we want to update these?

bachstatter (approved):

markets/perps-market/contracts/storage/SettlementStrategy.sol
leomassazza: originally, yes. We can add later a module to update a settlement strategy if needed, but here and in spot, you can only create a new one and set the old as disabled.

sunnyvempati (commented):
hold off on this until the settlement strategy refactor is in?

sunnyvempati (approved):
i think we are good to merge this.  we're probably keeping the current settlement strategy impl

PR #1669, opened 2023-06-19
Max oi cap

fritzschoff:
this check needs to be implemented for every order type. Limit and market


markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -148,6 +149,15 @@ library AsyncOrder {
         PerpsMarketConfiguration.Data storage marketConfig = PerpsMarketConfiguration.load(
             order.marketId
         );
+        uint256 sizeDeltaInUint256 = order.sizeDelta < 0
+            ? order.sizeDelta.mulDecimal(-1).toUint()
+            : order.sizeDelta.toUint();
+        if (marketConfig.maxMarketValue < perpsMarketData.size + sizeDeltaInUint256) {
```
bachstatter, line 155: I think we need some additional logic here..
The `perpsMarketData.size` is currently storing the absolute value of all shorts and long. I think `maxMarketValue` is suppose to be a cap on either side, at least that's how it works on perps v2.

So I think we need to track `shortSize` and `longSize` in `PerpsMarket.sol`.  (I would also think it's clearer if we name them `shortOI` and `longOI`).
Maybe @sunnyvempati can shine in if this is suppose to work differently compared to preps v2.

bachstatter: After talking to sunny, We can figure it out by looking at size + skew. see https://github.com/Synthetixio/synthetix-v3/blob/main/markets/perps-market/contracts/storage/PerpsMarket.sol#L226

fritzschoff: cool thanks! Will have a look!

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -199,7 +200,22 @@ library AsyncOrder {
         // TODO: validate position size
         oldPosition = PerpsMarket.load(order.marketId).positions[order.accountId];
 
+        if (
+            PerpsMarket.validatePositionSize(
+                perpsMarketData,
+                marketConfig.maxMarketValue,
+                oldPosition.size,
+                order.sizeDelta
+            )
```
sunnyvempati, line 209: let the `validatePositionSize` just revert instead of it returning a bool.  

that function needs to be refactored; but first would write the tests and make sure its working properly.

markets/perps-market/test/integration/Market/Size.test.ts
```diff
@@ -11,6 +11,7 @@ describe('Market - size test', () => {
         name: 'Ether',
         token: 'snxETH',
         price: bn(2000),
+        maxMarketValue: bn(9999999),
```
sunnyvempati, line 14: instead of needing to pass it in everywhere, in the `bootstrapPerpsMarket`, have a default value it can set it to when its not present 👍🏽

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.commit.test.ts
```diff
@@ -86,6 +87,28 @@ describe('Commit Offchain Async Order test', () => {
...
+            marketId: ethMarketId,
+            accountId: 2,
+            sizeDelta: bn(90_000),
+            settlementStrategyId: 0,
+            acceptablePrice: bn(1450), // 5% slippage
+            trackingCode: ethers.constants.HashZero,
+          }),
+        `MaxOpenInterestReached(${ethMarketId}, ${bn(5000).toString()})`
+      );
+    });
```
sunnyvempati, line 111: unfortunately, there are many other cases to be tested here. for one, we don't test if there's a negative sizes.

sunnyvempati (changes requested):
I'd recommend creating a new test file for testing max market size.  Some scenarios we can test:

- max market size is 100, current size is 50, someone creates new position for -100
- max market size is 100, current size is 100, someone creates new position for -101 (fails)
- one trader going from long to short, test both positive/negative cases
any others?  

let's be as thorough as possible.  pretty much every part of `validatePositionSize` should be tested with different order scenarios.  pretty critical to get this right 👍🏽

fritzschoff:
> I'd recommend creating a new test file for testing max market size. Some scenarios we can test:
> 
> * max market size is 100, current size is 50, someone creates new position for -100
> * max market size is 100, current size is 100, someone creates new position for -101 (fails)
> * one trader going from long to short, test both positive/negative cases
>   any others?
> 
> let's be as thorough as possible. pretty much every part of `validatePositionSize` should be tested with different order scenarios. pretty critical to get this right 👍🏽

on it. Found a TODO comment from you in the Size.test.ts file where you wrote `// TODO: test maxMarketSize here as well`

markets/perps-market/test/integration/bootstrap/bootstrapPerpsMarkets.ts
```diff
@@ -126,6 +126,13 @@ export const bootstrapPerpsMarkets = (
         );
       });
 
+      before('set max market value', async () => {
+        await contracts.PerpsMarket.connect(marketOwner).setMaxMarketValue(
+          marketId,
+          bn(10_000_000)
```
sunnyvempati, line 132: dont hardcode this for every market.  create an optional parameter that gets passed in here.  if it doesn't exist then default to this value

sunnyvempati (changes requested):
most of the comments are nonblocking, but lets remove the hardcoded value in `bootstrapPerpsMarket`

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -233,43 +233,43 @@ library PerpsMarket {
...
         uint maxSize,
         int oldSize,
         int newSize
-    ) internal view returns (bool) {
+    ) internal view {
         // Allow users to reduce an order no matter the market conditions.
-        if (MathUtil.sameSide(oldSize, newSize) && MathUtil.abs(newSize) <= MathUtil.abs(oldSize)) {
-            return false;
+        if (
+            !(MathUtil.sameSide(oldSize, newSize) && MathUtil.abs(newSize) <= MathUtil.abs(oldSize))
```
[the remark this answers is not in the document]
fritzschoff: you mean for readability ?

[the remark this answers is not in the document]
fritzschoff: there is a comment above the if statement

markets/perps-market/test/integration/bootstrap/bootstrapPerpsMarkets.ts
```diff
@@ -127,6 +127,13 @@ export const bootstrapPerpsMarkets = (
         );
       });
 
+      before('set max market value', async () => {
+        await contracts.PerpsMarket.connect(marketOwner).setMaxMarketValue(
+          marketId,
+          maxMarketValue ? maxMarketValue : bn(10_000_000)
```
sunnyvempati, line 133: 🖖🏽

PR #1671, opened 2023-06-20
Refactor price feed data

sunnyvempati:
- Setup nodes for settlement and index prices.
- Refactor settlement strategies to only have one / market.

TODO:
- [ ] refactor test helpers to use new price nodes
- [ ] fix tests

### DIFF
--- a/markets/perps-market/contracts/modules/MarketConfigurationModule.sol
+++ b/markets/perps-market/contracts/modules/MarketConfigurationModule.sol
@@ -5,37 +5,27 @@ import {IMarketConfigurationModule} from "../interfaces/IMarketConfigurationModu
 import {SettlementStrategy} from "../storage/SettlementStrategy.sol";
 import {PerpsMarketConfiguration} from "../storage/PerpsMarketConfiguration.sol";
 import {PerpsMarket} from "../storage/PerpsMarket.sol";
+import {PerpsPrice} from "../storage/PerpsPrice.sol";
 import {OrderFee} from "../storage/OrderFee.sol";
 
 contract MarketConfigurationModule is IMarketConfigurationModule {
     using PerpsMarket for PerpsMarket.Data;
+    using PerpsPrice for PerpsPrice.Data;
 
-    function addSettlementStrategy(
+    uint256 private constant _MIN_SETTLEMENT_DELAY = 1;
+
+    function setSettlementStrategy(
         uint128 marketId,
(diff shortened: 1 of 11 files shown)


markets/perps-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -5,37 +5,27 @@ import {IMarketConfigurationModule} from "../interfaces/IMarketConfigurationModu
...
 
-        strategy.settlementDelay = strategy.settlementDelay == 0 ? 1 : strategy.settlementDelay;
-
-        PerpsMarketConfiguration.Data storage config = PerpsMarketConfiguration.load(marketId);
-        strategyId = config.settlementStrategies.length;
-
-        config.settlementStrategies.push(strategy);
-        emit SettlementStrategyAdded(marketId, strategy);
-    }
+        strategy.settlementDelay = strategy.settlementDelay == 0
```
bachstatter, line 23: Maybe a tiny bit better to do `strategy.settlementDelay < _MIN_SETTLEMENT_DELAY` if we update to `_MIN_SETTLEMENT_DELAY` to something other than 1

markets/perps-market/contracts/storage/PerpsPrice.sol
```diff
@@ -27,46 +28,18 @@ library PerpsPrice {
...
-        price = INodeModule(factory.oracle).process(self.feedId);
+    function getSettlementPrice(uint128 marketId) internal view returns (uint price) {
+        bytes32 nodeId = load(marketId).settleNodeId;
+        return INodeModule(PerpsMarketFactory.load().oracle).process(nodeId).price.toUint();
     }
 
     function getCurrentPrice(uint128 marketId) internal view returns (uint price) {
-        return getCurrentPriceData(marketId).price.toUint();
+        bytes32 nodeId = load(marketId).indexNodeId;
+        return INodeModule(PerpsMarketFactory.load().oracle).process(nodeId).price.toUint();
```
bachstatter, line 38: One though about this new pattern.. So this can revert if the price is too stale, right?
This price function is used by many views, example: `getMarketSummary`. 
So does this means that a frontend calling `getMarketSummary` might not get any result until them or someone else request/ pay for a price update?

sunnyvempati:
closing this PR; after further discussion, we're going to keep settlement strategy array to scale to other oracle providers in case they support pull based

PR #1673, opened 2023-06-20
Make sure some of our collateral tests are non usd

bachstatter:
Make sure the withdrawal tests have some non snxUSD collateral


leomassazza (approved):
LGTM (we might need to make an adjustment, but that will change the functionality. See https://discord.com/channels/413890591840272394/1113827824776777758/1121088976170778644

PR #1675, opened 2023-06-20
Max liquidatable amount functionality

sunnyvempati:
- refactors the max liquidation amount per market as per this doc: https://docs.google.com/spreadsheets/d/1eiR9riut62MrTvUR-qlHCfSJw0lGZ9Sg0pPhqOBPEsU/edit#gid=2128570643


bachstatter (approved):

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -93,33 +94,40 @@ library PerpsMarket {
...
+        Data storage self,
+        uint256 requestedLiquidationAmount
+    ) internal returns (uint256 liquidatableAmount) {
+        PerpsMarketConfiguration.Data storage marketConfig = PerpsMarketConfiguration.load(self.id);
+
+        uint maxLiquidationAmountPerSecond = marketConfig.maxLiquidationPerSecond();
         uint timeSinceLastUpdate = block.timestamp - self.lastTimeLiquidationCapacityUpdated;
+        uint maxSecondsInLiquidationWindow = marketConfig.maxSecondsInLiquidationWindow;
 
         self.lastTimeLiquidationCapacityUpdated = block.timestamp.to128();
```
leomassazza, line 114: You need to save the latest time ONLY if liquidatebleAmount is > 0. Otherwise it will fail in the following scenario (please, add a test case for that scenario):

When there is a series of liquidations every X time, where X is less than `maxSecondsInLiquidationWindow`. 
It will fail to process liquidations after `maxAllowedLiquidationInWindow` is used (due to the else in the if), and will only be able to process liquidations again after a pause of at least `maxSecondsInLiquidationWindow` (when it can enter the first branch of the if)

leomassazza (changes requested):

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -344,113 +347,137 @@ library PerpsAccount {
...
-        // -- in loop
-        /*
+            // loop over losing positions,
+            // -- in loop
+            /*
             1. get the loss amount
             2. loss = find the min of loss, max liquidatable
             3. % = (loss * pnl) / all losing positions pnl
             4. colateral balance * % = value to deposit for that market
             5. withdraw from market keeper fee based on pnl that was liquidated
```
bachstatter, line 383: This comment is copied from losing pos loop and can be removed

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -344,113 +347,137 @@ library PerpsAccount {
...
+                (
+                    uint amountToLiquidate,
+                    int totalPnl,
+                    uint liquidationReward,
+                    int128 oldPositionSize
+                ) = _processMarketLiquidation(self, positionMarketId);
+
+                if (self.openPositionMarketIds.contains(positionMarketId)) {
+                    // partial liquidation
+                    fullLiquidation = false;
```
bachstatter, line 398: I dont think we need this if statement when going over losing positions?

bachstatter (approved):
Nice! I tried to go thorugh it in more detail this time and it make sense to me. Added one commit with some comment updates.

We need some more liquidation tests in general, but your test covers max liquidatable amount quite well!

PR #1677, opened 2023-06-21
Prepare, test, build and deploy scripts for main-subgraph

noisekit:
- Follow up to #1676 
- Added info on how to update, build and deploy to the README

### DIFF
--- a/protocol/synthetix/subgraph/README.md
+++ b/protocol/synthetix/subgraph/README.md
@@ -23,14 +23,7 @@ yarn codegen
 ## Testing the subgraph
 
 ```bash
-# 1. generate the Subgraph files (we are relying on CoreProxy only for optimism-goerli)
-yarn build:optimism-goerli
-
-# 2. prepare subgraph.yaml (only once)
-yarn prepare
-
-# 3. Run subgraph test with matchstick
-yarn graph test
+yarn test
 ```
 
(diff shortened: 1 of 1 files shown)


protocol/synthetix/subgraph/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```bash
+# 1. generate the Subgraph files (we are relying on CoreProxy only for optimism-goerli)
+yarn build:optimism-goerli
+
+# 2. prepare subgraph.yaml (only once)
+yarn prepare
+
+# 3. Run subgraph test with matchstick
+yarn graph test
```
bachstatter, line 33: `yarn test` works without any steps. It calls `yarn prepare` and tests (matchstick) always does a fresh build before running the tests

protocol/synthetix/subgraph/matchstick.yaml
```diff
@@ -1,2 +1,2 @@
 testFolder: tests/
-libsFolder: ../node_modules/
+libsFolder: ../../../node_modules/
```
bachstatter, line 2: If we add this:
manifestPath: ./subgraph.base.yaml

We can change the `test` npm script to be just `graph test`

bachstatter (approved):
Some minor suggestion, but looking good

PR #1678, opened 2023-06-22
Hotfix tests

noisekit:
Extracted few commits from #1666 to get the tests running in master

- fix the solc version
- update cannon and hardhat deps
- fix reference of router util in tests
- go back to node 16 temporarily
- provision -> import for tests


bachstatter (approved):
Finally green again ✅

PR #1680, opened 2023-06-22
More deposit failures

bachstatter:
Currently we have three sligtly different expected failures that all reverts with the same message `MaxCollateralExceeded`.

1. Trying to deposit a synth id that doesn't exist
2. Trying to deposit not enabled synth (max set to 0) 
3. Trying to deposit more than Max

It does sorta make sense that two and three reverts with the same message, but it would be clearer if we had more specific error messages.
Want me to add that? Or is it fine to reuse `MaxCollateralExceeded` since it's not "lying" per say


leomassazza:
Like your idea of adding a different error

leomassazza (approved):
LGTM, but I like the idea of having different errors for the different situations.

sunnyvempati:
@bachstatter another way to accomplish this would be creating `MaxCollateralExceeded(bytes message)` and then just return a string literal.  but I'm a fan of separate errors also!  just my 2️⃣ cents

PR #1684, opened 2023-06-22
Add test to verify fees paid

leomassazza:


markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -260,16 +260,20 @@ library AsyncOrder {
...
-        uint makerFee = MathUtil.abs(notionalDiff).mulDecimal(makerSize).mulDecimal(
+        // The proportions are computed as follows:
+        // makerSize = abs(marketSkew) => since we are reversing the skew, the maker size is the current skew
+        // takerSize = abs(marketSkew + sizeDelta) => since we are reversing the skew, the taker size is the new skew
+        //
+        // we then multiply the sizes by the fill price to get the notional value of each side, and that times the fee rate for each side
+
+        uint makerFee = MathUtil.abs(marketSkew).mulDecimal(fillPrice).mulDecimal(
             orderFeeData.makerFee
         );
```
leomassazza, line 272: Changed how we calculate the maker and taker sides since previous version had a rounding error at some proportions. i.e. with 
```
sizeDelta = -3 marketSkew
that means makerSide is 1/3 and takerSide is 2/3 that rounded at 
makerSide = 333333333333333333
takerSide = 666666666666666667 
```

markets/perps-market/contracts/storage/PerpsMarketConfiguration.sol
```diff
@@ -74,7 +74,7 @@ library PerpsMarketConfiguration {
         )
     {
         uint256 sizeAbs = MathUtil.abs(size.to256());
-        uint256 impactOnSkew = sizeAbs.divDecimal(self.skewScale);
+        uint256 impactOnSkew = self.skewScale == 0 ? 0 : sizeAbs.divDecimal(self.skewScale);
```
leomassazza, line 77: safeguard for skewScale = 0

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -280,6 +284,9 @@ library AsyncOrder {
         int size,
         uint price
     ) internal pure returns (uint) {
+        if (skewScale == 0) {
```
leomassazza, line 287: safeguard for skewScale = 0

markets/perps-market/test/integration/bootstrap/bootstrap.ts
```diff
@@ -144,3 +144,5 @@ export function bootstrapMarkets(data: BootstrapArgs) {
 
 export const bn = (n: number) => wei(n).toBN();
 export const toNum = (n: ethers.BigNumber) => Number(ethers.utils.formatEther(n));
+export const decimalMul = (a: ethers.BigNumber, b: ethers.BigNumber) => a.mul(b).div(bn(1));
```
leomassazza, line 147: some new helpers

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.fees.test.ts
```diff
@@ -1,6 +1,420 @@
...
+  let btcSynth: SynthMarkets[number];
+
+  before('identify actors', async () => {
+    ethMarketId = perpsMarkets()[0].marketId();
+    btcSynth = synthMarkets()[0];
+  });
+
+  const restoreToDepositCollateral = snapshotCheckpoint(provider);
+
+  const testCases: Array<{ name: string; collateralData: DepositCollateralData }> = [
```
sunnyvempati, line 54: the type here will generally be implied based on the data so no need to explicitly set it, but either way non issue, mostly fyi :)

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.fees.test.ts
```diff
@@ -1,6 +1,420 @@
...
+          {
+            synthMarket: () => btcSynth,
+            snxUSDAmount: () => bn(10_000),
+          },
+        ],
+      },
+    },
+  ];
+
+  for (let idx = 0; idx < testCases.length; idx++) {
```
sunnyvempati, line 101: fyi, i've been using forEach with describe and seems to work fine

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.fees.test.ts
```diff
@@ -1,6 +1,420 @@
...
+    const notionalMaker = decimalMul(makerSize, price);
+
+    const perpsMarketFee = decimalMul(notionalMaker, orderFees.makerFee).add(
+      decimalMul(notionalTaker, orderFees.takerFee)
+    );
+    const keeperFee = DEFAULT_SETTLEMENT_STRATEGY.settlementReward;
+
+    return { totalFees: perpsMarketFee.add(keeperFee), perpsMarketFee, keeperFee };
+  };
+});
```
sunnyvempati, line 420: awesome tests 👍🏽

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.fees.test.ts
```diff
@@ -1,6 +1,420 @@
...
+      keeperBalance,
+      accountPnl,
+    };
+  };
+
+  const computeFees: (
+    sizeBefore: ethers.BigNumber,
+    sizeDelta: ethers.BigNumber,
+    price: ethers.BigNumber
+  ) => {
```
sunnyvempati, line 391: generally, ive been trying not to write functions that mimic contract that calculate the value we're wanting to assert.  if there's a bug in the contract and the function mimics what hte contract is doing, then teh bug will be hidden.  i've resorted to having excel sheet with what i think the calc is, and then having afif and all of you checking the logic to make sure it all looks right.  and if those values match up with what the contract is returning then i feel confident.

so just to wary of that; not saying this function isn't returning the right value but yeah hope that makes sense 👍🏽

sunnyvempati (approved):
lgtm!

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.fees.test.ts
leomassazza: tbh, I prefer our code to have the values of the tests calculated in place, otherwise a minimun change in the behavior will need to change all hardcoded values. Also, if somebody doesn't have access to the sheet, it cannot verify the values makes sense. 
In this case I went the other way round. I did the sheet, the test helper and checked the values made sense. In fact with that function is where I found we had a rounding issue on the calculations and made the function kind of mimic the code of the test.

sunnyvempati: yea fair points.  the sheets is an experiement for sure and not scalable.  you might have convinced me to have _some_ form of code to determine values.   i'll keep iterating; thanks

bachstatter (approved):
Nice, looks good to me!

Regarding functions calculating things in test, I would normally see that as a sign to restructure tests. To be fair I would think that's the case when I see loops in tests too.
But given that we dont have easy way to mock/ dependency inject I'm not sure if those thoughts are valid here.. Still quite new to this!

PR #1687, opened 2023-06-26
Position min margin

bachstatter:
This a new liquidation parameter `minimumPositionMargin`, this is a constant value in USD that can control a minimum amount of required collateral that's *NOT* based on positions size

### DIFF
--- a/markets/perps-market/hardhat.config.ts
+++ b/markets/perps-market/hardhat.config.ts
@@ -21,7 +21,7 @@ const config = {
     templates,
   },
   mocha: {
-    timeout: 120000,
+    timeout: 200000,
   },
 };
 
--- a/markets/perps-market/hardhat.config.ts
+++ b/markets/perps-market/hardhat.config.ts
@@ -21,7 +21,7 @@ const config = {
     templates,
   },
   mocha: {
-    timeout: 200000,
+    timeout: 120000,
   },
 };
 
--- a/markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
+++ b/markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
@@ -2,6 +2,7 @@ import assertBn from '@synthetixio/core-utils/utils/assertions/assert-bignumber'
 import assertRevert from '@synthetixio/core-utils/utils/assertions/assert-revert';
 import { bn, bootstrapMarkets } from '../bootstrap';
 import { OpenPositionData, openPosition } from '../helpers';
+import { snapshotCheckpoint } from '@synthetixio/core-utils/utils/mocha/snapshot';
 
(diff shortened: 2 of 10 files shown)


markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
```diff
@@ -208,14 +214,84 @@ describe('Liquidation - margin', async () => {
...
+          maintenanceMarginFraction,
+          maxLiquidationLimitAccumulationMultiplier,
+          liquidationRewardRatio,
+          maxSecondsInLiquidationWindow,
+          minimumPositionMargin
+        );
+    });
+    // Changing minimumPositionMargin does not have an affect on available margin
+    it('has correct available margin', async () => {
+      assertBn.equal(await systems().PerpsMarket.getAvailableMargin(2), bn(0));
```
bachstatter, line 274: We have this `getAvailableMargin`, that just looks at the total account collateral value + the account PNL.
So it's missing the liquidations considerations, eg most of the time I will not be able to withdraw what it returns.. 
Might be worth calling this something else, and make `getAvailableMargin` take liquidations concerns into account?

markets/perps-market/hardhat.config.ts
```diff
@@ -21,7 +21,7 @@ const config = {
     templates,
   },
   mocha: {
-    timeout: 200000,
+    timeout: 120000,
```
sunnyvempati, line 24: lol

markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
sunnyvempati: the `getAvailableMargin` is what gets checked against your maintenance margin for liquidations.  its not meant to be a value that is withdrawable.  its something to display on the UI to check your margin against your pnls.  maybe a `getWithdrawableAmount` which subtracts out the maintenace margin could be nice to add

sunnyvempati (approved):
lgtm!

markets/perps-market/hardhat.config.ts
bachstatter: haha wait, I thought I increased it...Will revert this

markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
bachstatter: > the `getAvailableMargin` is what gets checked against your maintenance margin for liquidations. its not meant to be a value that is withdrawable. its something to display on the UI to check your margin against your pnls. maybe a `getWithdrawableAmount` which subtracts out the maintenace margin could be nice to add

Yea that what I want. And I like that name too!

PR #1688, opened 2023-06-26
RatioD18

fritzschoff:


fritzschoff:
> I think the ticket wasn't detailed enough... the intention was to get rid of `percentage` and move to `ratio` while also identifying how it should be interpreted `D18`, so i.e. `lockedOiPercent` should move to `lockedOiRatioD18` (or just `lockedOiRatio` for public params)

thanks. just chatted with sunny. We will chat after the v3 standies

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -400,19 +400,19 @@ library PerpsAccount {
...
                     runtime.totalLosingPnl
                 );
                 runtime.totalAvailableForDeposit = totalAvailableUsd.mulDecimal(
-                    runtime.percentageOfTotalLosingPnl
+                    runtime.totalLosingPnlRatioD18
                 );
 
                 runtime.amountToDeposit = runtime.totalAvailableForDeposit.mulDecimal(
-                    runtime.amountToLiquidatePercentage
+                    runtime.amountToLiquidateRatioD18
```
sunnyvempati, line 415: ignore liquidation ones, they are all about to change.  ty

markets/perps-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -72,36 +72,36 @@ contract MarketConfigurationModule is IMarketConfigurationModule {
...
+            initialMarginRatioD18,
+            maintenanceMarginRatioD18,
             liquidationRewardRatioD18,
             maxLiquidationLimitAccumulationMultiplier,
             maxSecondsInLiquidationWindow
         );
     }
 
-    function setLockedOiPercent(uint128 marketId, uint256 lockedOiPercent) external override {
+    function setLockedOiRatioD18(uint128 marketId, uint256 lockedOiRatioD18) external override {
```
sunnyvempati, line 100: no need to change name of setter.  it can just stay `setLockedOiRatio`.  the `D18` append is helpful mostly for devs looking at the code; and for auditors

sunnyvempati: also helpful in interfaces where external integrators can see what value type it is

sunnyvempati (approved):
pre approving but have a couple comments

markets/perps-market/contracts/storage/PerpsAccount.sol
fritzschoff: if they gonna change than I can leave it anyway?

PR #1689, opened 2023-06-27
Fill price

bachstatter:
We had a refactor comment for the fill price fn.. This is how I would do it, but unsure if it's worth it consider it might incur more gas costs? Either way the tests should be good, but can revert the refactor commit based on feedback.


bachstatter:
Realised that this task was moved to the backlog.. oh well

markets/perps-market/test/integration/Market/PerpsMarketModule.test.ts
```diff
@@ -53,14 +54,88 @@ describe('PerpsMarketModule', () => {
...
+      {
+        marketSkew: -10,
+        sizeAndExpectedPrice: [
+          { size: 1, price: bn(1010), expectedPrice: bn(1009.0405) },
+          { size: -1, price: bn(1010), expectedPrice: bn(1008.9395) },
+          { size: 11, price: bn(1010), expectedPrice: bn(1009.5455) },
+        ],
+      },
+    ];
+    before('add collateral', async () => {
```
leomassazza, line 116: not an issue here, but take into account that this `before` will be executed before the `describe "skewScale 0"`

leomassazza (approved):
Pre-approved. LGTM. Agree on your comment on the need of the refactor. It now looks better on readability, but if this change incurs more gas, we are asking people to pay for our code readability... not sure if we want that.

markets/perps-market/test/integration/Market/PerpsMarketModule.test.ts
bachstatter: Ah yea thanks, I'll just move it to make that clear

PR #1692, opened 2023-06-28
Cleanup docgen

noisekit:
- Remove leftover docs from the previous generator
- ~Update ABIs~ Remove ABIs (as docs are generated by hardhat from actual contracts sources)
- Fix clean testable
- Update docs from fresh contracts


bachstatter (approved):
I dont have the most context around docs, but looks good to me.

PR #1696, opened 2023-06-29
Update `OrderSettled` event

Tburm:
* Add `sizeDelta` to the `OrderSettled` event - This will help calculate volume (`sizeDelta * fillPrice`)
* Fix a typo in the `OrderSettled` event
* Renames a variable - A variable named `asyncOrderId` actually represents the `accountId`
* Updates all docs using docgen

### DIFF
--- a/markets/perps-market/contracts/interfaces/IAsyncOrderModule.sol
+++ b/markets/perps-market/contracts/interfaces/IAsyncOrderModule.sol
@@ -28,7 +28,7 @@ interface IAsyncOrderModule {
         int128 sizeDelta,
         int128 newSize,
         uint256 collectedFees,
-        uint256 settelementReward,
+        uint256 settlementReward,
         bytes32 indexed trackingCode,
         address settler
     );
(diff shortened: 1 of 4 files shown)


bachstatter:
Sorry I missed this draft. We might get some conflicts https://github.com/Synthetixio/synthetix-v3/pull/1697

Tburm:
@bachstatter All good, I'm going to leave a comment on your PR actually. I was getting new "stack too deep" errors so I think new values need to go in the runtime.

bachstatter:
> @bachstatter All good, I'm going to leave a comment on your PR actually. I was getting new "stack too deep" errors so I think new values need to go in the runtime.

Yea you can see how I worked around it by moving the `trackingCode` to `runtime`. Solidity wont allow you to move `PerpsMarket` to `runtime` since `runtime` is memory and the `market` will be storage.. Luckily we could move the trackingCode and "afford" to have market as a variable

bachstatter:
We probably want some of these changes.. I added size delta to the new MarketUpdated event. Do you still want a size delta on order settled?

Tburm:
@bachstatter Yes, we definitely want `sizeDelta` on `OrderSettled`. It will be easiest to calculate volume if we can just do `sizeDelta * price` from this event.

PR #1697, opened 2023-06-30
MarketUpdated event

bachstatter:
Include market skew and market size in the settle order event


markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -236,7 +237,9 @@ contract AsyncOrderModule is IAsyncOrderModule {
             runtime.newPositionSize,
             totalFees,
             runtime.settlementReward,
-            trackingCode,
+            market.size,
+            market.skew,
```
Tburm, line 241: @bachstatter My original idea was to add these to this event, but I'm wondering if a cleaner solution is creating a new event for `MarketStatus`. In V2 we had to fetch this data using `eth_call` because it wasn't in an event (`marketDebt` in V2)

leomassazza: kind of like your idea. We can add the event where we update those parameters (https://github.com/Synthetixio/synthetix-v3/blob/9fa0cdef61ee644d17c14d7d90ee9dab8a7b2ac8/markets/perps-market/contracts/storage/PerpsMarket.sol#L146) and will be available anytime we update it.

bachstatter: > In V2 we had to fetch this data using eth_call because it wasn't in an event (marketDebt in V2)

What would you put in `MarketStatus`, that's not in the settle event? Defintaly want to figure out away so you can avoid eth_call

> kind of like your idea. We can add the event where we update those parameters and will be available anytime we update it.

`oldPosition.updatePosition(newPosition);` is only called from settle, but I guess it make sense to put it there

bachstatter: > `oldPosition.updatePosition(newPosition);` is only called from settle

Actually this might be a bug, do we miss updating market.size/ market.skew on liquidations?

Tburm: @bachstatter
Definitely: `marketId`, `marketSize`, `marketSkew`, `currentFundingRate`
Potentially:  `currentFundingVelocity`, market price

In v2 I would have wanted `marketDebt` to avoid the eth_calls, but there isn't a similar value for v3. I'm still not sure how to measure the unrealized pnl of existing positions to help calculate staker pnl.

bachstatter:
A bit of a journey but I think I got there in the end.
Added `event MarketUpdated(uint128 marketId, int256 skew, uint256 size, int256 sizeDelta);`

Learned that if you want to emit an event from a storage library you need to define it:
1. In the library
2. In the contract using it. (or the contract's interface)

Doing this I hit a pushed the AsyncOrderModule over the size limit, so I created a new `AsyncOrderViewModule`

leomassazza (approved):
LGTM

markets/perps-market/contracts/modules/AsyncOrderViewModule.sol
```diff
@@ -0,0 +1,17 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.11 <0.9.0;
+
+import {IAsyncOrderViewModule} from "../interfaces/IAsyncOrderViewModule.sol";
+import {AsyncOrder} from "../storage/AsyncOrder.sol";
+import {PerpsMarket} from "../storage/PerpsMarket.sol";
+
+contract AsyncOrderViewModule is IAsyncOrderViewModule {
```
sunnyvempati, line 8: thoughts on creating an `AsyncOrderSettlementModule` instead?  Would be similar to what we have in spot market.  

can move settle related items into the above module.

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -30,6 +30,15 @@ library PerpsMarket {
 
     error PriceFeedNotSet(uint128 marketId);
 
+    event MarketUpdated(
```
sunnyvempati, line 33: would not recommend adding events to libraries.  libraries are more reusable code and should be used in conjunection with a contract.  you can then define your event and emit it within contract.

you'll also notice there's no events in any other storage libraries for reason above.

sunnyvempati: my recommendation would be to create a new interface called: `IMarketEvents.sol`:
```
interface IMarketEvents {
    event MarketUpdated(...);
}


```

Then in `AsyncOrderModule`, you can inteherit from this interface like so:
```
contract AsyncOrderModule is IAsyncOrderModule, IMarketEvents 
```

and now you can do:

```
emit MarketUpdated(...);
```

This way in liquidations, I can also inherit from `IMarketEvents` and emit the same event

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -144,6 +153,16 @@ library PerpsMarket {
...
+        int128 sizeDelta = newPosition.size - oldPositionSize;
+        // TODO add current market debt
+        emit MarketUpdated(
+            self.id,
+            self.skew,
+            self.size,
+            sizeDelta,
+            currentFundingRate(self),
+            currentFundingVelocity(self)
+        );
```
sunnyvempati, line 165: maybe return this data as a tuple then emit the MarketUpdated event in async order settlement

markets/perps-market/test/integration/helpers/funding-calcs.ts
```diff
@@ -0,0 +1,22 @@
...
+  // Avoid a panic due to div by zero. Return 0 immediately.
+  if (skewScale.eq(0)) {
+    return wei(0);
+  }
+
+  // Ensures the proportionalSkew is between -1 and 1.
+  const pSkew = wei(skew).div(skewScale);
+  const pSkewBounded = Wei.max(Wei.min(pSkew, wei(1)), wei(-1));
+
+  return pSkewBounded.mul(maxFundingVelocity);
```
sunnyvempati, line 21: maybe returns bignumber via `.toBN()` here?

sunnyvempati (changes requested):

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -144,6 +153,16 @@ library PerpsMarket {
...
         oldPosition.updatePosition(newPosition);
+        int128 sizeDelta = newPosition.size - oldPositionSize;
+        // TODO add current market debt
+        emit MarketUpdated(
+            self.id,
+            self.skew,
+            self.size,
+            sizeDelta,
+            currentFundingRate(self),
+            currentFundingVelocity(self)
```
sunnyvempati, line 164: randomly thought of this but `currentFundingRate` also calls `currentFundingVelocity` in its calculation so we're doing double the calcs here.

You can probably just use `self.lastFundingRate` for `currentFundingRate` since within the same transaction it was updated to the current block timestamp, the velocity won't have any impact

markets/perps-market/contracts/modules/AsyncOrderViewModule.sol
bachstatter: Happy to align with spot market!

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: My idea was that instead of having to remember to `emit MarketUpdates` in both the order and the liquidation module they both can now call ``Perps.updatePositionData` and the event is emitted for them.

But agree that is' a little weird to emit event from the library, so if you think manually emitting the same event twice is cleaner I'm happy to move to a IMarketEvents architecture

markets/perps-market/test/integration/helpers/funding-calcs.ts
bachstatter: Coming from the frontend we usually keep everything in Wei and just `toBN` before sending to ethers contract call. Wei usually makes math and most things easier. 
But we do seem to like/use BigNumber more in contract tests, so I guess it make sense to return bn.

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: Yep that make sense if we move to `IMarketEvents`

markets/perps-market/contracts/storage/PerpsMarket.sol
sunnyvempati: yeah understood.  i think its ok to return the relevant data to the caller and let the module decide if it wants to emit an event. thoughts?

markets/perps-market/test/integration/helpers/funding-calcs.ts
sunnyvempati: ok cool yeah fair enough!

markets/perps-market/contracts/modules/AsyncOrderModule.sol
bachstatter: Added

markets/perps-market/contracts/modules/AsyncOrderViewModule.sol
bachstatter: Done

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: Yep, will do. Just explained my thinking

bachstatter: I would probably leave a comment that caller should emit this though. If you call this and forget emit an event offchain/subgraph tracking will be wrong

sunnyvempati: yeah, nothing contract side breaks but yeah, in general, we'll audit all events together with troy at offsite maybe

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: Fixed

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: Done

markets/perps-market/contracts/interfaces/IMarketEvents.sol
```diff
@@ -0,0 +1,13 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.11 <0.9.0;
+
+interface IMarketEvents {
```
sunnyvempati, line 4: any others around that we can throw in here maybe? 👀

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -131,6 +131,7 @@ library AsyncOrder {
         uint initialRequiredMargin;
         uint totalRequiredMargin;
         Position.Data newPosition;
+        bytes32 trackingCode;
```
sunnyvempati, line 134: still need this?

sunnyvempati (approved):
awesome! tyty

sunnyvempati:
hmm some tests failed; seems legitimate?

markets/perps-market/contracts/storage/AsyncOrder.sol
bachstatter: Yea, without it we get "Stack too deep." Just at the limit now 😓

bachstatter:
> hmm some tests failed; seems legitimate?

It does looks like that.. Looking now, pretty sure I ran it locally...

bachstatter:
> > hmm some tests failed; seems legitimate?
> 
> It does looks like that.. Looking now, pretty sure I ran it locally...

Was a fluke :(... Reran fixed it

markets/perps-market/contracts/interfaces/IMarketEvents.sol
bachstatter: Most modules will have their events defined in their own interface.  I had a look, didn't see any clear ones to moce

PR #1698, opened 2023-06-30
Fill price js impl

bachstatter:
I had a convo with david about how hard it is to debug tests with a bunch of hardcoded numbers..
And then I saw a TODO about fill price, so went ahead an implemented it.

Do we want to use it in the fillPrice tests? It does feel a little wrong to me... Should I create a spreadsheet?

### DIFF
--- a/markets/perps-market/test/integration/helpers/fillPrice.ts
+++ b/markets/perps-market/test/integration/helpers/fillPrice.ts
@@ -1,0 +1,19 @@
+import Wei, { wei } from '@synthetixio/wei';
+
+// Calculates PD
+const calculatePD = (skew: Wei, skewScale: Wei) => wei(skew).div(wei(skewScale));
+// Calculates the price with pd applied
+const calculateAdjustedPrice = (price: Wei, pd: Wei) => wei(price).add(wei(price).mul(pd));
+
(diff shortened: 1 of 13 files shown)


leomassazza:
Agree we shouldn't use hardcoded values. Ideally, we should add a test for the test helper. It can be done with plain numbers, but the helper will be useful on plenty of other tests where we have a derived number (fillPrice, fees, pnl, etc.)

leomassazza (approved):

markets/perps-market/test/integration/helpers/fillPrice.ts
```diff
@@ -0,0 +1,19 @@
+import Wei, { wei } from '@synthetixio/wei';
+
+// Calculates PD
+const calculatePD = (skew: Wei, skewScale: Wei) => wei(skew).div(wei(skewScale));
+// Calculates the price with pd applied
+const calculateAdjustedPrice = (price: Wei, pd: Wei) => wei(price).add(wei(price).mul(pd));
```
sunnyvempati, line 6: price is already wei so can prolly remove the extra one

bachstatter: Thanks!

PR #1700, opened 2023-07-03
Market debt

leomassazza:


markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -101,11 +107,33 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
...
+            uint256 activeMarketsLength = activeMarkets.length();
+            for (uint i = 1; i < activeMarketsLength; i++) {
+                uint128 marketId = activeMarkets.valueAt(i).to128();
+                totalDebt += PerpsMarket.load(marketId).marketDebt(
+                    PerpsPrice.getCurrentPrice(marketId)
+                );
+            }
+
+            // TODO Should revert if totalDebt is positive and larger than totalCollateralValue???
+            return MathUtil.abs(totalCollateralValue.toInt() - totalDebt);
```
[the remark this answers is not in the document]
sunnyvempati: also shouldn't it be `totalCollateralValue + totalDebt`?
```
example 1:
total collateral: $1000 
total markets debt: $500

reported Debt == $1500

example 2:
total collateral value = $1000
total market debt = -$800

reported debt = $200
```

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -141,8 +143,20 @@ library PerpsMarket {
...
 
+        // update the market debt correction accumulator before losing oldPosition details
+        // by adding the new updated notional (old - new size) plus old position pnl
+        // debtCorrectionAccumulator += newPosition.size * currentPrice.toInt() - oldPosition.size * currentPrice.toInt() + oldPositionPnl ;
+        uint currentPrice = newPosition.latestInteractionPrice;
+        (int oldPositionPnl, , , ) = oldPosition.getPnl(currentPrice);
+
+        self.debtCorrectionAccumulator +=
+            currentPrice.toInt() *
+            (newPosition.size - oldPositionSize) -
```
sunnyvempati, line 157: looks like this is `+` oldPositionPnl in the sheet?  i could be wrong but just double check

sunnyvempati (changes requested):

markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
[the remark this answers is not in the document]
leomassazza: Need to double check together with the `+` or `-` on `oldPositionPnl` since it depends on how marketDebt is seen (if positive or negative)

markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -36,7 +36,7 @@ contract LiquidationModule is ILiquidationModule {
             .load()
             .liquidatableAccounts;
 
-        for (uint i = 0; i < liquidatableAccounts.length(); i++) {
+        for (uint i = 1; i <= liquidatableAccounts.length(); i++) {
```
sunnyvempati, line 39: fixed this in my PR; we actually can't use `liquidateAccounts` directly because if there's a full account liquidation, it gets removed and we get an error.  instead i move the liquidate accounts values into an array in memory which solves the issue 👍🏽

sunnyvempati: feel free to keep this; and i'll fix merge conflict or vice versa depending on which PR goes in first

markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -101,11 +108,37 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
...
+                    PerpsPrice.getCurrentPrice(marketId)
+                );
+            }
+
+            int reportedDebt = totalCollateralValue.toInt() + totalDebt;
+            if (reportedDebt < 0) {
+                return 0;
+            } else {
+                return reportedDebt.toUint();
+            }
```
sunnyvempati, line 138:
```suggestion
            return reportedDebt < 0 ? 0 : reportedDebt.toUint();
```

nonblocking

markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -101,11 +108,37 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
...
+        }
+
+        // TODO Should revert if perpsMarketId is not correct???
```
sunnyvempati, line 141: 0 is fine.  we probably should never revert on reported debt since it could brick the core system

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -141,8 +143,19 @@ library PerpsMarket {
...
 
+        // update the market debt correction accumulator before losing oldPosition details
+        // by adding the new updated notional (old - new size) plus old position pnl
+        // debtCorrectionAccumulator += newPosition.size * currentPrice.toInt() - oldPosition.size * currentPrice.toInt() + oldPositionPnl ;
+        uint currentPrice = newPosition.latestInteractionPrice;
+        (int oldPositionPnl, , , ) = oldPosition.getPnl(currentPrice);
+
+        self.debtCorrectionAccumulator +=
+            currentPrice.toInt().mulDecimal(newPosition.size - oldPositionSize) +
+            oldPositionPnl;
```
sunnyvempati, line 157: do we need to do anything with accumulator on liquidiations?  i dont think so since it becomes realized right?  i'll think through this a bit more

sunnyvempati (approved):
some comments but lgtm!

markets/perps-market/contracts/modules/LiquidationModule.sol
leomassazza: I'll keep it otherwise tests will fail, but will fix if your PR comes first

markets/perps-market/contracts/storage/PerpsMarket.sol
leomassazza: most likely yes. Maybe the numbers get balanced together, but not completely sure on some corner cases

sunnyvempati, line 157:
```suggestion
        uint currentPrice = newPosition.latestInteractionPrice;
        (int oldPositionPnl, int accruedFunding, , ) = oldPosition.getPnl(currentPrice);

        self.debtCorrectionAccumulator +=
            currentPrice.toInt().mulDecimal(newPosition.size - oldPositionSize) +
            oldPositionPnl - accruedFunding;  // don't include accrued funding in this
```

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -280,4 +293,15 @@ library PerpsMarket {
...
+    /**
+     * @dev Returns the market debt incurred by all positions
+     * @notice  Market debt is the sum of all position sizes multiplied by the price, and old positions pnl that is included in the debt correction accumulator.
+     */
+    function marketDebt(Data storage self, uint price) internal view returns (int) {
+        // all positions sizes multiplied by the price is equivalent to skew times price
+        // and the debt correction accumulator is the sum of all positions pnl
+
+        return self.skew.mulDecimal(price.toInt()) - self.debtCorrectionAccumulator;
+    }
```
sunnyvempati, line 306:
```suggestion
    function marketDebt(Data storage self, uint price) internal view returns (int) {
        // all positions sizes multiplied by the price is equivalent to skew times price
        // and the debt correction accumulator is the sum of all positions pnl
        int traderUnrealizedPnl = self.skew.mulDecimal(price.toInt()) - self.debtCorrectionAccumulator;
        int totalAccruedFunding = self.skew.mulDecimal(price.toInt()).mulDecimal(currentFundingRate();

        return traderUnrealizedPnl + totalAccruedFunding;
    }
```

sunnyvempati: this will take care of unrealized funding.  if we subtract it out of the accumulator then it would be counted double, so we only use trader pnl in the accumulator.

sunnyvempati (changes requested):
hey, suggested some changes.  verified on the google sheet, but this makes the most sense to me.  let me know your thoughts

sunnyvempati (approved):

PR #1702, opened 2023-07-03
Docs - some interfaces

leomassazza:


markets/perps-market/contracts/interfaces/IMarketConfigurationModule.sol
```diff
@@ -91,15 +193,37 @@ interface IMarketConfigurationModule {
...
     function getFundingParameters(
         uint128 marketId
     ) external view returns (uint256 skewScale, uint256 maxFundingVelocity);
 
+    /**
+     * @notice Gets the max market value of a market.
+     * @param marketId id of the market.
+     * @return maxMarketValue the max market value.
+     */
     function getMaxMarketValue(uint128 marketId) external view returns (uint256 maxMarketValue);
```
bachstatter, line 211: Maybe mention: 
Max Market Value controls the maximum open interest a market an have on either side (Long | Short). So the total Open Interest for a market is Max Market Value * 2. This value is not defined in USD but in the native asset

markets/perps-market/contracts/interfaces/IAsyncOrderModule.sol
```diff
@@ -49,12 +91,51 @@ interface IAsyncOrderModule {
...
+     * @param accountId Id of the account used for the trade.
+     */
     function cancelOrder(uint128 marketId, uint128 accountId) external;
 
+    /**
+     * @notice Settles an offchain order. It's expected to revert with the OffchainLookup error with the data needed to perform the offchain lookup.
+     * @param marketId Id of the market used for the trade.
+     * @param accountId Id of the account used for the trade.
+     */
+    function settle(uint128 marketId, uint128 accountId) external view;
```
bachstatter, line 116: I've added these too in the new AsyncOrderSettlementModule (but currently without comment). No biggie but might have to rebase

bachstatter (approved):
One suggestion to be a bit more elaborate on maxMarketValue. Not sure if we want to avoid that but I just know that it's common to get question about it

PR #1705, opened 2023-07-04
Accrued funding test refactor

sunnyvempati:
- removed hard coded values and added the calc


markets/perps-market/test/integration/Position/Funding.test.ts
```diff
@@ -64,48 +70,82 @@ describe('Position - funding', () => {
...
+      { prevFrVelocity, prevFundingRate, prevAccruedFunding, prevSkew, accDaysElapsed },
+      { daysElapsed, newOrderSize }
+    ) => {
+      accDaysElapsed += daysElapsed;
+
+      // ACCRUED FUNDING CALC ----------------------------------------------
+      const currentSkew = prevSkew.add(newOrderSize);
+      const frVelocity = wei(currentSkew).div(_SKEW_SCALE).mul(_MAX_FUNDING_VELOCITY);
+      const fundingRate = prevFrVelocity.mul(daysElapsed).add(prevFundingRate);
+      const expectedAccruedFunding = Wei.avg(wei(prevFundingRate), wei(fundingRate))
```
bachstatter, line 93: I added `calcCurrentFundingVelocity` in a `helpers/funding-calcs.ts` that you could use. We could also move some of these there if we think other tests will assert funding updates

bachstatter (approved):

markets/perps-market/test/integration/Position/Funding.test.ts
leomassazza: +1

leomassazza (approved):

PR #1706, opened 2023-07-05
Update `OrderSettled` event

Tburm:
Adds `sizeDelta` to the `OrderSettled` event, and removes `pnl` from that event.

* `sizeDelta` is critical to send when an order is settled. We currently log the new position size, but it makes sense to add the order size to this event as well
* `pnl` in this case refers to a change in the trader's margin. The realized pnl displayed by integrators uses a different method, only realizing pnl when the trader is reducing their position size or paying fees. Since the relevant information is not stored, it's best not to be opinionated about pnl in this event and allow offchain indexers to implement the logic
* There are some other changes to the runtime and variables in memory where these were already available. Notably the `trackingCode` is not cleared on `asyncOrder.reset()` so we do not need to store it in the runtime.


bachstatter (approved):
Nice!

PR #1707, opened 2023-07-06
Remove duplicated getOrder

bachstatter:
We had `getAsyncOrderClaim` and `getOrder` which were doing the same thing.
I think it makes a bit more sense to have it live in the OrderModule, so opted to remove `getAsyncOrderClaim` from account


bachstatter:
Noticed that spot does have `getAsyncOrderClaim`, not sure I fully under stand why it's a "claim" though?

leomassazza (approved):

PR #1708, opened 2023-07-06
Add price to `MarketUpdated` event

Tburm:
* Adds latest fill price to `MarketUpdated` event: This enables USD denominated OI calculations from the events
* Adds natspec docs to event


markets/perps-market/contracts/modules/AsyncOrderSettlementModule.sol
```diff
@@ -157,6 +157,7 @@ contract AsyncOrderSettlementModule is IAsyncOrderSettlementModule, IMarketEvent
             .updatePositionData(runtime.accountId, newPosition);
         emit MarketUpdated(
             updateData.marketId,
+            fillPrice,
```
bachstatter, line 160: Are you sure you want fillPrice from the order here?
This will be:
`(priceBeforeOrder  + priceAfterOrder) /2`
I would think to enable USD denominated OI calculations, you'd want the current price based on the current skew?

markets/perps-market/storage.dump.sol
```diff
@@ -593,7 +593,6 @@ library PerpsMarket {
         uint128 marketId;
         int256 skew;
         uint256 size;
-        int256 sizeDelta;
```
bachstatter, line 596: This dump file looks a bit weird to me..Can you doublecheck that it's correct?

Tburm: This is correct actually. I moved this `sizeDelta` into the runtime during order settlement. I think the dump file wasn't updated in the previous PR.

markets/perps-market/contracts/modules/AsyncOrderSettlementModule.sol
Tburm: You're right, we want the pyth price here.

markets/perps-market/storage.dump.sol
bachstatter: Cool thanks for checking.. We should get to the bottom of the dump.sol checker, CI should fail if dump file isn't updated.

bachstatter (approved):
Nice

leomassazza (approved):

PR #1711, opened 2023-07-07
Prevent modifyCollateral with pending orders

leomassazza:


markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -54,6 +54,11 @@ contract PerpsAccountModule is IAccountModule {
             account.id = accountId;
         }
 
+        // Check if there are pending orders
+        if (account.hasPendingOrders) {
```
Tburm, line 58: Should we add a check here that `amountDelta < 0`? I don't think a user should ever be prevented from depositing collateral, but it makes sense to prevent withdrawals.

Tburm: How could a person manipulate the timing by adding more collateral? I see how they could withdraw to cause the execution to revert, then deposit again when they want to execute.

bachstatter:
What's the difference between this and checking if the account have an order by `PerpsMarket.loadValid(marketId).asyncOrders[accountId];`?

leomassazza:
> What's the difference between this and checking if the account have an order by `PerpsMarket.loadValid(marketId).asyncOrders[accountId];`?

it needs to be done in a loop for each marketId

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.pending.test.ts
```diff
@@ -0,0 +1,184 @@
...
+              sizeDelta: bn(0.01),
+              settlementStrategyId: 0,
+              acceptablePrice: bn(10050), // 5% slippage
+              trackingCode: ethers.constants.HashZero,
+            }),
+          'PendingOrdersExist()',
+          systems().PerpsMarket
+        );
+      });
+    });
```
bachstatter, line 95: No sure how valuable, but we could add an assertion for an order on the same market too?

markets/perps-market/contracts/modules/PerpsAccountModule.sol
bachstatter: Yea troy's point does make sense to me

davidvuong:
You can move orders from market to an account level and track a mapping of `accountId -> struct { order, marketId }[]` then you can avoid the need for an additional variable you have to remember to turn on/off. Would require shuffling a few things around but it's one less variable that have to be kept tracked and stored in-sync.

If you track `marketId` in the `Order` struct then it can even just be `mapping(uint128 -> Order[])`, without needing a wrapper struct.

Orders aren't coupled to markets and they don't have an impact until settlement occurs. For settlement, as long as the keeper can find the associated marketId this order belongs to then it will operate just fine too.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
kaleb-keny: > How could a person manipulate the timing by adding more collateral? I see how they could withdraw to cause the execution to revert, then deposit again when they want to execute.

it just gives the user more tools to play with, like imagine with me during crashing market prices, the user submits intention at the brink of what is possible with his margin, that by the time the order is elegible for execution, he doesn't have enough margin to get executed, he gets a free option to add collateral to be allowed to execute the pending order during that execution time window

kaleb-keny: It is a safer design therefore, to just allow modification of collateral if no pending order is there

leomassazza: so, basically (we had this discussion on perpsV2, and there are some complex vectors or side effects) we want to prevent any change of collateral while orders are pending.

kaleb-keny: @Tburm yeah the attacker would need to check for a favorable with respect to the more lively prices offchain (which are expected to impact his position post-execution)

Tburm: Makes sense, thanks for the details. We should just make sure frontends are well aware of this, since users should know if/when they are prevented from depositing to fix underwater positions.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -54,6 +54,11 @@ contract PerpsAccountModule is IAccountModule {
             account.id = accountId;
         }
 
+        // Check if there are pending orders
+        if (AsyncOrder.load(accountId).sizeDelta != 0) {
+            revert PendingOrdersExist();
+        }
```
sunnyvempati, line 60: could move this into the account storage and do `account.checkPendingOrders()` or smth.  as much as possible trying to move validations into storage libraries.  non blocking; only if you think its cleaner

markets/perps-market/storage.dump.sol
```diff
@@ -547,6 +547,7 @@ library PerpsAccount {
     struct Data {
         mapping(uint128 => uint256) collateralAmounts;
         uint128 id;
+        bool hasPendingOrders;
```
sunnyvempati, line 550: dont think this is used; re-run `yarn build` 👍🏽

sunnyvempati (changes requested):
minor changes; logic is clear and looks good 👍🏽

PR #1712, opened 2023-07-07
Clean lint warnings

leomassazza:
reduce the lint number of warnings from 30 to 7. The remaining 7 are in proxy.sol contracts that cannot be changed (even change comments).


sunnyvempati (approved):
🙏🏽

PR #1713, opened 2023-07-10
Ensure all the cannon libs are upgraded throughout

leomassazza:
- using `yarn upgrade-interactive`
- Upgrade all cannon deps  (router plugin, cli, builder) to the latest version
- Regenerate subgraphs after contract updates
- Update spot-market subgraph OrderSettled event signature
- Increase default tests timeout for spot-markets from 40s to 200s (matching perps market config)


leomassazza:
LGTM

PR #1717, opened 2023-07-11
Add `accruedFunding` to `OrderSettled` event

Tburm:
Add `accruedFunding` to the `OrderSettled` event. Additionally, move the entire runtime into the event to avoid `stack too deep` errors and future-proof this event.


markets/perps-market/contracts/interfaces/IAsyncOrderSettlementModule.sol
```diff
@@ -22,22 +22,18 @@ interface IAsyncOrderSettlementModule {
...
-     * @param newSize New size of the position after settlement.
      * @param collectedFees Amount of fees collected by the protocol.
-     * @param settlementReward Amount of fees collected by the settler.
      * @param trackingCode Optional code for integrator tracking purposes.
      * @param settler address of the settler of the order.
      */
     event OrderSettled(
         uint128 indexed marketId,
         uint128 indexed accountId,
+        SettleOrderRuntime order,
```
leomassazza, line 34: This can hide changes in the event signature later if we change the SettleOrderRuntime struct that is used to hold internal variables to prevent stack too deep errors.

Tburm: Do you know of a time this was done in the past? I'm wondering what the effect would be. I think it could break our indexers if we removed properties. I also think it could have an impact on Dune, but would like to check how that is handled if we have an example.

PR #1719, opened 2023-07-12
SIP-326, additional pool configuration

Rickk137:


protocol/synthetix/contracts/modules/core/PoolModule.sol
```diff
@@ -228,6 +228,66 @@ contract PoolModule is IPoolModule {
...
+        uint128 poolId,
+        address collateral
+    ) external view override returns (bool) {
+        return !Pool.load(poolId).disabledCollaterals[collateral];
+    }
+
+    /**
+     * @inheritdoc IPoolModule
+     */
+    function setPoolCollateralIssuanceRatioD18(
```
noahlitvin, line 268: Unsure if we need D18 in the function name here? (Don't think we're doing this elsewhere.) Good to have it clarify in the parameter name though.

protocol/synthetix/contracts/modules/core/PoolModule.sol
```diff
@@ -228,6 +228,66 @@ contract PoolModule is IPoolModule {
...
+
+        pool.issuanceRatioD18[collateral] = issuanceRatioD18;
+
+        emit PoolCollateralDisabled(poolId, collateral, msg.sender);
+    }
+
+    /**
+     * @inheritdoc IPoolModule
+     */
+    function getPoolCollateralIssuanceRatioD18(
```
noahlitvin, line 284: Same as above

protocol/synthetix/contracts/modules/core/IssueUSDModule.sol
```diff
@@ -84,9 +84,11 @@ contract IssueUSDModule is IIssueUSDModule {
         // If the resulting debt of the account is greater than zero, ensure that the resulting c-ratio is sufficient
         (, uint256 collateralValue) = pool.currentAccountCollateral(collateralType, accountId);
         if (newDebt > 0) {
+            uint256 minIssuanceRatioD18 = pool.issuanceRatioD18[collateralType];
```
noahlitvin, line 87: Would it be bad practice to just put this on line 91 instead of creating a new variables?

protocol/synthetix/storage.dump.sol
```diff
@@ -544,6 +544,8 @@ library Pool {
         uint64 __reserved1;
         uint64 __reserved2;
         uint64 __reserved3;
+        mapping(address => bool) disabledCollaterals;
+        mapping(address => uint256) IssuanceRatio;
```
noahlitvin, line 548: Strange that this is UpperCamelCase, maybe need to regenerate? Unsure if it matters.

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -113,6 +118,8 @@ library Pool {
         uint64 __reserved1;
         uint64 __reserved2;
         uint64 __reserved3;
+        mapping(address => bool) disabledCollaterals;
```
noahlitvin, line 121: maybe this is `collateralTypeDisabled` ?

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -507,4 +514,14 @@ library Pool {
             );
         }
     }
+
+    /**
+     * @notice Shows if a given collateral type is enabled for deposits and delegation in given pool.
+     * @param collateral The address of the collateral.
+     */
+    function checkDelegationEnabled(Data storage self, address collateral) internal view {
+        if (self.disabledCollaterals[collateral]) {
```
noahlitvin, line 523: I think we should add the check that it's not the stablecoin here in anticipation of the cross-chain pool SIP. I can update the 326 SIP accordingly.

dbeal-eth (changes requested):
can we implement these new pool settings using a pattern similar to what I did in my conflicting PR https://github.com/Synthetixio/synthetix-v3/pull/1730/files#diff-8c613240ec445a67e372a22ee684cac640c6d15b9468bace9db73646fec17198R200

aka having a pool collateral configuration storage type, setting the whole structure at once in `setPoolCollateralConfiguration` call, etc.

PR #1721, opened 2023-07-14
Use load instead of mapping for AsyncOrder

leomassazza:
This PR updates the location of the AsyncOrder instead of adding it to the Account Data struct, it gets loaded from the AsyncOrder library


sunnyvempati (approved):
nice; very clean

PR #1722, opened 2023-07-17
Fix governance epoch settings storage

mjlescano:
This PR removes the `nextElectionSettings` object, and makes sure that the settings are moved the following epoch on resolution.

### DIFF
--- a/protocol/governance/storage.dump.sol
+++ b/protocol/governance/storage.dump.sol
@@ -154,7 +154,6 @@ library Council {
         SetUtil.AddressSet councilMembers;
         mapping(address => uint) councilTokenIds;
         uint lastElectionId;
-        ElectionSettings.Data nextElectionSettings;
     }
     function load() internal pure returns (Data storage store) {
         bytes32 s = _SLOT_COUNCIL_STORAGE;
@@ -178,8 +177,8 @@ library DebtShare {
     bytes32 private constant _SLOT_DEBT_SHARE_STORAGE = keccak256(abi.encode("io.synthetix.governance.DebtShare"));
     struct Data {
         address debtShareContract;
-        uint128[] debtShareIds;
-        CrossChainDebtShare.Data[] crossChainDebtShareData;
+        mapping(uint => uint128) debtShareIds;
+        mapping(uint => CrossChainDebtShare.Data) crossChainDebtShareData;
     }
     function load() internal pure returns (Data storage debtShare) {
         bytes32 s = _SLOT_DEBT_SHARE_STORAGE;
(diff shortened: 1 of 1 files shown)


dbeal-eth (commented):
It looks like this or gives the current owner of the election module contract control over the current, active election settings. This could be considered a governance attack vector. If thos is expected/fine, then I am ready to approve. It just seems like Amy voting parameter changes should always apply to the next epoch, the one active for the election after the next/current

protocol/governance/contracts/storage/DebtShare.sol
```diff
@@ -12,10 +12,10 @@ library DebtShare {
...
         // Synthetix c2 DebtShare contract used to determine vote power in the local chain
         IDebtShare debtShareContract;
-        // Array of debt share snapshot id's for each epoch
-        uint128[] debtShareIds;
-        // Array of CrossChainDebtShareData's for each epoch
-        CrossChainDebtShare.Data[] crossChainDebtShareData;
+        // debt share snapshot id's for each epoch
+        mapping(uint => uint128) debtShareIds;
+        // CrossChainDebtShareData's for each epoch
+        mapping(uint => CrossChainDebtShare.Data) crossChainDebtShareData;
```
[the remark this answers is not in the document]
dbeal-eth: Is there a deeper reason the out of bounds error was happening in the first place?

protocol/governance/contracts/storage/Election.sol
```diff
@@ -29,6 +29,7 @@ library Election {
         mapping(address => bytes32) ballotIdsByAddress;
         // Number of votes for each candidate
         mapping(address => uint) candidateVotes;
+        // Pointer to current election settings
```
dbeal-eth, line 32: Not really a pointer

Btw if new settings are ever added to epoch then suddenly the whole data structure is locked. Should we do an *actual* pointer (ie mapping within another global structure) instead?

dbeal-eth (changes requested):

protocol/governance/contracts/Proxy.sol
```diff
@@ -4,9 +4,8 @@ pragma solidity >=0.8.11 <0.9.0;
 import {UUPSProxyWithOwner} from "@synthetixio/core-contracts/contracts/proxy/UUPSProxyWithOwner.sol";
 
 contract Proxy is UUPSProxyWithOwner {
-    // solhint-disable-next-line no-empty-blocks
     constructor(
         address firstImplementation,
         address initialOwner
-    ) UUPSProxyWithOwner(firstImplementation, initialOwner) {}
+    ) UUPSProxyWithOwner(firstImplementation, initialOwner) {} // solhint-disable-line no-empty-blocks
```
leomassazza, line 10: even changing a comment will result in a new proxy deployed => change of proxy address

barrasso (commented):
code changes and election module tests LGTM.

just see one cheeky failure on the `test (core-utils)` job: https://github.com/Synthetixio/synthetix-v3/actions/runs/5591845729/jobs/10223488775?pr=1722

dbeal-eth (approved):
I still have a comment unresolved regarding initOrUpgrade setting function parameters

ajsantander (commented):
LGTM

PR #1723, opened 2023-07-19
Combined Readme for all contracts

noisekit:
- Combine all contract docs into a single file for an easy copy-paste to the docs
- Update page template
- Enable docs diff check for PRs


bachstatter (approved):
I dont have too much experience with natspec, but looks good to me

PR #1730, opened 2023-07-26
SIP-333 + SIP-334

dbeal-eth:
implements 2 sips pending:
* ability to limit capacity of collaterals in a pool. This is important/required for the passive snx pool.
* oracle manager, ability to specify additional runtime parameters in the `process` call. Can be used for perps market to verify collateral value on liquidation


noahlitvin:
Closing in favor of #1753

PR #1731, opened 2023-07-27
add new views

sunnyvempati:
- Adds three new views: `computeOrderFees`, `getWithdrawableMargin`, `getRequiredMargins`.
- Uses initial margin instead of maintenance margin for withdrawable margin calc.
- Tests (and a small `computeFees` types change)

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -75,13 +75,18 @@ library PerpsAccount {
     )
         internal
         view
-        returns (bool isEligible, int256 availableMargin, uint256 requiredMaintenanceMargin)
+        returns (
+            bool isEligible,
+            int256 availableMargin,
+            uint256 requiredInitialMargin,
+            uint256 requiredMaintenanceMargin
+        )
     {
         availableMargin = getAvailableMargin(self);
         if (self.openPositionMarketIds.length() == 0) {
-            return (false, availableMargin, 0);
+            return (false, availableMargin, 0, 0);
         }
-        requiredMaintenanceMargin = getAccountMaintenanceMargin(self);
+        (requiredInitialMargin, requiredMaintenanceMargin) = getAccountRequiredMargins(self);
         isEligible = requiredMaintenanceMargin.toInt() > availableMargin;
     }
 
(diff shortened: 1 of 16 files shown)


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -211,9 +217,9 @@ library PerpsAccount {
     /**
      * @notice  This function returns the minimum margin an account requires to stay above liquidation threshold
      */
-    function getAccountMaintenanceMargin(
+    function getAccountRequiredMargins(
         Data storage self
-    ) internal view returns (uint accountMaintenanceMargin) {
+    ) internal view returns (uint accountMaintenanceMargin, uint initialMaintenanceMargin) {
```
Tburm, line 222: The naming looks backwards here compared to the return, and compared to where it is called on `getAccountRequiredMargins`

markets/perps-market/contracts/storage/PerpsMarketConfiguration.sol
```diff
@@ -99,8 +99,10 @@ library PerpsMarketConfiguration {
         maintenanceMarginRatio = impactOnSkew.mulDecimal(self.maintenanceMarginRatioD18);
 
         uint256 notional = sizeAbs.mulDecimal(price);
-        initialMargin = notional.mulDecimal(initialMarginRatio);
-        maintenanceMargin = notional.mulDecimal(maintenanceMarginRatio);
+        initialMargin = notional.mulDecimal(initialMarginRatio) + self.minimumPositionMargin;
```
Tburm, line 102: Why is the minimum added to every position instead of doing something like the max of the two values? Will this cause problems at really small margin values?

For example: If the minimum is $50 and the trader is opening a position which requires $50 in margin, won't this require them to have $100?

markets/perps-market/test/integration/Account/Margins.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      const [initialMargin, maintenanceMargin] = await systems().PerpsMarket.getRequiredMargins(
+        accountId
+      );
+      assertBn.equal(
+        initialMargin,
+        btcInitialMargin.add(ethInitialMargin).add(minimumPositionMargin).toBN()
+      );
+      assertBn.equal(
+        maintenanceMargin,
+        btcMaintenanceMargin.add(ethMaintenanceMargin).add(minimumPositionMargin).toBN()
```
Tburm, line 160: Same as before, I'm not sure why we would add the minimum margin here. Instead I would expect to test a situation where the required margin is _below_ the minimum, but this function returns the minimum.

leomassazza (approved):
LGTM, contingent of tburn's comment (that initially I think is not an issue but please answer it)

PR #1732, opened 2023-07-29
Add Owner Guardianship

mjlescano:
This PR creates a new "Guardian Role" which is the only address that will be able to request for an owner change. And this new owner can only accept the ownership after a configured delay have passed (`7 days`).

This new Guardian can also nominate another Guardian.

TODOs
- [x] Create `GuardianModule` to be able to manage the new role 
- [x] Override `OwnerModule` logic to take into account the Guardian role
- [ ] Add tests


dbeal-eth (approved):
Really like the strategy here

PR #1736, opened 2023-07-31
Fix minimumCredit (plus small fixes)

leomassazza:
- Fix minimumCredit calculation
- remove some unused imports
- remove D18 from event name and function


sunnyvempati (approved):
🚢

PR #1740, opened 2023-08-01
New liquidation params

sunnyvempati:
- removed `maintenanceMarginRatio`
- added `minimumInitialRatio` -> this ratio is added to the computed initial ratio based on the configured `initialMarginRatio`
- added `maintenaceMarginScalar` -> this scalar is multiplied by initial margin ratio to get the maintenance margin ratio


markets/perps-market/test/integration/Liquidation/Liquidation.margin.test.ts
```diff
@@ -307,7 +314,7 @@ describe('Liquidation - margin', async () => {
   describe('price change - eligible for liquidation', () => {
     [
       bn(31000), // btc
-      bn(1775), // eth
+      bn(1820), // eth
```
sunnyvempati, line 317: made it so its closer to liquidation

leomassazza (approved):
LGTM

PR #1745, opened 2023-08-04
Add missing docs on interfaces and modules

leomassazza:


markets/perps-market/contracts/interfaces/IPerpsMarketFactoryModule.sol
```diff
@@ -5,16 +5,50 @@ import {IMarket} from "@synthetixio/main/contracts/interfaces/external/IMarket.s
...
 
+    /**
+     * @notice Initializes the factory.
+     * @dev this function should be called only once.
+     * @return globalPerpsMarketId Id of the global perps market id.
+     */
     function initializeFactory() external returns (uint128);
 
+    /**
+     * @notice Sets the syntetix system.
```
Tburm, line 34:
```suggestion
     * @notice Sets the synthetix system.
```

markets/perps-market/contracts/interfaces/IPerpsMarketModule.sol
```diff
@@ -7,35 +7,89 @@ import {AsyncOrder} from "../storage/AsyncOrder.sol";
...
+    /**
+     * @notice Market Summary structured data.
+     */
     struct MarketSummary {
+        // @dev Skew of the market
         int256 skew;
+        // @dev Size of the market
         uint256 size;
+        // @dev Max open interest of the market
         uint256 maxOpenInterest;
```
Tburm, line 19: Should we specify the units are in the native asset?

PR #1750, opened 2023-08-07
SIP-325: add multicall through

dbeal-eth:
implements [SIP-325](https://sips.synthetix.io/sips/sip-325/).

most of the code was already there and ready to go, just needed to be commented out for its inevitable inclusion in the protocol.


dbeal-eth:
this needs to be reworked

sunnyvempati (approved):

dbeal-eth:
in the future we may take up the auditors suggestion to proxy through a secondary contract, but for now this is what we need for the testnet trading comp

dbeal-eth:
merging because remaining issues are related to existing main branch failures

PR #1751, opened 2023-08-08
Add liq reward to async order validation

sunnyvempati:
- previously we were only checking liquidation rewards of maintenance margin and not taking into account the changes to liquidation reward based on the new order size.  this pr fixes that.
- thorough test testing the different scenarios.


leomassazza (approved):
LGTM

PR #1753, opened 2023-08-08
SIP 323 326 331 332 333 334 335

Rickk137:


markets/spot-market/test/SpotMarketFactory.test.ts
```diff
@@ -37,6 +37,24 @@ describe('SpotMarketFactory', () => {
       });
     });
 
+    it('renounce market ownership', async () => {
```
bachstatter, line 40: Wanna add a tests where a non owner tried to renounce too?

protocol/synthetix/contracts/interfaces/IMulticallModule.sol
```diff
@@ -11,7 +11,7 @@ interface IMulticallModule {
      * @param data Array of calldata objects, one for each function that is to be called in the system.
      * @return results Array of each `delegatecall`'s response corresponding to the incoming calldata array.
      */
-    function multicall(bytes[] calldata data) external payable returns (bytes[] memory results);
+    function multicall(bytes[] calldata data) external returns (bytes[] memory results);
```
bachstatter, line 14: What was the reason for this change? Will I still be able to call payable methods when the `multicall` isn't payable?

protocol/synthetix/contracts/modules/core/IssueUSDModule.sol
```diff
@@ -96,8 +100,14 @@ contract IssueUSDModule is IIssueUSDModule {
...
 
+        AssociatedSystem.Data storage usdToken = AssociatedSystem.load(_USD_TOKEN);
+
         // Mint stablecoins to the sender
-        AssociatedSystem.load(_USD_TOKEN).asToken().mint(msg.sender, amount);
+        usdToken.asToken().mint(address(this), amount);
+
+        account.collaterals[usdToken.getAddress()].increaseAvailableCollateral(
+            CollateralConfiguration.load(usdToken.getAddress()).convertTokenToSystemAmount(amount)
+        );
```
bachstatter, line 110: Is this line here just to ensure we revert if we loose precision?

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -507,4 +528,25 @@ library Pool {
             );
         }
     }
+
+    function checkPoolCollateralLimit(
+        Data storage self,
+        address collateralType,
+        uint256 collateralAmountD18
+    ) internal view {
+        uint256 maxDeposit = self.collateralConfigurations[collateralType].maxDepositD18;
```
bachstatter, line 537: `self.vaults[collateralType].currentCollateral()` we can store this in a variable since it's used twice

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -507,4 +528,25 @@ library Pool {
...
+
+        if (
+            (self.collateralDisabledByDefault && maxDeposit == 0) ||
+            (maxDeposit > 0 &&
+                self.vaults[collateralType].currentCollateral() + collateralAmountD18 > maxDeposit)
```
bachstatter, line 542: Could we simplify this a bit, maybe event revert with some more specific errors? 
Should we revert if 
`collateralDisabledByDefault = false` and `maxDeposit = 0`?

protocol/synthetix/test/integration/modules/core/PoolModuleFundAdmin.test.ts
```diff
@@ -683,4 +687,102 @@ describe('PoolModule Admin', function () {
...
+    });
+
+    it('disable the collateral by pool owner', async () => {
+      await systems()
+        .Core.connect(user1)
+        .setPoolCollateralConfiguration(thirdPoolId, collateralAddress(), {
+          maxDepositD18: bn(10),
+          issuanceRatioD18: bn(2),
+        });
+    });
```
bachstatter, line 731: Missing some assert here?

protocol/synthetix/test/integration/modules/core/VaultModule.test.ts
```diff
@@ -314,6 +326,106 @@ describe('VaultModule', function () {
...
+        await systems()
+          .Core.connect(user1)
+          .delegateCollateral(
+            accountId,
+            fakeVaultId,
+            collateralAddress(),
+            depositAmount.div(50),
+            ethers.utils.parseEther('1')
+          );
+      });
```
bachstatter, line 397: Missing some asserts in this and the it block above

utils/core-modules/contracts/modules/AssociatedSystemsModule.sol
```diff
@@ -126,7 +126,10 @@ contract AssociatedSystemsModule is IAssociatedSystemsModule {
         AssociatedSystem.Data storage store = AssociatedSystem.load(id);
 
         if (store.proxy != address(0)) {
-            _upgradeToken(id, impl);
+            if (store.impl != impl) {
+                _upgradeToken(id, impl);
+            }
+            ITokenModule(store.proxy).initialize(name, symbol, decimals);
```
bachstatter, line 132: Might be worth adding a comment here. We do we wanna call `initialize` even when we're not calling `_upgradeToken`?

protocol/synthetix/contracts/modules/core/IssueUSDModule.sol
fritzschoff: yes but maybe Im too careful and we can save gas here

protocol/synthetix/storage.dump.sol
```diff
@@ -665,3 +676,8 @@ library CcipClient {
         bool strict;
     }
 }
+
+// @custom:artifact hardhat/console.sol:console
+library console {
```
bachstatter, line 681: Can we do a yarn clean and yarn build to get this removed ( and fix wrong version number)

PR #1754, opened 2023-08-09
Add view and modify event

sunnyvempati:
- added new view to get all existing market ids
- add accrued funding to `OrderSettled` event


docs/Contracts.md
```diff
@@ -3798,71 +3798,6 @@ There is a synthetix v3 core system supply cap also set. If the current supply b
 
 - [Back to TOC](#smart-contracts)
 
-### Async Order Module
```
sunnyvempati, line 3801: i guess we needed to run `yarn docgen`.  ignore these changes, they're needed

markets/perps-market/test/integration/Orders/OffchainAsyncOrder.settle.test.ts
```diff
@@ -375,6 +375,7 @@ describe('Settle Offchain Async Order test', () => {
               accountId,
               fillPrice,
               0,
```
leomassazza, line 377: Would like to have a test that actually uses a value here.

leomassazza (approved):
LGTM. Just one test asked, but can be done later if needed

PR #1760, opened 2023-08-11
v3 CI fix hardhat not loading solc on Node18

noisekit:
Update Node version to avoid hardhat silently bail with success

For context:
Hardhat randomly just silently quitting with exit code 0 when downloading solc compiler
It was an issue with NodeJS in *some* versions and looks like 18.16.0 is one of the affected versions (see https://github.com/NomicFoundation/hardhat/issues/3877)
Using fresher (or older version of node solves it).

on `18.17.1` couple of the core-utils tests are failing, so sticking to `18.15.0` for now.

### DIFF
--- a/.github/workflows/check-packages.yml
+++ b/.github/workflows/check-packages.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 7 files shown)


dbeal-eth (approved):
It's time we update to node 18 anyway. Maybe node 20 while we are at it

PR #1762, opened 2023-08-11
Run only tests for PRs and run coverage report on main

noisekit:
Even though we run coverage report on each commit we never actually submit those reports to Codecov

@dbeal-eth suggested not to run cov report on PRs. That makes sense as `main` takes like 15min to do cov report (because running on hardhat net is slow!)

Because of that now we only run tests on PRs (must be pretty quick) and coverage report (which also includes all the tests) will only run on main branch

This should speed up day to day dev a lot buy running tests much much faster in CI

### DIFF
--- a/.github/workflows/test-with-coverage.yml
+++ b/.github/workflows/test-with-coverage.yml
@@ -1,10 +1,12 @@
-name: Test
+name: Coverage
 
 on:
-  pull_request: {}
+  push:
+    branches:
+      - "main"
 
 jobs:
-  test:
+  coverage:
     runs-on: ubuntu-latest
     env:
       CANNON_IPFS_URL: "http://127.0.0.1:5001"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -15,46 +12,26 @@ jobs:
...
+            "@synthetixio/oracle-manager",
+            # "@synthetixio/governance", # no tests :/
+
+            "@synthetixio/core-contracts",
+            "@synthetixio/core-utils",
+            "@synthetixio/core-modules",
+            "@synthetixio/hardhat-storage",
+            "@synthetixio/sample-project",
+
+            # "@synthetixio/legacy-market", # tests fail
```
dbeal-eth, line 26: Tests shouldn't fail, what can I say. The tests were refreshed just a couple weeks ago.

dbeal-eth (approved):
Definitely goodness in here

PR #1764, opened 2023-08-11
M 1 renounce pool ownership

fritzschoff:


protocol/synthetix/storage.dump.sol
```diff
@@ -665,3 +666,8 @@ library CcipClient {
         bool strict;
     }
 }
+
+// @custom:artifact hardhat/console.sol:console
+library console {
```
bachstatter, line 671: Do a yarn build again so that this little guy gets removed

bachstatter: That should also fix the version number

bachstatter: try a yarn clean first maybe

PR #1767, opened 2023-08-11
Add order expired event

leomassazza:


markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -151,23 +151,9 @@ library AsyncOrder {
         checkWithinSettlementWindow(order, strategy);
     }
 
-    /**
-     * @dev Reverts if the order does not belongs to the market or not exists. Otherwise, returns the order.
-     * @dev non-existent order is considered an order with sizeDelta == 0.
-     */
-    function createValid(
```
leomassazza, line 158: Not useful anymore since we need the intermediate order to get the event information

sunnyvempati (approved):
yeah kind of annoying but not a huge deal

PR #1771, opened 2023-08-15
CircleCI

noisekit:
- Replace GH Actions with CircleCI
- Simulate release is still with GH Actions until cannon has fixed RPC proxy
- Coverage report is also GH Actions on main branch (will move soon too, maybe even in this PR)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,7 @@ jobs:
       - restore_cache:
           keys:
             - cannon-{{ .Environment.CACHE_VERSION }}-{{ checksum "/tmp/version--contracts.txt" }}
+            - cannon-{{ .Environment.CACHE_VERSION }}-
 
       - version-testable-contracts
       - restore_cache:
@@ -200,6 +201,7 @@ jobs:
       - restore_cache:
           keys:
             - cannon-{{ .Environment.CACHE_VERSION }}-{{ checksum "/tmp/version--contracts.txt" }}
+            - cannon-{{ .Environment.CACHE_VERSION }}-
       - restore_cache:
           keys:
             - testable-hardhat-cache-{{ .Environment.CACHE_VERSION }}-{{ checksum "/tmp/version--contracts.txt" }}


protocol/synthetix/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "yarn build:contracts",
     "build:contracts": "hardhat storage:verify && hardhat cannon:build",
     "generate-testable": "rm -rf contracts/generated && hardhat generate-testable",
-    "build-testable": "yarn generate-testable && hardhat cannon:build cannonfile.test.toml",
```
bachstatter, line 13: `yarn generate-testable` not needed?

.circleci/config.yml
```diff
@@ -84,7 +86,361 @@ commands:
...
+
+      - restore_cache:
+          keys:
+            - hardhat-{{ .Environment.SOLC_VERSION }}
+
+      - version-contracts
+      - restore_cache:
+          keys:
+            - cannon-{{ .Environment.CACHE_VERSION }}-{{ checksum "/tmp/version--contracts.txt" }}
+
```
dbeal-eth, line 130: Cannon builds can be greatly increased in speed if we load an inexact cache , since only changes contracts will be rebuilt into the image. Would reccomend loading from additional cache prefixes even if contracts chance

dbeal-eth: Aka adding a second restore key with the checksum removed should do the trick

.circleci/config.yml
```diff
@@ -84,7 +86,361 @@ commands:
...
+      - checkout
+      - yarn-install
+
+      - restore_cache:
+          keys:
+            - hardhat-{{ .Environment.SOLC_VERSION }}
+
+      - run:
+          name: "Compile TS"
+          command: yarn workspaces foreach --topological-dev --recursive --verbose --from "<< parameters.workspace >>" run build:ts
```
dbeal-eth, line 329: Technically core contracts don't depend on anything so there is no need to go topological here

.circleci/config.yml
```diff
@@ -183,10 +539,123 @@ jobs:
...
+
+      - test-subgraph:
+          name: "test-subgraph-spot-market"
+          workspace: "@synthetixio/spot-market-subgraph"
+
+      - test-subgraph:
+          name: "test-subgraph-perps-market"
+          workspace: "@synthetixio/perps-market-subgraph"
+
+      #- simulate-release: TODO: enable when cannon has fixed RPC proxy
```
dbeal-eth, line 611: Technically you can start simulating release now for testnets and it's kind of important because there are some invalid release configs that people have tried to do in the past

dbeal-eth (approved):
Please consider changes but what we have here is good for start

PR #1774, opened 2023-08-15
Limit number of positions and collaterals per account

leomassazza:


markets/perps-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -71,6 +71,12 @@ contract AsyncOrderModule is IAsyncOrderModule {
             );
         }
 
+        // Check if there's a previous position open on this market, if not check if the account can open a new position
+        if (PerpsMarket.accountPosition(commitment.marketId, commitment.accountId).size == 0) {
+            //is a new position, check if not exceeding max positions per account
+            PerpsAccount.load(commitment.accountId).canOpenNewPosition();
+        }
```
sunnyvempati, line 78: i would move the if condition into the validation itself.  maybe something like this:

```
PerpsAccount.validateMaxPositions(commitment.marketId, commitment.accountId);
```

and load the position and account in the validation itself.  just cleans up the modules a bit more.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -53,6 +53,11 @@ contract PerpsAccountModule is IPerpsAccountModule {
         PerpsAccount.Data storage account = PerpsAccount.create(accountId);
         uint128 perpsMarketId = perpsMarketFactory.perpsMarketId;
 
+        if (account.collateralAmounts[synthMarketId] == 0) {
+            // if the account has no collateral in this market, we need to check if it can add a new collateral
+            account.canAddNewCollateral();
+        }
```
sunnyvempati, line 59: same as above, move the if condition into the validation

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -69,6 +73,24 @@ library PerpsAccount {
         }
     }
 
+    function canOpenNewPosition(Data storage self) internal view {
```
sunnyvempati, line 76: would rename this to `validateMaxPositions`.  `can` implies it will return a boolean.

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -69,6 +73,24 @@ library PerpsAccount {
...
+        uint128 maxPositionsPerAccount = GlobalPerpsMarketConfiguration
+            .load()
+            .maxPositionsPerAccount;
+        if (maxPositionsPerAccount <= self.openPositionMarketIds.length()) {
+            revert MaxPositionsPerAccountReached(maxPositionsPerAccount);
+        }
+    }
+
+    function canAddNewCollateral(Data storage self) internal view {
```
sunnyvempati, line 85: same as above, maybe `validateMaxCollaterals`

markets/perps-market/contracts/modules/AsyncOrderModule.sol
sunnyvempati: wish we still had `createValid`, then this call for validating max positions can just live there.

i have some ideas on this, but can leave it to a separate PR 👍🏽

markets/perps-market/contracts/modules/AsyncOrderSettlementModule.sol
```diff
@@ -140,6 +141,11 @@ contract AsyncOrderSettlementModule is IAsyncOrderSettlementModule, IMarketEvent
         PerpsMarketFactory.Data storage factory = PerpsMarketFactory.load();
         PerpsAccount.Data storage perpsAccount = PerpsAccount.load(runtime.accountId);
 
+        if (oldPosition.size == 0) {
+            //is a new position, check if not exceeding max positions per account
+            perpsAccount.canOpenNewPosition();
+        }
```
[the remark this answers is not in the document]
leomassazza: Yes... I was wondering how that can happen, and keep it just in case. But, yeah, I'll get rid of it.

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -151,6 +151,25 @@ library AsyncOrder {
...
+     * @dev Reverts if accont cannot open a new position (due to max allowed reached).
+     */
+    function updateValid(
+        Data storage self,
+        OrderCommitmentRequest memory newRequest,
+        SettlementStrategy.Data storage strategy
+    ) internal {
+        checkPendingOrder(newRequest.accountId);
+
+        PerpsAccount.load(newRequest.accountId).validateMaxPositions(newRequest.marketId);
```
sunnyvempati, line 166: can just do the account loading in the `validateMaxPositions` function.

so could just do `PerpsAccount.validateMaxPositions(newRequest.accountId, newRequest.marketId);`

non blocking

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -69,6 +73,28 @@ library PerpsAccount {
         }
     }
 
+    function validateMaxPositions(Data storage self, uint128 marketId) internal view {
+        if (PerpsMarket.accountPosition(marketId, self.id).size == 0) {
+            uint128 maxPositionsPerAccount = GlobalPerpsMarketConfiguration
+                .load()
+                .maxPositionsPerAccount;
+            if (maxPositionsPerAccount <= self.openPositionMarketIds.length()) {
```
sunnyvempati, line 81: why <= instead of just <?

if max is set to 5, can't you have 5 open positions?

sunnyvempati (approved):
one question but looks good

markets/perps-market/contracts/storage/AsyncOrder.sol
leomassazza: good call, will update

markets/perps-market/contracts/storage/PerpsAccount.sol
leomassazza: because we check here if we can add another => in your example we'll check when adding the 5th so we need to have 4 already there at most

PR #1775, opened 2023-08-15
Perps V3: Update subgraph

Tburm:
Update the subgraph to include:
- Match the latest testnet deploy
- `accruedFunding` on the `OrderSettled` event
- Index the `PreviousOrderExpired` event
- Index the `CollateralModified` event

Note: Since the function signatures changes, we will be missing `OrderSettled` events from before the deploy on August 14. Since there is not mainnet deploy of the old version it shouldn't be necessary to handle both signatures.


Tburm:
@noisekit Looking at your PR, I think that will be really useful to split by network. I've already run into some conflicts as we add things to the contracts and deploy to cannon -> testnet -> mainnet where a flow like this will help a lot.

Tburm:
@noisekit I think the contracts are currently in sync with cannon and OP Goerli, so likely no conflicts. We haven't changed any events since the last testnet deploy.

PR #1776, opened 2023-08-15
Sip-332: update pool collateral configuration logic

Rickk137:


protocol/synthetix/contracts/modules/core/PoolModule.sol
```diff
@@ -197,6 +197,24 @@ contract PoolModule is IPoolModule {
         emit PoolConfigurationSet(poolId, newMarketConfigurations, msg.sender);
     }
 
+    /**
+     * @inheritdoc IPoolModule
+     */
+    function disablePoolCollateralByDefault(uint128 poolId) external override {
+        Pool.Data storage pool = Pool.loadExisting(poolId);
```
dbeal-eth, line 204: check msg.sender = pool owner?

dbeal-eth (changes requested):

PR #1783, opened 2023-08-22
Refactor synthetix subgraph splitting all handlers and tests

noisekit:
This is a prep work to get reusable handlers and tests across multiple network deployments

Overall structure might change, but this per-function split makes it much easier to move forward and make changes


bachstatter (commented):
Looks good to me, but might require some coordination with jz to get his PR in too

jmzwar (approved):

PR #1787, opened 2023-08-23
Add satellites NFT distribution

mjlescano:


barrasso (approved):

protocol/governance/contracts/modules/core/BaseElectionModule.sol
```diff
@@ -346,16 +346,32 @@ contract BaseElectionModule is
...
+                this._recvResolve.selector,
+                election.winners.values(),
+                newEpochIndex
+            ),
+            _CROSSCHAIN_GAS_LIMIT
+        );
+    }
+
+    function _recvResolve(address[] calldata winners, uint256 newEpochIndex) external {
+        CrossChain.onlyOnChainAt(0);
```
barrasso, line 362: I just noticed this line here: `CrossChain.onlyOnChainAt(0);`

I think this should be removed since `_recvResolve` needs to be called on every network in order to distribute the NFTs on each chain.

PR #1788, opened 2023-08-23
max pd / endorsed liquidator

sunnyvempati:


markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -127,6 +132,24 @@ library PerpsMarket {
             self.lastUtilizedLiquidationCapacity += liquidatableAmount;
         }
 
+        // liquidatable amount is 0 only if there's no more capacity in the window
+        // but if maxLiquidationPd is set, and the current market p/d is less than that,
+        // then allow for an extra block of liquidation
+        uint maxLiquidationPd = marketConfig.maxLiquidationPd;
```
leomassazza, line 138: it should check the latest liquidation was at least in the previous block

leomassazza (approved):

PR #1798, opened 2023-08-29
Add notEqual

bachstatter:


utils/core-utils/src/utils/assertions/assert-bignumber.ts
```diff
@@ -97,4 +97,15 @@ export = {
       });
     }
   },
+  notEqual(a: BigNumberish, b: BigNumberish) {
```
davidvuong, line 100: Doc above this fn but LGTM.

davidvuong (approved):

PR #1803, opened 2023-08-31
Withdraw different collateral fix

leomassazza:
Issue: if an account wants to withdraw collateral that is available in the system but not on the account it reverts with overflow. 
Fix: Add an explicit revert

Hidden issue: when withdrawing collateral, use the snxUSD value to check against limits, not the nominal value

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -45,7 +45,10 @@ library PerpsAccount {
         SetUtil.UintSet openPositionMarketIds;
     }
 
-    error InsufficientCollateralAvailableForWithdraw(uint available, uint required);
+    error InsufficientCollateralAvailableForWithdraw(
+        uint availableUsdDenominated,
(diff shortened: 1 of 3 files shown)


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -176,8 +176,14 @@ library PerpsAccount {
...
         Data storage self,
         uint128 synthMarketId,
-        uint256 amountToWithdraw
+        uint256 amountToWithdraw,
+        ISpotMarketSystem spotMarket
     ) internal view returns (uint256 availableWithdrawableCollateralUsd) {
+        uint collateralAmount = self.collateralAmounts[synthMarketId];
+        if (collateralAmount < amountToWithdraw) {
+            revert InsufficientSynthCollateral(synthMarketId, collateralAmount, amountToWithdraw);
+        }
```
sunnyvempati, line 185: moved this up since if they don't have enough of hte collateral, we should abort anyway before other checks

PR #1813, opened 2023-09-05
All functions payable

dbeal-eth:
to ensure we dont have any problems with multicall, we are switching all functions to be payable. This change also includes a new package which enforces the use of payable so it does not get forgotten.

UpgradeModule and related immutable modules have been excluded from lint checks.

Other warnings cleaned up.


noahlitvin:
Closing in favor of #1817

PR #1817, opened 2023-09-07
Testnet Competition Code Changes

noahlitvin:
Notes to auditors :
- Changed `getCollateralPrice` to take in an amount.  This will ensure when synths need prices, the amount can be sent into the oracle node to determine price. (Changed by **sunny**)
- Removed `checkPriceDeviation` from order settlements on both spot/perps markets.
- Perps market settlement calls `getUpdateFee` first prior to calling `parseUpdatePriceFeeds`


noahlitvin:
Released as `3.3.3-dev.32324850`

noahlitvin:
Released as `3.3.3-dev.05f72e70`

dbeal-eth (approved):
good.

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -476,7 +479,7 @@ library Pool {
     ) internal view returns (uint256 collateralAmountD18, uint256 collateralValueD18) {
         uint256 collateralPriceD18 = CollateralConfiguration
             .load(collateralType)
-            .getCollateralPrice();
+            .getCollateralPrice(DecimalMath.UNIT);
```
bachstatter, line 482: So we do not want to value the users collateral based on amounts?

PR #1819, opened 2023-09-10
Refactor max liquidation amount to use one array

sunnyvempati:
- Continue pushing to `Liquidation.Data[]` and check the previous timestamps for the liquidation amount in the current liquidation window.

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsMarket.sol
+++ b/markets/perps-market/contracts/storage/PerpsMarket.sol
@@ -190,9 +190,7 @@ library PerpsMarket {
         latestLiquidationTimestamp = self.liquidationData[currentIndex].timestamp;
         uint windowStartTimestamp = block.timestamp - marketConfig.maxSecondsInLiquidationWindow;
 
-        while (
-            currentIndex >= 0 && self.liquidationData[currentIndex].timestamp > windowStartTimestamp
-        ) {
+        while (self.liquidationData[currentIndex].timestamp > windowStartTimestamp) {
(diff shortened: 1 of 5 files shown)


markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -148,78 +149,61 @@ library PerpsMarket {
...
-            latestLiquidationTimestamp = MathUtil
-                .max(latestLiquidationTimestamp, liquidationTimestamp)
-                .to128();
-
-            if (liquidationTimestamp >= windowStartTimestamp) {
-                accumulatedLiquidationAmounts += self.liquidationAmounts[i];
-            }
-        }
+        while (
+            currentIndex >= 0 && self.liquidationData[currentIndex].timestamp > windowStartTimestamp
```
leomassazza, line 194: checking the index is done inside the look on the break condition
```suggestion
            self.liquidationData[currentIndex].timestamp > windowStartTimestamp
```

leomassazza (approved):

kaleb-keny:
hey @sunnyvempati my only comment is at this line
https://github.com/Synthetixio/synthetix-v3/blob/dfb878f0a91a8c0f7468066a9473be7d0a22e2ab/markets/perps-market/contracts/storage/PerpsMarket.sol#L145C21-L145C43
Depending on afif's feedback, 
instead of using `maxLiquidationInWindow`, maybe we should be using
`skewScale*abs(maxLiquidationPd-currentPd)` on that line

PR #1828, opened 2023-09-15
Update deps

noisekit:
Only these left untouched

- `node-fetch` + `chalk` moved to use ESM and no longer compatible
- `solc` is locked to what we use for contracts
- `ethers` 6 is breaking change and has poor tooling support, still cannot migrate

image

NOTE: the deps updated in PR are related to build, testing, docs generation, publishing. So passing all the tests should be a good indicator all is good. I did also dry-run publish with lerna and it did work ok. If anything I missed is broken - will fix asap as usual.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -13,10 +13,7 @@
   "command": {
     "version": {
       "allowBranch": [
-        "main",
-        "lerna",
-        "teleporters",
-        "feature/3.3.4"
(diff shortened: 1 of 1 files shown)


bachstatter (approved):
Looks good. A little scary to update so many things at once.. I guess we can always roll back if something is broken

PR #1829, opened 2023-09-15
add base ElectionModuleSatellite setup

mjlescano:
This PR enables voting cast and tokens distribution logic on Satellite chains.

The idea for implementing deployments is to deploy an instance of `cannonfile.toml` in the Mothership network, and then one of `cannonfile.satellite.toml` on each satellite network. 

Keeping that in mind, the `ElectionModuleSatellite.sol` is intended to be as small as possible, and to keep the least amount of state, and let the mothership to take care of everything state related. So, the Satellite networks just keep in state who are the member councils that have Gov Tokens.

## TODO
- [x] Allow the crosschain module to setup a target address when broadcasting, in case that the contracts are different.


protocol/governance/cannonfile.satellite.toml
```diff
@@ -0,0 +1,63 @@
...
+artifact = "contracts/modules/council-nft/CouncilTokenModule.sol:CouncilTokenModule"
+
+[router.CoreRouter]
+contracts = [
+    "AssociatedSystemsModule",
+    "CrossChainModule",
+    "ElectionModuleSatellite",
+    "InitialModuleBundle",
+]
+depends = [
```
dbeal-eth, line 38: dont need this anymore

protocol/governance/cannonfile.satellite.toml
```diff
@@ -0,0 +1,63 @@
...
+target = ["InitialProxy"]
+from = "<%= settings.owner %>"
+func = "upgradeTo"
+args = ["<%= contracts.CoreRouter.address %>"]
+factory.CoreProxy.abiOf = ["CoreRouter"]
+factory.CoreProxy.event = "Upgraded"
+factory.CoreProxy.arg = 0
+depends = ["contract.InitialProxy", "router.CoreRouter"]
+
+[invoke.init_election_module_satellite]
```
dbeal-eth, line 55: if you remove the `_satellite` postfix the step name could be overridden by the mothership cannonfile with its own init function

protocol/governance/cannonfile.satellite.toml
```diff
@@ -0,0 +1,63 @@
+name = "synthetix-governance-satellite"
+description = "On-chain voting for synthetix councils - Logic for satellite chains"
+version = "<%= package.version %>"
+include = ["./tomls/proxy-base.toml", "./tomls/council-token.toml"]
+
+[setting.salt]
+defaultValue = "governance-satellite"
+
+# Deployment Owner, defaults to first hardhat account
+[setting.owner]
```
dbeal-eth, line 10: these settings are also shared with mothership

protocol/governance/contracts/modules/core/ElectionModuleSatellite.sol
```diff
@@ -0,0 +1,78 @@
...
+import {CouncilMembers} from "../../storage/CouncilMembers.sol";
+
+contract ElectionModuleSatellite is IElectionModuleSatellite, ElectionCredentials {
+    using SetUtil for SetUtil.AddressSet;
+    using CrossChain for CrossChain.Data;
+    using CouncilMembers for CouncilMembers.Data;
+
+    uint256 private constant _CROSSCHAIN_GAS_LIMIT = 100000;
+
+    function initCouncilMembers(
```
dbeal-eth, line 19: im guessing this function has to be called with the same parameters on all chains (including mothership?)

dbeal-eth (changes requested):
the mothership is basically the same as a Satellite deployment with some additions (caused by use of ElectionModule instaed of ElectionModuleSatellite, and calling different init method). This difference should be reflected in the structure of the cannonfiles (ie by having mothership cannonfile `include` from the `satellite` unless there is some explicit reason why that is not possible)

dbeal-eth:
regarding the choice to use a different address for satellites and mothership deployment, it shouldn't be necessary becuase both the satellites and mothership should always share the same initialmodulebundle. so please revert the changes which would allow for address of the system to be different on different networks, we can preserve the optimization/simplicity of having the same address on every network for now :smile:

PR #1835, opened 2023-09-22
More Audit fixes

leomassazza:
- [x] [5.4.1 liquidateFlagged() could run out of gas](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#541-liquidateflagged-could-run-out-of-gas)
- [x] [5.4.2 Perps super market cannot re-register to a new core deployment](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#542-perps-super-market-cannot-re-register-to-a-new-core-deployment)
- [x] [5.4.3 Changing the spot market does not ensure all balances on the old market are settled](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#543-changing-the-spot-market-does-not-ensure-all-balances-on-the-old-market-are-settled)
- [x] [5.4.4 Setting some configuration values should ensure the value is not logically incoherent with its purpose](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#544-setting-some-configuration-values-should-ensure-the-value-is-not-logically-incoherent-with-its-purpose)
- [x] [Gas optimisations](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#gas-optimisations)
- [x] [Unused Variables](https://gist.github.com/iosiro-security/3be89b4425d6e33ef7e025a6c1e25231#unused-variables)
- [x] Use `PerpsMarketFactoryModule.name()`
- [x] Rename `marketId` to `synthMarketId` in `PerpsAccount.deductFromAccount()` to improve readability.
- [x] fix tests

### DIFF
--- a/markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
+++ b/markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
@@ -9,6 +9,7 @@ import {GlobalPerpsMarketConfiguration} from "../storage/GlobalPerpsMarketConfig
 import {GlobalPerpsMarket} from "../storage/GlobalPerpsMarket.sol";
 import {IGlobalPerpsMarketModule} from "../interfaces/IGlobalPerpsMarketModule.sol";
 import {OwnableStorage} from "@synthetixio/core-contracts/contracts/ownership/OwnableStorage.sol";
+import "@synthetixio/core-contracts/contracts/errors/AddressError.sol";
 
 /**
  * @title Module for global Perps Market settings.
@@ -141,7 +142,7 @@ contract GlobalPerpsMarketModule is IGlobalPerpsMarketModule {
         }
 
         if (referrer == address(0)) {
-            revert InvalidReferrerAddress(referrer);
+            revert AddressError.ZeroAddress();
         }
 
         GlobalPerpsMarketConfiguration.load().referrerShare[referrer] = shareRatioD18;
--- a/markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
+++ b/markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
@@ -69,6 +69,10 @@ contract GlobalPerpsMarketModule is IGlobalPerpsMarketModule {
         uint256 maxLiquidationRewardUsd
     ) external override {
         OwnableStorage.onlyOwner();
+        if (minLiquidationRewardUsd > maxLiquidationRewardUsd) {
+            revert InvalidLiquidationRewardGuards(minLiquidationRewardUsd, maxLiquidationRewardUsd);
+        }
+
         GlobalPerpsMarketConfiguration.Data storage store = GlobalPerpsMarketConfiguration.load();
         store.minLiquidationRewardUsd = minLiquidationRewardUsd;
         store.maxLiquidationRewardUsd = maxLiquidationRewardUsd;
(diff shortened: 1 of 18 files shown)


markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -55,18 +55,51 @@ contract LiquidationModule is ILiquidationModule, IMarketEvents {
...
-        for (uint i = 0; i < liquidatableAccounts.length; i++) {
+        uint numberOfAccountsToLiquidate = maxNumberOfAccounts > liquidatableAccounts.length
+            ? liquidatableAccounts.length
+            : maxNumberOfAccounts;
+
+        for (uint i = 0; i < numberOfAccountsToLiquidate; i++) {
             uint128 accountId = liquidatableAccounts[i].to128();
             liquidationReward += _liquidateAccount(PerpsAccount.load(accountId));
         }
     }
```
sunnyvempati, line 74:
```suggestion
        for (uint i = 0; i < liquidatableAccounts.length; i++) {
            uint128 accountId = liquidatableAccounts[i].to128();
            liquidationReward += _liquidateAccount(PerpsAccount.load(accountId));
            
            maxNumberOfAccounts--;
            if (maxNumberOfAccounts == 0) break;
        }
    }
```

nit pik, just another approach

could also use `MathUtil.min` btw

markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -41,37 +41,23 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
...
 
         uint128 perpsMarketId;
         if (factory.perpsMarketId == 0) {
-            perpsMarketId = factory.synthetix.registerMarket(address(this));
+            setSynthetix(synthetix);
+            setSpotMarket(spotMarket);
+            factory.name = marketName;
+
+            perpsMarketId = synthetix.registerMarket(address(this));
             factory.perpsMarketId = perpsMarketId;
```
sunnyvempati, line 61: thoughts on moving all this to `factory.initialize(name, synthetix, spotMarket)` func that returns the `perpsMarketId`?

also in omnibus, this initialize will get re-run and this block will get skipped as expected since the market id exists.  so name will not be able to be set.  maybe add a `setPerpsMarketName` function?

sunnyvempati (approved):
couple notes, most are non blocking but i think adding a way to change the name of the perps market is probably good.  thanks!

markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
```diff
@@ -69,6 +70,10 @@ contract GlobalPerpsMarketModule is IGlobalPerpsMarketModule {
         uint256 maxLiquidationRewardUsd
     ) external override {
         OwnableStorage.onlyOwner();
+        if (minLiquidationRewardUsd > maxLiquidationRewardUsd) {
+            revert InvalidLiquidationRewardGuards(minLiquidationRewardUsd, maxLiquidationRewardUsd);
```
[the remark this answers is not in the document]
leomassazza: oh, interesting. Like it!

markets/perps-market/contracts/modules/LiquidationModule.sol
leomassazza: if there isn't any race condition you identify, I prefer mine approach (but using `MathUtil.min`) becase uses less gas.

PR #1843, opened 2023-09-28
Perps Subgraph: cover all the remaining handlers with tests

noisekit:
At least one bug found so far:
- `handleMarketUpdated` must update existing Market and not create a new one unconditionally


bachstatter (approved):
Nice!

PR #1848, opened 2023-09-30
Use accumulated liquidation rewards for margin requirements

sunnyvempati:
- previously we were using `GlobalPerpsMarketConfiguration.liquidationReward()` which compares liquidation rewards with the global min/max to determine margin requirements.  The accumulated liquidation rewards are a more accurate representation of potential rewards than the global min/max which are subject to liquidation windows.
- Note: if the accumulated rewards are lower than the min liquidation reward, then we should use the min liquidation reward to avoid having LPs pay the difference in margin when a small accnt is liquidated


leomassazza (changes requested):
the min should still be used for super small positions.

leomassazza (approved):
LGTM

PR #1849, opened 2023-10-02
Main subgraph

noisekit:
1. fix publishing scripts for subgraph
2. fix market snapshot calculations
3. refactor market create, deposit and withdrawal tests
4. Remove call to CoreProxy to get market debt data as this is impossible with ERC7412, but also does not make sense as the data is "current" anyway and is not usable for subgraph reprocessing.


markets/perps-market/subgraph/package.json
```diff
@@ -22,7 +22,7 @@
     "@graphprotocol/graph-ts": "^0.31.0",
     "@usecannon/cli": "2.8.1",
     "ethers": "^5.7.2",
-    "matchstick-as": "^0.5.2",
+    "matchstick-as": "0.6.0-beta.2",
```
jmzwar, line 25: This is stable?

jmzwar (approved):

PR #1856, opened 2023-10-05
Add Cancel order

leomassazza:


leomassazza:
Self PR. 
- Need to add a test to check the keeper got paid and the account was deducted the keeper fee

markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol
```diff
@@ -0,0 +1,237 @@
...
+        runtime.settlementReward = settlementStrategy.settlementReward;
+        runtime.sizeDelta = asyncOrder.request.sizeDelta;
+
+        // check if account is flagged
+        GlobalPerpsMarket.load().checkLiquidation(runtime.accountId);
+
+        // Validate Order still exists and is valid
+        if (runtime.sizeDelta == 0) {
+            revert AsyncOrder.ZeroSizeOrder();
+        }
```
sunnyvempati, line 118: i think this and the check for within settlement window can be grouped into a `asyncOrder.validateCancellation()` function.  

- there has to be a pending order
- and the order needs to be within settlement window

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -505,4 +502,16 @@ library AsyncOrder {
         // this is the required margin for the new position (minus any order fees)
         return requiredMarginForNewPosition + requiredLiquidationRewardMargin;
     }
+
+    /**
+     * @notice Checks if the fill price exceeds the acceptable price set at submission.
+     */
+    function acceptablePriceExceeded(
```
sunnyvempati, line 509: nice

sunnyvempati (changes requested):

markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol
```diff
@@ -0,0 +1,237 @@
...
+    using Position for Position.Data;
+    using SafeCastU256 for uint256;
+    using SafeCastI256 for int256;
+
+    int256 private constant PRECISION = 18;
+
+    /**
+     * @inheritdoc IAsyncOrderCancelModule
+     */
+    function cancelOrder(uint128 accountId) external view {
```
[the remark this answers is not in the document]
leomassazza: this check can only be done in cancelPythOrder (since we need the pyth price time). And is done here https://github.com/Synthetixio/synthetix-v3/blob/c44a0ff3bdfeac7cfc085413362b4d7ccbfdbcc5/markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol#L209-L210

markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol
```diff
@@ -0,0 +1,237 @@
...
+
+        PerpsAccount.Data storage account = PerpsAccount.load(runtime.accountId);
+
+        bool isEligible;
+        (isEligible, runtime.currentAvailableMargin, , , , ) = account.isEligibleForLiquidation();
+
+        // Check order is not elegible for liqudiation
+        if (isEligible) {
+            revert PerpsAccount.AccountLiquidatable(runtime.accountId);
+        }
```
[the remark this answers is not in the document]
leomassazza: yeah, had a discussion with myself around it and went to the safest side, but agree. It's not crucial here. 
The only drawback is that if the account IS liquidatable, we are using a fraction of the margin that should go to the liquidator/LPs to pay for the cancellation fee.

markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol
leomassazza: as mentioned before, the settlement window was checked when parsing the pyth price some function calls above

markets/perps-market/contracts/modules/AsyncOrderCancelModule.sol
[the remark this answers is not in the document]
sunnyvempati: we should be checking that the block timestamp is within the settlement window otherwise the price deviation reward isn't eligible.

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -369,6 +368,53 @@ library AsyncOrder {
...
+    ) internal view returns (uint256 fillPrice) {
+        SimulateDataRuntime memory runtime;
+        runtime.sizeDelta = order.request.sizeDelta;
+        runtime.accountId = order.request.accountId;
+        runtime.marketId = order.request.marketId;
+
+        // Validate Order still exists and is valid
+        if (runtime.sizeDelta == 0) {
+            revert AsyncOrder.ZeroSizeOrder();
+        }
```
sunnyvempati, line 393: the `loadValid` on AsyncOrder takes care of this so this is not needed

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -369,6 +368,53 @@ library AsyncOrder {
...
+
+        checkWithinSettlementWindow(order, strategy);
```
sunnyvempati, line 395: ok yeah this is what i mentioned above, as long as this is there, we should be fine

sunnyvempati (approved):
nice.  i think you can get rid of the 0 size check but otherwise lgtm!  send note to Iosiro 🤙🏽

PR #1858, opened 2023-10-05
Fix tests

leomassazza:


sunnyvempati (approved):
have you seen flakiness because of this?  weird

PR #1859, opened 2023-10-06
fix reward distribution module overflow

dbeal-eth:
caused by data type size 128 instead of 256 in a big number multiplication operation


jmzwar (changes requested):
Invalid character line 153.

Also can't divide a uint256 by int32

dbeal-eth:
> Invalid character line 153.
> 
> Also can't divide a uint256 by int32

you know this is a draft right?

also it would probably be faster if you got far enough to find these errors to just push the fixes yourself? you do have permission :laughing:

jmzwar:
> > Invalid character line 153.
> > Also can't divide a uint256 by int32
> 
> you know this is a draft right?
> 
> also it would probably be faster if you got far enough to find these errors to just push the fixes yourself? you do have permission 😆

All good. I'm adding some tests as well :)

jmzwar (approved):

PR #1860, opened 2023-10-06
Variable Staleness Tolerance

noahlitvin:
The Why:

Introducing staleness tolerance functionality to our price feeds will allow us to have more control over _when_ to use a stricter staleness tolerance.  What we've noticed during the testnet competition was that the price updates happen way too frequently and the cost is must too high due to the call data size that needs to be propagated to L1 which is subject to L1 gas.

Solution:
1. To solve this, we're introducing a new configurable parameter (one per market): `strictStalenessTolerance`

2. Changes to feedId oracle manager node functionality
i. The node that is configured for the price of a market (feedId) will be initialized with some default staleness tolerance via parameters (during creation of the node).  
ii. This staleness tolerance can be overriden if a different staleness tolerance is sent in via runtimeArgs.

Based on the above:
- Liquidations will use the `strictStalenessTolerance` which means to get prices of the perps markets, the feedId node will be processed with runtime with the configured staleness tolerance which forces the price to be much fresher than the default tolerance.

This change also propagated to the spot market since collateral prices using pyth oracles will also need to be told if it should use a stricter tolerance.  Consider during liquidations, if a strict staleness tolerance is used to determine perps markets, the same is required to determine the trader's account balance comprised of synths.  

Notable Interface Changes:

All spot market quoting functions now require type of tolerance to use (i.e strict, default):
```
function quoteSellExactIn(
        uint128 marketId,
        uint256 synthAmount,
        Price.Tolerance stalenessTolerance
    ) external view returns (uint256 returnAmount, OrderFees.Data memory fees);
```

This allows the perps market, during quote request of a particular synth, to send the tolerance it accepts for the price.  For all instances, except liquidation, the default is used.  Otherwise, during liquidation, the strict tolerance is used.

Staleness Circuit Breaker Node functionality:
- The circuit breaker can be created with a default staleness tolerance, but also now accepts runtime args that will override the tolerance if one is sent in.
- The node also excepts at least one parent.  
  - If 1 parent and tolerance is exceeded, reverts with the default `StalenessToleranceExceeded()`
  - If 2 parents and tolerance is exceeded, the second parent is processed which causes the `OracleDataRequired` error (which leads to the erc7412 orchestration)


protocol/oracle-manager/contracts/modules/NodeModule.sol
```diff
@@ -214,7 +216,7 @@ contract NodeModule is INodeModule {
             nodeDefinition.nodeType == NodeDefinition.NodeType.STALENESS_CIRCUIT_BREAKER
         ) {
             //check if parents are processable
-            _processParentNodeOutputs(nodeDefinition);
+            _processParentNodeOutputs(nodeDefinition, new bytes32[](0), new bytes32[](0));
```
leomassazza, line 219: this will fail if the parent node does a check on the validity of runtimeParams

markets/spot-market/contracts/modules/WrapperModule.sol
```diff
@@ -67,7 +67,7 @@ contract WrapperModule is IWrapperModule {
         (amountToMint, fees, config) = MarketConfiguration.quoteWrap(
             marketId,
             wrapAmountD18,
-            Price.getCurrentPrice(marketId, Transaction.Type.WRAP)
+            Price.getCurrentPrice(marketId, Transaction.Type.WRAP, true)
```
leomassazza, line 70: please, use a constant to give meaning to the `true` like 
`bool private constant _USE_STRICT_TOLERANCE = true;`

markets/spot-market/contracts/modules/WrapperModule.sol
```diff
@@ -123,7 +123,7 @@ contract WrapperModule is IWrapperModule {
         (returnCollateralAmountD18, fees, config) = MarketConfiguration.quoteUnwrap(
             marketId,
             unwrapAmount,
-            Price.getCurrentPrice(marketId, Transaction.Type.UNWRAP)
+            Price.getCurrentPrice(marketId, Transaction.Type.UNWRAP, true)
```
leomassazza, line 126: same here

markets/spot-market/contracts/modules/SpotMarketFactoryModule.sol
```diff
@@ -106,7 +106,7 @@ contract SpotMarketFactoryModule is ISpotMarketFactoryModule, AssociatedSystemsM
     function reportedDebt(
         uint128 marketId
     ) external view override returns (uint256 reportedDebtAmount) {
-        uint256 price = Price.getCurrentPrice(marketId, Transaction.Type.SELL);
+        uint256 price = Price.getCurrentPrice(marketId, Transaction.Type.SELL, false);
```
leomassazza, line 109: use constant to give meaning to the false

markets/spot-market/contracts/modules/AtomicOrderModule.sol
```diff
@@ -32,7 +32,7 @@ contract AtomicOrderModule is IAtomicOrderModule {
         spotMarketFactory.validateMarket(marketId);
 
         MarketConfiguration.Data storage config;
-        uint256 price = Price.getCurrentPrice(marketId, Transaction.Type.BUY);
+        uint256 price = Price.getCurrentPrice(marketId, Transaction.Type.BUY, true);
```
leomassazza, line 35: dito

markets/spot-market/contracts/modules/AsyncOrderSettlementModule.sol
```diff
@@ -53,7 +53,7 @@ contract AsyncOrderSettlementModule is IAsyncOrderSettlementModule {
                 _settleOrder(
                     marketId,
                     asyncOrderId,
-                    Price.getCurrentPrice(marketId, asyncOrderClaim.orderType),
+                    Price.getCurrentPrice(marketId, asyncOrderClaim.orderType, true),
```
leomassazza, line 56: dito

markets/spot-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -57,11 +58,12 @@ contract AsyncOrderModule is IAsyncOrderModule {
             amountEscrowed = amountProvided;
         } else if (orderType == Transaction.Type.ASYNC_SELL) {
             // Get the dollar value of the provided synths
-            uint256 usdAmount = Price.synthUsdExchangeRate(
+            uint256 currentPrice = Price.getCurrentPrice(
                 marketId,
-                amountProvided,
-                Transaction.Type.ASYNC_SELL
+                Transaction.Type.ASYNC_SELL,
+                true
```
leomassazza, line 64: dito

markets/spot-market/test/common/bootstrapSynthMarkets.ts
```diff
@@ -34,19 +36,19 @@ export function bootstrapSynthMarkets(
...
+      buyAggregator: MockPythExternalNode,
       sellNodeId: string,
-      sellAggregator: AggregatorV3Mock,
+      sellAggregator: MockPythExternalNode,
       synthAddress: string,
       synth: SynthRouter;
 
     before('create price nodes', async () => {
-      const buyPriceNodeResult = await createOracleNode(
+      const buyPriceNodeResult = await createPythNode(
```
leomassazza, line 46: love it!

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
```diff
@@ -1,15 +1,24 @@
...
+import {NodeOutput} from "@synthetixio/oracle-manager/contracts/storage/NodeOutput.sol";
+import {NodeDefinition} from "@synthetixio/oracle-manager/contracts/storage/NodeDefinition.sol";
+import {SafeCastI256} from "@synthetixio/core-contracts/contracts/utils/SafeCast.sol";
+import {GlobalPerpsMarketConfiguration} from "./GlobalPerpsMarketConfiguration.sol";
 
 /**
  * @title Main factory library that registers perps markets.  Also houses global configuration for all perps markets.
  */
 library PerpsMarketFactory {
+    using GlobalPerpsMarketConfiguration for GlobalPerpsMarketConfiguration.Data;
```
leomassazza, line 18: do we need those added usings and imports?

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -56,6 +56,11 @@ library GlobalPerpsMarketConfiguration {
          * @notice If set to a larger number (larger than number of collaterals enabled) it means is unlimited.
          */
         uint128 maxCollateralsPerAccount;
+        /**
+         * @dev each collateral used has a strict staleness tolerance that is used when dealing with liquidations since we want the most recent price.
+         * @dev this mapping has the configured tolerance for each collateral.
+         */
+        mapping(uint128 => uint) collateralStalenessTolerances;
```
leomassazza, line 63: not used. It can be removed

leomassazza: and the setter/getter + roll back the setConfiguration to setMax (since is the only configuration)

protocol/oracle-manager/contracts/modules/NodeModule.sol
sunnyvempati: runtime args shouldn't be part of `isValid`, only the configured parameters.

leomassazza (approved):

.circleci/config.yml
```diff
@@ -672,7 +672,7 @@ workflows:
 
       - test-contracts:
           name: "test-perps-market"
-          requires: [build-testable]
+          requires: [test-spot-market, build-testable]
```
noisekit, line 675: There should not be dependency between tests, it is not a local monoreo and all tests are isolated. This dependency only increases build time without any benefits. Please remove this.

noahlitvin:
Closing in favor of #1904

PR #1862, opened 2023-10-09
Check Pyth publish itme on our contracts

leomassazza:


sunnyvempati (approved):
would double check with @noahlitvin if we need a SIP for this since spot market code is on mainnet.  not sure since its just a check but 🤷🏽 

iosiro should review though

markets/perps-market/contracts/utils/OffchainUtil.sol
```diff
@@ -56,6 +66,20 @@ library OffchainUtil {
...
 
         IPythVerifier.PriceFeed memory pythData = priceFeeds[0];
+
+        // Double check publishTime is valid
+        if (
+            pythData.price.publishTime < asyncOrder.settlementTime ||
+            pythData.price.publishTime >
+            asyncOrder.settlementTime + settlementStrategy.priceWindowDuration
+        ) {
+            revert PythPriceOutsideSettlementWindow(
```
bachstatter, line 76: Me and david chatted about this for l1perps. I was saying that if we cant trust pyth to check validity window how can we trust that the publish time is correct?

PR #1870, opened 2023-10-12
Add CollateralDeducted event

leomassazza:


markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
```diff
@@ -140,7 +102,17 @@ contract AsyncOrderSettlementModule is IAsyncOrderSettlementModule, IMarketEvent
         // since margin is deposited, as long as the owed collateral is deducted
         // fees are realized by the stakers
         if (runtime.amountToDeduct > 0) {
-            perpsAccount.deductFromAccount(runtime.amountToDeduct);
+            (uint128[] memory deductedSynthIds, uint256[] memory deductedAmount) = perpsAccount
+                .deductFromAccount(runtime.amountToDeduct);
+            for (uint256 i = 0; i < deductedSynthIds.length; i++) {
+                if (deductedAmount[i] > 0) {
```
sunnyvempati, line 108: ~~this should never be less than 0?  i'd think the `deductFromAccount` would never return a synth id with 0 if nothing was deducted from it~~

i see what's happening in the impl.  ignore me

sunnyvempati (approved):
nice, this looks great 👍🏽

PR #1872, opened 2023-10-13
Fix spot market subgraph

noisekit:
1. we cannot call contract methods from subgraph processor.
2. if we need some extra data it needs to be in the event. Or we should listen to other events and update records accordingly
3. some events already had necessary entities included so I reused them
4. Some events did not have relevant entity included - so added todo (in case UI still relies on such data)
5. all the handlers now have basic tests coverage

### DIFF
--- a/markets/perps-market/subgraph/codegen.sh
+++ b/markets/perps-market/subgraph/codegen.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+export CANNON_IPFS_URL="https://ipfs.synthetix.io"
 
 codegen() {
   namespace=$1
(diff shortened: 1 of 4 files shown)


markets/spot-market/subgraph/optimism-mainnet/handleOrderCommitted.ts
```diff
@@ -16,13 +15,15 @@ export function handleOrderCommitted(event: OrderCommitted): void {
...
-  );
+  order.status = 'Commited';
+
+  // TODO: order claim may need to be added to OrderCommitted event again (used to be there)
+  // let claim = event.params.asyncOrderClaim;
+  // order.amountEscrowed = claim.amountEscrowed;
+  // order.settlementStrategyId = claim.settlementStrategyId;
+  // order.settlementTime = claim.settlementTime;
+  // order.minimumSettlementAmount = claim.minimumSettlementAmount;
+  // order.settledAt = claim.settledAt;
```
bachstatter, line 26: Requires someone with more context than me to review this properly. I would think we should define what's  required data and then remove unneeded things from the schema and add required things to the event?

PR #1875, opened 2023-10-18
Test OrderSettled event

leomassazza:


sunnyvempati (changes requested):
tbh the funding test looks a lot more confusing now.  i would really prefer if this test isn't messed with so we can isolate testing funding and not have other items being tested as part of it.

could you create a separate test that just does the following?  it's fine if we hardcode the funding value for this.  all we care about is that accrued funding is returned in the order settled event:

- open 1 order
- settle it
- check that the order settled has the accrued funding returned. (hardcoded is fine)

sunnyvempati (approved):

PR #1879, opened 2023-10-18
parsePriceFeedUpdatesUnique and new pyth mocks

bachstatter:
So I started this PR just to update the PythMock to add the new [parsePriceFeedUpdatesUnique](https://docs.pyth.network/evm/parse-price-feed-updates-unique
) which all our markets want to use.
Instead of just adding it to the mock I checked how pyth updated their mocks from [pyth github repo]( https://github.com/pyth-network/pyth-crosschain/blob/main/target_chains/ethereum/sdk/solidity/AbstractPyth.sol)
It seemed to be some useful updates around comments, and the ability to pass in a previous timestamp to `createPriceFeedUpdateData`.

So I updated our mocks, then I noticed that we had the wrong signature for `getUpdateFee` in both our mocks and in our "real" PythVerifier contracts. It seems to be working ok, but I updated it to have the correct signature. 
This required quite a few updates in tests, but mostly straightforward.

I also updated spot and perps to use `parsePriceFeedUpdatesUnique` instead of `parsePriceFeedUpdates`.
Maybe this should be happening in a different PR?


bachstatter:
This probably should've been a draft, some issues to fix still as tests expect the old mock. The real contracts should be backwards compatible so I was hoping (gambling) the mock would just work.

markets/perps-market/contracts/utils/OffchainUtil.sol
```diff
@@ -44,11 +44,11 @@ library OffchainUtil {
         updateData[0] = result;
 
         IPythVerifier verifier = IPythVerifier(settlementStrategy.priceVerificationContract);
-        uint256 msgValue = verifier.getUpdateFee(1);
+        uint256 msgValue = verifier.getUpdateFee(updateData);
```
bachstatter, line 47: @leomassazza  or @sunnyvempati  I'd expect the keeper that settles the transaction to pass along the right fee on `msg.value`. Why are we looking it up?

bachstatter, line 47:
```suggestion
        uint256 msgValue = msg.value
```
To be clear, I think this would work just fine

markets/perps-market/contracts/interfaces/external/IPythVerifier.sol
```diff
@@ -82,8 +82,27 @@ interface IPythVerifier {
...
+        uint64 minPublishTime,
+        uint64 maxPublishTime
+    ) external payable returns (PriceFeed[] memory priceFeeds);
+
     /// @notice Returns the required fee to update an array of price updates.
-    /// @param updateDataSize Number of price updates.
+    /// @param updateData Array of price update data.
     /// @return feeAmount The required fee in Wei.
-    function getUpdateFee(uint updateDataSize) external view returns (uint feeAmount);
+    function getUpdateFee(bytes[] calldata updateData) external view returns (uint feeAmount);
```
leomassazza, line 107: Thanks. Pyth updated this interface and we were using an old one.

markets/perps-market/contracts/utils/OffchainUtil.sol
leomassazza: Mmmm... for this, it can be tricky since we can receive founds (msg.value) not only to be sent to Pyth verifier, if we do that we are diverging all the funds to this call.

leomassazza (approved):
LGTM. Answered your questions

bachstatter:
Hmm yea I think we should be using the pyth mock from cannon instead. So I'm leaning towards closing this. But it would be good if perps and spots starts using the new method

leomassazza:
Note this branch was merged into https://github.com/Synthetixio/synthetix-v3/tree/feat/pyth-erc7412-wrapper

PR #1888, opened 2023-10-27
optimize contracts

dbeal-eth:
wont affect proxy contracts becuase they are exclusively locked by the common config


leomassazza:
Love it!

leomassazza (approved):

PR #1891, opened 2023-10-30
Oracle manager changes to support pyth offchain natively

sunnyvempati:
- Changed functionality to check node validity to call `isValid` on parent nodes instead of processing the parent nodes.
- Created new node `PythOffchainLookupNode` that will revert with the `OracleDataRequired` error.
- staleness tolerance node now accepts the nodeDefintion and it figures out the order in which the parents are processed, so it can call `PythOffchainLookupNode` or the second node process when it wants to revert.


protocol/oracle-manager/contracts/nodes/pyth/PythOffchainLookupNode.sol
```diff
@@ -0,0 +1,82 @@
...
+        bytes32[] memory runtimeValues
+    ) internal pure returns (NodeOutput.Data memory) {
+        (address pythAddress, bytes32 priceId, uint256 stalenessTolerance) = abi.decode(
+            parameters,
+            (address, bytes32, uint256)
+        );
+
+        for (uint256 i = 0; i < runtimeKeys.length; i++) {
+            if (runtimeKeys[i] == "stalenessTolerance") {
+                stalenessTolerance = uint256(runtimeValues[i]);
```
leomassazza, line 30:
```suggestion
                // solhint-disable-next-line numcast/safe-cast
                stalenessTolerance = uint256(runtimeValues[i]);
```

leomassazza (approved):
PGTM. Lint is failing, added a suggestion with the fix

PR #1902, opened 2023-11-12
Interest rate

sunnyvempati:
- introduce interest rate.
Description: TBD


noisekit:
@sunnyvempati FYI I've rebased and force-pushed

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -552,6 +553,7 @@ library PerpsAccount {
                 marketId: marketId,
                 latestInteractionPrice: price.to128(),
                 latestInteractionFunding: perpsMarket.lastFundingValue.to128(),
+                latestInterestAccrued: 0,
```
sunnyvempati, line 556: .

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -383,10 +387,19 @@ library PerpsMarket {
     function marketDebt(Data storage self, uint price) internal view returns (int) {
         // all positions sizes multiplied by the price is equivalent to skew times price
         // and the debt correction accumulator is the  sum of all positions pnl
-        int traderUnrealizedPnl = self.skew.mulDecimal(price.toInt());
-        int unrealizedFunding = self.skew.mulDecimal(calculateNextFunding(self, price));
+        int positionPnl = self.skew.mulDecimal(price.toInt());
+        int fundingPnl = self.skew.mulDecimal(calculateNextFunding(self, price));
```
sunnyvempati, line 391: marketPositionPnl, marketFundingPnl

markets/perps-market/test/integration/Position/InterestRate.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        const trader2Oi = wei(Math.abs(newPositionSize)).mul(_BTC_PRICE).mul(_BTC_LOCKED_OI_RATIO);
+
+        const expectedInterest = trader2Oi
+          .mul(wei(await systems().PerpsMarket.interestRate()))
+          .mul(normalizedTime);
+        assertBn.near(chargedInterest, expectedInterest.toBN(), bn(0.00001));
+      });
+    });
+  });
+});
```
sunnyvempati, line 226: change delegated collateral from the core side then call `update` external.

protocol/synthetix/test/common/stakedPool.ts
```diff
@@ -59,6 +59,7 @@ export const createStakedPool = (
 
   before('stake', async () => {
     const [, staker] = r.signers();
+    console.log('STAKEDAMT', stakedAmount);
```
sunnyvempati, line 62: .

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
```diff
@@ -93,4 +93,24 @@ library PerpsMarketFactory {
...
+        uint256 activeMarketsLength = activeMarkets.length();
+        for (uint i = 1; i <= activeMarketsLength; i++) {
+            uint128 marketId = activeMarkets.valueAt(i).to128();
+
+            accumulatedMinimumCredit += PerpsMarket.requiredCredit(marketId);
+        }
+    }
+
+    function utilizationRate(Data storage self) internal view returns (uint256 utilization) {
+        uint256 withdrawableUsd = self.synthetix.getWithdrawableMarketUsd(self.perpsMarketId);
```
bachstatter: I thought this guy was the right method to call to get the delgegatedCollateralValue directly?
    /**
     * @notice Returns the total snxUSD value of the collateral for the specified market.
     * @param marketId The id of the market whose collateral is being queried.
     * @return valueD18 The market's total snxUSD value of collateral, denominated with 18 decimals of precision.
     */
    function getMarketCollateral(uint128 marketId) external view returns (uint256 valueD18);

markets/perps-market/contracts/utils/BigNumber.sol
```diff
@@ -0,0 +1,52 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.11 <0.9.0;
+
+/**
+ * @title Interest Rate
+ */
+library BigNumber {
```
bachstatter, line 7: Is this used?

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -63,6 +65,30 @@ library GlobalPerpsMarket {
         }
     }
 
+    function utilizationRate(
+        Data storage self
+    ) internal view returns (uint128 rate, uint256 delegatedCollateralValue, uint256 lockedCredit) {
+        uint256 withdrawableUsd = PerpsMarketFactory.totalWithdrawableUsd();
+        delegatedCollateralValue = withdrawableUsd - totalCollateralValue(self);
```
bachstatter, line 72: I was planning to use `getMarketCollateral`
```
    /**
     * @notice Returns the total snxUSD value of the collateral for the specified market.
     * @param marketId The id of the market whose collateral is being queried.
     * @return valueD18 The market's total snxUSD value of collateral, denominated with 18 decimals of precision.
     */
    function getMarketCollateral(uint128 marketId) external view returns (uint256 valueD18);
```
Is there a reason that shouldn't be used?

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -238,17 +239,20 @@ library PerpsMarket {
...
+        (, int pricePnl, , int fundingPnl, , ) = oldPosition.getPnl(currentPrice);
 
         int sizeDelta = newPositionSize - oldPositionSize;
         int fundingDelta = calculateNextFunding(self, currentPrice).mulDecimal(sizeDelta);
         int notionalDelta = currentPrice.toInt().mulDecimal(sizeDelta);
 
         // update the market debt correction accumulator before losing oldPosition details
         // by adding the new updated notional (old - new size) plus old position pnl
-        self.debtCorrectionAccumulator += fundingDelta + notionalDelta + totalPositionPnl;
+        self.debtCorrectionAccumulator += fundingDelta + notionalDelta + pricePnl + fundingPnl;
```
bachstatter, line 250: Should interestPaid not be included in the debtCorrectionAcc?

sunnyvempati: this debt correction accumulator is per market whereas the interest rate for us is at the global level across all markets since the utilization rate is based on the supermarket's utilization.

the unrealized interest rate doesn't need to be reported since it's not "debt".  it's credit and when the credit is realized (when trades close) it gets deposited for LPs to gain.  think of it as a fee essentially when closing your trade except that since its continuous, it eats up your margin.

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
sunnyvempati: `getMarketCollateral` is the value of collateral deposited by the market, not the collateral delegated to it.  maybe in your case, you can use this to get the `totalCollateralValue` but you will want to subtract that out from whatever `getWithdrawableUsd` returns which includes everything including whatever market has deposited.

bachstatter: hmm I think getMarketCollateral is the delegated collateral from pools.
Core have `getMarketCollateralValue` and `getMarketCollateralAmount` which return the collateral deposited by the maket

bachstatter:
```
 function getMarketCollateral(uint128 marketId) external view override returns (uint256) {
        return Market.load(marketId).poolsDebtDistribution.totalSharesD18;
    }
```
I'm not totally sure how it works, but it's getting the data from pool's debt distribution

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: That does make sense!

bachstatter: Hmm although shouldn't that credit be reported as profit/negative debt and give the market more liquidity?

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
bachstatter: I think doing it they way you do it:`withdrawableUsd - totalCollateralValue(self);` means we not only get delegated collateral since the `withdrawableUsd` also takes the market's creditCapacity into account. 
Maybe we want that?

It basically means that if the market has a bunch of unrealised gains the the utilisation will go down, as the credit capacity takes those that into account through the reportedDebt

markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
```diff
@@ -246,4 +248,53 @@ contract GlobalPerpsMarketModule is IGlobalPerpsMarketModule {
...
+    function setInterestRateParameters(
+        uint128 lowUtilizationInterestRateGradient,
+        uint128 interestRateGradientBreakpoint,
+        uint128 highUtilizationInterestRateGradient
+    ) external override {
+        OwnableStorage.onlyOwner();
+        GlobalPerpsMarketConfiguration.Data storage store = GlobalPerpsMarketConfiguration.load();
+        store.lowUtilizationInterestRateGradient = lowUtilizationInterestRateGradient;
+        store.interestRateGradientBreakpoint = interestRateGradientBreakpoint;
+        store.highUtilizationInterestRateGradient = highUtilizationInterestRateGradient;
```
leomassazza, line 264: should we check that `highUtilizationInterestRateGradient >= lowUtilizationInterestRateGradient` as a checkpoing? I don't find a reasonable usage of `highUtilizationInterestRateGradient` being low that `lowUtilizationInterestRateGradient`

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -227,32 +235,44 @@ library PerpsMarket {
         uint128 accountId,
         Position.Data memory newPosition
     ) internal returns (MarketUpdate.Data memory) {
+        PositionDataRuntime memory runtime;
         Position.Data storage oldPosition = self.positions[accountId];
 
-        int128 oldPositionSize = oldPosition.size;
-        int128 newPositionSize = newPosition.size;
+        // int128 oldPositionSize = oldPosition.size;
```
leomassazza, line 241: remove rogue commented out vars

markets/perps-market/contracts/storage/InterestRate.sol
```diff
@@ -0,0 +1,104 @@
...
+
+        (
+            uint128 lowUtilizationInterestRateGradient,
+            uint128 interestRateGradientBreakpoint,
+            uint128 highUtilizationInterestRateGradient
+        ) = GlobalPerpsMarketConfiguration.loadInterestRateParameters();
+
+        // if no interest parameters are set, interest rate is 0 and the interest accrued stays the same
+        if (
+            lowUtilizationInterestRateGradient == 0 ||
```
leomassazza, line 48: should be `AND` and not `OR`

markets/perps-market/contracts/storage/InterestRate.sol
```diff
@@ -0,0 +1,104 @@
...
+            interestRateGradientBreakpoint,
+            highUtilizationInterestRateGradient
+        );
+        self.lastTimestamp = block.timestamp;
+
+        return (self.interestRate, self.interestAccrued);
+    }
+
+    function proportionalElapsed(Data storage self) internal view returns (uint128) {
+        return (block.timestamp - self.lastTimestamp).divDecimal(AVERAGE_SECONDS_PER_YEAR).to128();
```
leomassazza, line 72: This needs a review (bug). `block.timestamp` is a normal integer, not 18 digits. 
So, first issue: `divDecimal` shouldn't be used here; and second issue: since it's not 18 decimal based, will return 0 unless delta is 1 year.

protocol/synthetix/test/common/stakedPool.ts
leomassazza: rogue console.log

markets/perps-market/contracts/storage/Position.sol
```diff
@@ -72,7 +84,31 @@ library Position {
...
+        chargedInterest = interestAccrued(self, price);
+
+        totalPnl = pricePnl + accruedFunding - chargedInterest.toInt();
+    }
+
+    function interestAccrued(
+        Data storage self,
+        uint price
+    ) internal view returns (uint256 chargedInterest) {
+        uint nextInterestAccrued = InterestRate.load().calculateNextInterest();
```
bachstatter, line 97: `calculateNextInterest` does this: `return self.interestAccrued + unrecordedInterest(self);` where `unrecordedInterest` return the raw value based on the stored interestRate. (`self.interestRate.mulDecimalUint128(proportionalElapsed(self)).to256();`)

Is there a reason this deviates from how we do unrecorded funding?
Inside unrecordedInterest I expected to see:
```
uint256 avgUtilizationRate =(self.interestRate + currentInterestRate(...)).divDecimal(DecimalMath.UNIT * 2)
return avgUtilizationRate.mulDecimalUint128(proportionalElapsed(self)).to256();
```
With the idea that the unrecorded funding is the average between the previous funding rate and the current one.

Basically the question comes down to:
If I open a position when utilisation is 1%
And close when it's 3%
With the average we would charge 2%. Whereas they way it's written here I think we would charge 3%.

sunnyvempati: you get charged 3% from the time it changed to 3% to when the trader closed the position, _not_ the average.  

the trader does not get charged 3%, they get charged 1% on their position until the rate goes to 3% (due to another position or LP undelegating), then charged 3% from that time forward till they close.  you can't average it, its time based.

funding is using average for calculating rate across continuous time NOT accrued funding per unit of market.

markets/perps-market/contracts/storage/InterestRate.sol
sunnyvempati: will add comment ty!

markets/perps-market/contracts/storage/InterestRate.sol
sunnyvempati: hmmm, initially i did think if any of these are 0, it should ignore it, but you're right, maybe low could be 0 and the other two could be set which might be pretty common.  changing to AND, thx

markets/perps-market/contracts/modules/GlobalPerpsMarketModule.sol
sunnyvempati: i'd say nice to have but will add it.  if the gradient is lower, then it will always just use whatever the high gradient value is so codewise it won't break but probably good to have some guardrails 👍🏽

leomassazza (approved):

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
bachstatter: We do want creditCapacity included so this is correct

markets/perps-market/contracts/storage/PerpsMarket.sol
bachstatter: No need to include unrealised interest in credit capacity

bachstatter (approved):

PR #1903, opened 2023-11-13
Subgraph parity

jmzwar:


markets/perps-market/subgraph/utils/calculations.ts
```diff
@@ -0,0 +1,35 @@
...
+export function calculateLeverage(size: BigInt, lastPrice: BigInt, margin: BigInt): BigInt {
+  if (size.equals(BigInt.fromI32(0))) return BigInt.fromI32(0);
+  return size.times(lastPrice).div(margin).abs();
+}
+
+export function calculatePnl(lastPrice: BigInt, avgEntryPrice: BigInt, size: BigInt): BigInt {
+  return lastPrice.minus(avgEntryPrice).times(size).div(BigInt.fromI32(10).pow(18));
+}
+
+export function calculateAccruedPnlForReducingPositions(
```
noisekit, line 16: these values should come from events and not calculated in subgraph processing. As perps is not yet on mainnet I strongly suggest extending events and avoid repeating calculations in subgraph.

PR #1904, opened 2023-11-13
Perps andromeda

leomassazza:


markets/spot-market/contracts/modules/AsyncOrderModule.sol
```diff
@@ -57,11 +58,12 @@ contract AsyncOrderModule is IAsyncOrderModule {
             amountEscrowed = amountProvided;
         } else if (orderType == Transaction.Type.ASYNC_SELL) {
             // Get the dollar value of the provided synths
-            uint256 usdAmount = Price.synthUsdExchangeRate(
+            uint256 currentPrice = Price.getCurrentPrice(
```
noisekit, line 61: May I suggest adding `currentPrice` and `usdAmount` to the `OrderCommitted` event, because this data is transient and (almost) impossible to recover retrospectively.

noisekit: This comment is valid for all other places in the code where we rely on current prices and then do mutations. Events should include that price at the time of mutation.

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -510,42 +507,81 @@ library AsyncOrder {
...
+            runtime.accumulatedLiquidationRewards,
+            runtime.maxNumberOfWindows
+        );
 
         // this is the required margin for the new position (minus any order fees)
-        return requiredMarginForNewPosition + requiredLiquidationRewardMargin;
+        return runtime.requiredMarginForNewPosition + runtime.requiredRewardMargin;
+    }
+
+    function getNewPositionsCount(
```
sunnyvempati, line 575: should rename this function cc @leomassazza

markets/spot-market/contracts/modules/AsyncOrderModule.sol
sunnyvempati: this price is valid for selling and it's on commitment, and only used for validation.  on `OrderSettled`, we do emit the price it was settled at which is far more accurate.

sunnyvempati (approved):

noahlitvin (approved):

barrasso (approved):

PR #1905, opened 2023-11-13
Feature/extra events

jmzwar:


markets/perps-market/subgraph/codegen.sh
```diff
@@ -39,4 +39,4 @@ releaseVersion="latest"
 #codegen optimism-mainnet 10 "synthetix-perps-market:$releaseVersion" main
 codegen optimism-goerli 420 "synthetix-perps-market:$releaseVersion" main
 #codegen base-goerli 84531 "synthetix-perps-market:$releaseVersion" main
-codegen base-goerli-competition 84531 "synthetix-perps-market:latest" competition
+codegen base-goerli 84531 "synthetix-perps-market:$releaseVersion" main
```
noisekit, line 42: this is not a correct change.

markets/perps-market/subgraph/subgraph.base-goerli.yaml
```diff
@@ -60,6 +60,8 @@ dataSources:
             uint128,uint256,int256,int256,int128,int128,uint256,uint256,uint256,uint256,indexed
             bytes32,address)
           handler: handleOrderSettled
+        - event: OrderCancelled(indexed uint128, indexed uint128, uint256, uint256, int128, uint256, indexed bytes32, address)
```
noisekit, line 63: subgraphs cannot be updated in the same PR because they rely on deployed and published changes after code changes are already merged and released.

noisekit (changes requested):
remove subgraph changes

markets/perps-market/subgraph/codegen.sh
jmzwar: this is removed from this branch

PR #1914, opened 2023-11-14
Update subgraphs

noisekit:
- [x] Andromeda subgraphs for synthetix, spot market and perps market
- [x] Fix ABI incompatibility for perps
- [x] Refactor generation to use omnibus deployment
- [x] Separate CCI workflows for every auto-generator


jmzwar (approved):
We can probably remove competition configuration.

PR #1917, opened 2023-11-14
add epoch update to satellites

mjlescano:


fritzschoff (approved):
wanna add tests for that?
LGTM

PR #1920, opened 2023-11-15
Fix snapshot contract management

mjlescano:
This PR includes several fixes for snapshot contract usage:

* It was impossible to set an snapshot contract on first epoch (epochIndex `0`).
* When calling `takeVotePowerSnapshot` we weren't validating that the snapshot contract was valid and enabled.

> Keep into account that to easily fix those points, I changed the logic when calling `setSnapshotContract` to make it set the snapshot contract for the current epoch, instead of the coming one. This logic is still valid because this method can only be called during Administration period.

### DIFF
--- a/protocol/governance/contracts/modules/core/SnapshotVotePowerModule.sol
+++ b/protocol/governance/contracts/modules/core/SnapshotVotePowerModule.sol
@@ -63,7 +63,10 @@ contract SnapshotVotePowerModule is ISnapshotVotePowerModule {
         uint128 currentEpoch = Council.load().currentElectionId.to128();
         SnapshotVotePower.Data storage snapshotVotePower = SnapshotVotePower.load(snapshotContract);
 
-        // TODO: check that snapshotVotePower is enabled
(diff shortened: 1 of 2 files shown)


protocol/governance/contracts/modules/core/SnapshotVotePowerModule.sol
```diff
@@ -75,6 +63,7 @@ contract SnapshotVotePowerModule is ISnapshotVotePowerModule {
         uint128 currentEpoch = Council.load().currentElectionId.to128();
         SnapshotVotePower.Data storage snapshotVotePower = SnapshotVotePower.load(snapshotContract);
 
+        // TODO: check that snapshotVotePower is enabled
```
fritzschoff, line 66: why todo? why not doing it in this pr?

fritzschoff (changes requested):

fritzschoff (approved):

fritzschoff:
please fix the lint

PR #1932, opened 2023-11-16
Fix flagger rewards

leomassazza:


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -516,6 +516,7 @@ library PerpsAccount {
             uint128 amountToLiquidate,
             int128 newPositionSize,
             int128 sizeDelta,
+            uint128 positionFlaggedAmount,
```
sunnyvempati, line 519: maybe just return `oldPositionSize`?  `positionFlaggedAmount` is probably unnecessary logic here; can move this to where liquidate is called

markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -160,27 +163,44 @@ contract LiquidationModule is ILiquidationModule, IMarketEvents {
...
                 MarketUpdate.Data memory marketUpdateData
-            ) = account.liquidatePosition(positionMarketId, price);
+            ) = account.liquidatePosition(runtime.positionMarketId, runtime.price);
+
+            // using amountToLiquidate to calculate liquidation reward
+            runtime.liquidationReward = includeFlaggingReward
+                ? PerpsMarketConfiguration
+                    .load(runtime.positionMarketId)
+                    .calculateLiquidationReward(positionFlaggedAmount.mulDecimal(runtime.price))
+                : 0;
```
sunnyvempati, line 185: it might be easier to understand what's happening her if just calculated liquidation rewards here regardless of `includeFlagginReward` and check the flag after.

markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -190,31 +210,16 @@ contract LiquidationModule is ILiquidationModule, IMarketEvents {
...
-            ) {
-                runtime.totalLiquidationRewards += liquidationReward;
-            }
         }
 
         runtime.totalLiquidationCost =
             KeeperCosts.load().getLiquidateKeeperCosts() +
             costOfFlagExecution;
-        if (runtime.totalLiquidated > 0) {
+        if (includeFlaggingReward || runtime.totalLiquidated > 0) {
```
sunnyvempati, line 222: could we rename `includeFlaggingReward` -> `positionFlagged`.  a bit easier to understand imo

sunnyvempati (changes requested):

sunnyvempati (approved):
approving but suggested a few times that might help

markets/perps-market/contracts/storage/PerpsAccount.sol
leomassazza: will maybe rename to something has better meaning here, but the idea is to use that this value is the `abs(oldPositionSize)`, so maybe just that name makes sense: `oldPositionSizeAbs`

markets/perps-market/contracts/modules/LiquidationModule.sol
leomassazza: did a change on the logic (just renaming variables and moving small pieces of code to a different place). Please reassess.

PR #1938, opened 2023-11-22
Perps Andromeda cleanup

leomassazza:
- remove unused imports
- remove unused function
- fix a failing test
- reduce timeout (only extends failing time when the weird `ensure "done()" is called` happens)


sunnyvempati (approved):
fixed the test locally but forgot to push lol

PR #1947, opened 2023-11-29
fix trusted multicall forwarder

dbeal-eth:
with feedback from a call, renamed `allowFailure` to `requireSuccess`

some fo the audit feedback got deleted from a recent merge

take 2


auxiliary/TrustedMulticallForwarder/src/TrustedMulticallForwarder.sol
```diff
@@ -95,7 +95,7 @@ contract TrustedMulticallForwarder is ERC2771Forwarder {
             (result.success, result.returnData) = call.target.call(
                 abi.encodePacked(call.callData, msg.sender)
             );
-            if (requireSuccess && !result.success) {
+            if (!requireSuccess && !result.success) {
```
noisekit, line 98:
```suggestion
            if (requireSuccess && !result.success) {
```
Guess this one shouldn't change

dbeal-eth: should be fixed

PR #1956, opened 2023-12-05
Release v3.3.5

barrasso:


noisekit (changes requested):
release needs to be in master branch, workspace deps  must not be changed.

PR #1963, opened 2023-12-07
Add mocked OVM Gas Oracle

leomassazza:


dbeal-eth:
fyi alternative possible solution proposed on synthetix deployments https://github.com/Synthetixio/synthetix-deployments/pull/99

PR #1970, opened 2023-12-14
test: correct amount of sUSD to trader1 after withdrawal

tommyrharper:
Currently the perps market withdrawal with open position tests just check the internal accounting after withdrawing margin.

I have added this test to also check that the actual `sUSD` is correctly minted to the withdrawing account.

This ensures that `_withdrawMargin` is correctly executed within the call to `modifyCollateral`, not just `updateCollateralAmount` which updates the internal accounting, but doesn't mint any sUSD to the withdrawer.


sunnyvempati (approved):
thanks

PR #1991, opened 2024-02-05
Perps v3.2: Multi-collateral

sunnyvempati:
https://www.notion.so/Perps-v3-2-multi-collateral-spec-7763444dc140495896c5753e8cc07aea?pvs=4


markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -82,11 +82,28 @@ contract PerpsAccountModule is IPerpsAccountModule {
...
     /**
      * @inheritdoc IPerpsAccountModule
      */
-    function totalCollateralValue(uint128 accountId) external view override returns (uint256) {
-        return PerpsAccount.load(accountId).getTotalCollateralValue(PerpsPrice.Tolerance.DEFAULT);
+    function totalCollateralValue(uint128 accountId) external view override returns (uint) {
+        return
+            PerpsAccount.load(accountId).getTotalCollateralValue(
+                PerpsPrice.Tolerance.DEFAULT,
+                false
```
sunnyvempati, line 105: maybe make it a constant?

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -32,10 +34,9 @@ library GlobalPerpsMarketConfiguration {
          */
         mapping(address => uint256) referrerShare;
         /**
-         * @dev mapping of configured synthMarketId to max collateral amount.
-         * @dev USD token synth market id = 0
+         * @dev previously maxCollateralAmounts[collateralId] was used in storage slot
          */
-        mapping(uint128 => uint256) maxCollateralAmounts;
+        mapping(uint128 => uint256) __unused_1;
```
sunnyvempati, line 39: moved to collateral configuration.  the function name for setting max collateral has changed which will force the value to be set again.

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -84,6 +85,14 @@ library GlobalPerpsMarketConfiguration {
          * @dev interest rate gradient applied to utilization after hitting the gradient breakpoint
          */
         uint128 highUtilizationInterestRateGradient;
+        /**
+         * @dev ratio of the collateral liquidation reward. 1e18 is 100%.
+         */
+        uint128 collateralLiquidateRewardRatioD18;
+        /**
+         * @dev reward distributor implementation. This is used as a base to be cloned to distribute rewards to the liquidator.
```
sunnyvempati, line 93: @leomassazza does this comment still hold true?

markets/spot-market/test/SpotMarketFactory.test.ts
```diff
@@ -130,6 +130,27 @@ describe('SpotMarketFactory', () => {
...
+        );
+      });
+
+      it('returns price with valid args', async () => {
+        const buyPrice = await systems()
+          .SpotMarket.connect(marketOwner)
+          .indexPrice(marketId(), 1, 0); // buy feed
+        const sellPrice = await systems()
+          .SpotMarket.connect(marketOwner)
+          .indexPrice(marketId(), 2, 0); // buy feed
```
leomassazza, line 147: comment missing: sell feed

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -202,13 +203,11 @@ library GlobalPerpsMarketConfiguration {
         return notionalValue.mulDecimal(self.collateralLiquidateRewardRatioD18);
     }
 
-    function updateCollateralMax(
+    function updateSupportedCollaterals(
         Data storage self,
         uint128 collateralId,
         uint256 maxCollateralAmount
```
leomassazza, line 209: max is not used anymore. Why?

markets/perps-market/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@synthetixio/perps-market",
-  "version": "3.3.20",
+  "version": "3.3.17",
```
leomassazza, line 3: downgrading?

markets/perps-market/contracts/interfaces/IPerpsAccountModule.sol
```diff
@@ -135,4 +133,18 @@ interface IPerpsAccountModule {
             uint256 requiredMaintenanceMargin,
             uint256 maxLiquidationReward
         );
+
+    /**
+     * @notice Allows anyone to pay an account's debt
+     * @param accountId Id of the account.
+     * @param amount debt amount to pay off
+     */
+    function payDebt(uint128 accountId, uint256 amount) external;
```
kaleb-keny, line 142: Hey @sunnyvempati can you please tell me why this function was designed this way? Is there any blocker of having the debt absorbed by LP's and the collateral as well, instead of having the liquidator repay the debt and accordingly receive the collateral?

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
```diff
@@ -101,4 +107,21 @@ library PerpsMarketFactory {
     function withdrawMarketUsd(Data storage self, address to, uint256 amount) internal {
         self.synthetix.withdrawMarketUsd(self.perpsMarketId, to, amount);
     }
+
+    function transferLiquidatedSynth(
```
kaleb-keny, line 111: Hey @sunnyvempati can you tell me what this function is used for please. afaik no synth gets ever sold, all collaterals are distributed to stakers.

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -186,18 +196,23 @@ library GlobalPerpsMarketConfiguration {
...
         Data storage self,
-        uint128 synthMarketId,
+        uint256 notionalValue
+    ) internal view returns (uint256) {
+        return notionalValue.mulDecimal(self.collateralLiquidateRewardRatioD18);
+    }
+
+    function updateSupportedCollaterals(
+        Data storage self,
+        uint128 collateralId,
```
kaleb-keny, line 208: Is this where we update the collateral parameters too? Like scalar and lower and upper limits of the discounts?

markets/perps-market/contracts/storage/PerpsCollateralConfiguration.sol
```diff
@@ -0,0 +1,158 @@
...
+            uint256 impactOnSkew = amount.divDecimal(skewScale).mulDecimal(self.discountScalar);
+            discount = (
+                MathUtil.max(
+                    MathUtil.min(impactOnSkew, self.lowerLimitDiscount),
+                    self.upperLimitDiscount
+                )
+            );
+        }
+        // first get value of collateral in usd
+        uint128 sellTxnType = 2;
```
kaleb-keny, line 147: I believe this is what we discussed the other day, about updating to use the sport price without any pd impact.

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -113,6 +119,41 @@ library PerpsAccount {
...
+                self.debt += (-leftoverCredit).toUint();
+            }
+        }
+    }
+
+    function isEligibleForMarginLiquidation(
+        Data storage self,
+        PerpsPrice.Tolerance stalenessTolerance
+    ) internal view returns (bool isEligible, int256 availableMargin) {
+        availableMargin = getAvailableMargin(self, stalenessTolerance);
```
kaleb-keny, line 153: Does this update the latest pnl numbers before giving back the number? 
I mean assume someone is elegible for liquidation, but his debt doesn't show, I think we need to update debt number somewhere?

markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -236,6 +272,21 @@ contract LiquidationModule is ILiquidationModule, IMarketEvents {
...
+            ERC2771Context._msgSender() !=
+            PerpsMarketConfiguration.load(runtime.positionMarketId).endorsedLiquidator
+        ) {
+            // Use max of collateral or positions flag rewards
+            uint256 totalCollateralLiquidateRewards = GlobalPerpsMarketConfiguration
+                .load()
+                .calculateCollateralLiquidateReward(totalCollateralValue);
+
+            runtime.totalFlaggingRewards = totalCollateralLiquidateRewards >
+                runtime.totalFlaggingRewards
```
kaleb-keny, line 285: hey @sunnyvempati is there a cap missing, can't seem to find it

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -54,40 +56,64 @@ contract PerpsAccountModule is IPerpsAccountModule {
...
+    function debt(uint128 accountId) external view override returns (uint256 accountDebt) {
+        Account.exists(accountId);
+        PerpsAccount.Data storage account = PerpsAccount.load(accountId);
+
+        accountDebt = account.debt;
+    }
+
+    // 1. call depositMarketUsd and deposit amount directly to core system
+    // 2. look up account and reduce debt by amount
+    // 3. transfer synth to sender
```
kaleb-keny, line 97: Sorry i dont understand here, what synth is sent to sender?

markets/perps-market/contracts/interfaces/ILiquidationModule.sol
```diff
@@ -61,6 +71,14 @@ interface ILiquidationModule {
      */
     function liquidate(uint128 accountId) external returns (uint256 liquidationReward);
 
+    /**
+     * @notice Liquidates an account's margin when no other positions exist.
+     * @dev if available margin is negative and no positions exist, then account margin can be liquidated by calling this function
+     * @param accountId Id of the account to liquidate.
+     * @return liquidationReward total reward sent to liquidator.
+     */
+    function liquidateMarginOnly(uint128 accountId) external returns (uint256 liquidationReward);
```
kaleb-keny, line 80: Is there no need to include a check that the margin has already been liquidated? Specifically concerned about the case, where a liquidation is rate limited, and someone calls this many times, during the rate limit.

kaleb-keny:
Hey @sunnyvempati one thing please, maybe  worth adding to our testing toolkit. A test where many markets live, and all of the feeds are low tolerance stale, the user should commit an order and execute it later. The execution need 1 price update, if requests for more, the test failed.

kaleb-keny:
@sunnyvempati one more thing that i just remembered, is that you mentioned that INIFINITE stalenss of sale of margin when paying a users settlement fees would be incorporated in this PR? Maybe you already have this task noted down somewhere.

kaleb-keny:
Should be done with the SIP today https://github.com/Synthetixio/SIPs/blob/multi-collateral-margin-draft/content/sips/sip-381.md

markets/perps-market/package.json
sunnyvempati: this PR is not where we should be upgrading; its when we deploy.  i think it was bumped up by mistake here

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
sunnyvempati: as this function states, this is for withdrawing "usd".  we withdraw usd for many reasons (i.e to pay keepers, to pay out profits etc).  

also synths do get sold when paying for fees as we discussed.

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
sunnyvempati: no this is just updating supported collaterals in the system so when integrators call `getSupportedCollaterals` it tells you which ones are active based on if they have a max set.

markets/perps-market/contracts/storage/PerpsCollateralConfiguration.sol
sunnyvempati: yeah, we're using the sell feed on the spot market so whatever that oracle returns we use without any skew/util/fixed fees.

markets/perps-market/contracts/storage/PerpsAccount.sol
sunnyvempati: yes, please see `getAvailableMargin` logic.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
sunnyvempati: oops stray comment; updated.  initially i thought this function would also return to the caller the amt of synth but thats not the case.

markets/perps-market/contracts/interfaces/ILiquidationModule.sol
sunnyvempati: this type of liquidation is never rate limited as there are no positions.  it only gets triggered when there are no positions and your available margin is lower than the debt you need to pay.

markets/perps-market/contracts/storage/PerpsCollateralConfiguration.sol
kaleb-keny: but what's the hardcoded type 2 there? isn't there a better way to fetch that?

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
kaleb-keny: Alright so i guess this is where  `Price.Tolerance.DEFAULT` needs to be  `INFINITE`? And aren't profits already in snxUSD, so why do we need to use this quoter?

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
kaleb-keny: But how do we reconfigure the other parameters in case of update? re-set them?

kaleb-keny:
Hey @sunnyvempati SIP is ready for review https://github.com/Synthetixio/SIPs/pull/1957

markets/perps-market/contracts/storage/PerpsMarketFactory.sol
kaleb-keny: no worries, trying hard to avoid another sip 378. This needs to be done, otherwise liquidators will once again be not compensated for updating the necessary feeds involved with the position being liquidated.

markets/perps-market/contracts/storage/PerpsCollateralConfiguration.sol
kaleb-keny: Like probably i don't understand this fully, but let's say ETH price is 2000$. So if we specify 1, it returns `1/2000`, if we specify `2` it returns 2000?
And the `1` and `2` are things we made up here to call the right function, in order to point to the buy or sell function. Ok my only concern is that spot price can be pushed around by price impact?  Since I think this should be purely the pyth price, not the price comping from the skew adjusted spot market.

markets/perps-market/contracts/storage/LiquidationAssetManager.sol
```diff
@@ -0,0 +1,152 @@
...
+         * @dev Collateral Id (same as synth id)
+         */
+        uint128 id;
+        /**
+         * @dev Distributor address used for reward distribution. If address is 0x0, a new distributor will be created.
+         */
+        address distributor;
+        /**
+         * @dev Addresses of collateral types delegated to the pool. Used to distribute rewards.
+         * @dev Needs to be manually maintained in synch with pool configuration to distribute proportionally to all LPs.
```
keanumaharaj, line 34: TYPO: sync

leomassazza (approved):

PR #1976, opened 2024-01-02
Add requiredMarginForOrderWithPrice

leomassazza:

PR #1995, opened 2024-02-08
SIP-359: Ignore margin requirements on size reduction

sunnyvempati:
- When trader is reducing size, ignore margin requirements as long as account is not in a liquidatable state already.

https://sips.synthetix.io/sips/sip-359/

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsMarket.sol
+++ b/markets/perps-market/contracts/storage/PerpsMarket.sol
@@ -359,13 +359,12 @@ library PerpsMarket {
         uint maxSize,
         uint maxValue,
         uint price,
-        int oldSize,
-        int newSize
+        int128 oldSize,
+        int128 newSize
     ) internal view {
         // Allow users to reduce an order no matter the market conditions.
-        bool isNotReducingInterest = !(MathUtil.sameSide(oldSize, newSize) &&
-            MathUtil.abs(newSize) <= MathUtil.abs(oldSize));
-        if (isNotReducingInterest) {
+        bool isReducingInterest = MathUtil.isSameSideReducing(oldSize, newSize);
+        if (!isReducingInterest) {
             int newSkew = self.skew - oldSize + newSize;
 
             int newMarketSize = self.size.toInt() -
--- a/markets/perps-market/contracts/utils/MathUtil.sol
+++ b/markets/perps-market/contracts/utils/MathUtil.sol
@@ -45,7 +45,7 @@ library MathUtil {
     }
 
     function isSameSideReducing(int128 a, int128 b) internal pure returns (bool) {
-        return sameSide(a, b) && abs(b) <= abs(a);
+        return sameSide(a, b) && abs(b) < abs(a);
     }
 
     function ceilDivide(uint a, uint b) internal pure returns (uint) {
--- a/markets/perps-market/contracts/utils/MathUtil.sol
+++ b/markets/perps-market/contracts/utils/MathUtil.sol
@@ -44,6 +44,10 @@ library MathUtil {
         return (a == 0) || (b == 0) || (a > 0) == (b > 0);
     }
 
+    function isSameSideReducing(int128 a, int128 b) internal pure returns (bool) {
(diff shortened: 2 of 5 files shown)


markets/perps-market/contracts/storage/PerpsMarket.sol
kaleb-keny, line 1: So i'm not entirely certain implementing this inside `getRequiredMarginWithNewPosition`  would not break other simulation things that leo worked on. like in order to be able to estimate what would be the margin for any given order... So maybe worth thinking about implementing inside `validateRequest` if we have parts that could break?
`requiredMarginForOrderWithPrice` or `requiredMarginForOrder` maybe could break?

kaleb-keny: example https://discord.com/channels/413890591840272394/811367528810020924/1205261678539841627

markets/perps-market/contracts/utils/MathUtil.sol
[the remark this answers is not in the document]
leomassazza: regarding <=, mathematically it should be strict less, otherwise is not reducing but keeping it same size. But since we already checked delta is not zero.

markets/perps-market/contracts/storage/PerpsMarket.sol
leomassazza: mmmm... here, I don't know. Technically, if there's no margin requirement to reduce an order, then it's ok that those will work the same way, no matter the price. 
In fact, If we don't use this condition here, the `WithPrice` view can show some weird results vs the reality.

leomassazza:
Please, add a test to cover positions starting as short to validate the maths both ways

markets/perps-market/contracts/storage/PerpsMarket.sol
kaleb-keny: Hey @leomassazza worried about integrators using those functions, to get a read on the margin given a change in a position, which might result in problems their end. What i mean is that the purpose of these functions is to be able to service limit orders scheduled by integrators, and if they keep track of the required margin with these functions, adding the return 0 here messes up the integration.

markets/perps-market/contracts/utils/MathUtil.sol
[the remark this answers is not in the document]
sunnyvempati: yeah good call, i'll change it to `<`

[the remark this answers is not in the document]
sunnyvempati: @kaleb-keny i'll add a comment above it to make sure i call out the importance of the order here.  but yeah old/new don't make sense for "MathUtil" where it's more generic.

markets/perps-market/contracts/storage/PerpsMarket.sol
sunnyvempati: @kaleb-keny if integrators want something different, we can make it a different function entirely but `getRequiredMarginWithNewPosition` returning something other than 0 when size is reducing is plain wrong.  this function exists mainly because it's shared between calculating it for simulation and the actual order.  it's be design that they both should be the exact same logic.

again, if integrators want something different, i'd suggest a different function that has some caveats.

sunnyvempati:
@leomassazza yessir, on it.

markets/perps-market/contracts/storage/PerpsMarket.sol
kaleb-keny: hey @sunnyvempati integrators asked for functions that allow them to get back the initial margin with different prices / position sizes... 
Leo worked on this in this PR https://github.com/Synthetixio/synthetix-v3/pull/1976 and he used this function.
So making this change, will break what integrators are already using to get an idea about the initialMargin given different prices / sizes

sunnyvempati: @kaleb-keny currently this function returns one uint256 that is the required margin for an order.  this should be enough for integrators to knwo if their limit order will go through.

I think what you are pointing out is knowing what the initial/maint margins are _after_ the order is complete.  this is a different spec entirely and we can create a separate function for them if they request it.  but currently, the function `getRequiredMarginWithNewPosition` is accurate in its intention.

PR #1997, opened 2024-02-13
Update deps

noisekit:
- Upgrade all the deps
- still keep incompatible releases from updating (like ethers, node-fetch, etc)
- apply prettier
- turn off a couple of solhint rules that are not compatible with current codebase and give over 1000 warnings (temporarily?)
- upgraded -> reverted -> locked version of cannon to `2.10.6` (later ones have issues)


leomassazza (approved):
LGTM.

PR #1998, opened 2024-02-13
Lint fix `explicit-types`

noisekit:
Some of the instances like `type(int).max` made parser fail, and were fixed manually. All the rest was automatic.


leomassazza (approved):
LGTM

PR #1999, opened 2024-02-13
Cannon 2.11.x

noisekit:


utils/core-utils/src/utils/solidity/compiler.ts
```diff
@@ -88,11 +88,7 @@ async function _getCompiler(version: string): Promise<any> {
   const isDownloaded = await downloader.isCompilerDownloaded(version);
 
   if (!isDownloaded) {
-    await downloader.downloadCompiler(
-      version,
-      async () => {},
-      async () => {}
-    );
+    await downloader.downloadCompiler(version);
```
noisekit, line 91: This is needed.

noisekit: Typecheck fails. It is type-related only.

yarn.lock
```diff
@@ -3144,6 +3144,20 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@synthetixio/router@npm:^3.3.6":
```
noisekit, line 3147: We have 2 routers now. Need to dedupe/resolve them. 

IMG_1607

PR #2002, opened 2024-02-16
remove name from initialize

sunnyvempati:
- will be using the setter for setting market name instead of as part of `initializeFactory`.  cannon caches the `initializeFactory` call so it wouldn't get called again and the function call in cannon was not updated to include name.

### DIFF
--- a/markets/perps-market/storage.dump.sol
+++ b/markets/perps-market/storage.dump.sol
@@ -50,11 +50,11 @@ library ERC2771Context {
 
 // @custom:artifact @synthetixio/core-contracts/contracts/utils/HeapUtil.sol:HeapUtil
 library HeapUtil {
-    uint private constant _ROOT_INDEX = 1;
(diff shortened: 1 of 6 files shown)


markets/perps-market/storage.dump.sol
```diff
@@ -50,11 +50,11 @@ library ERC2771Context {
 
 // @custom:artifact @synthetixio/core-contracts/contracts/utils/HeapUtil.sol:HeapUtil
 library HeapUtil {
-    uint private constant _ROOT_INDEX = 1;
```
sunnyvempati, line 53: i dont know why storage dump changed, it might just have not been committed in a while

leomassazza: This is due to the explicit types new rule in the repo (uint/int should be explicit uint256/int256). A PRs added that a couple of days ago.
Storage allocation should be the same.

leomassazza (approved):

PR #2008, opened 2024-02-24
Refactor tests: Chapter One

noisekit:
Only a draft here, still plenty of work to be done
- Need to unwrap all the `it` cases because for each `it` there is a lot of preparation work being done in before/beforeAll blocks (many of them are hidden away)
- Remove all `forEach` because they are executed in parallel and not sequence

All of the `forEach` need to go! 
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,19 +138,18 @@ commands:
             fi
 jobs:
   build-testable:
+    parameters:
+      wat:
+        type: string
+        default: "--all"
     docker:
       - image: cimg/node:<< pipeline.parameters.node-version >>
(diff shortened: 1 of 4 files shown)


leomassazza (approved):
the refactor looks good. Don't know if that will fix the tests since looks like is related to other issue, but at least it will run quicker.

noisekit:
> the refactor looks good. Don't know if that will fix the tests since looks like is related to other issue, but at least it will run quicker.

Yeah the idea is to increase stability. As we run one Anvil instance per test file - we have much better way to scale now and reduce flakiness. Failures indeed are unrelated.

PR #2011, opened 2024-02-26
Add a flag to use ecotone or legacy

leomassazza:
- [x] Add a flag + try/catch to identify if the precompiled gas oracle is using ecotone or not
- [x] Fix tests and mocks to work properly

### DIFF
--- a/auxiliary/OpGasPriceOracle/contracts/OpGasPriceOracle.sol
+++ b/auxiliary/OpGasPriceOracle/contracts/OpGasPriceOracle.sol
@@ -73,20 +73,45 @@ contract OpGasPriceOracle is IExternalNode {
         RuntimeParams memory runtimeParams
     ) internal view returns (uint256 costOfExecutionGrossEth) {
         IOVM_GasPriceOracle ovmGasPriceOracle = IOVM_GasPriceOracle(ovmGasPriceOracleAddress);
+        bool isEcotone;
+        try ovmGasPriceOracle.isEcotone() returns (bool _isEcotone) {
+            isEcotone = _isEcotone;
+        } catch {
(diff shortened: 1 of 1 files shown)


auxiliary/OpGasPriceOracle/contracts/OpGasPriceOracle.sol
```diff
@@ -73,20 +73,45 @@ contract OpGasPriceOracle is IExternalNode {
...
+                l1BaseFee *
+                16 +
+                blobBaseFeeScalar *
+                blobBaseFee) / (16 * 10 ** decimals);
 
-        uint256 l1GasPrice = (baseFeeScalar * l1BaseFee * 16 + blobBaseFeeScalar * blobBaseFee) /
-            (16 * 10 ** decimals);
+            costOfExecutionGrossEth = ((gasUnitsL1 * l1GasPrice) + (gasUnitsL2 * gasPriceL2));
+        } else {
+            // If it's not an ecotone, use the legacy formula and interface
```
kaleb-keny, line 103: for the sake of consistency, not important thought, using `ovmGasPriceOracle.baseFee()` instead of `ovmGasPriceOracle.gasPrice`

kaleb-keny: We also need to add the try/catch on the validation step i think @leomassazza ?
https://github.com/Synthetixio/synthetix-v3/blob/main/auxiliary/OpGasPriceOracle/contracts/OpGasPriceOracle.sol#L135

leomassazza: We are not doing the try/catch in the verification and just checking the common interface. The intention is to at least know the contract address looks like the precompiled one

kaleb-keny (approved):

PR #2014, opened 2024-02-26
Apply underflow fix (prevent revert)

leomassazza:


markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -73,13 +73,14 @@ library GlobalPerpsMarket {
         Data storage self
     ) internal view returns (uint128 rate, uint256 delegatedCollateralValue, uint256 lockedCredit) {
         uint256 withdrawableUsd = PerpsMarketFactory.totalWithdrawableUsd();
-        delegatedCollateralValue = withdrawableUsd - totalCollateralValue(self);
+        int256 delegatedCollateralValueAsInt = withdrawableUsd.toInt() -
+            totalCollateralValue(self).toInt();
```
kaleb-keny, line 77: this number `totalWithdrawableUsd` is comprised of `creditCapacity`  which is basically
1) lp's delegated collateral 
2) pnl (i.e. debt)

PLUS  `getDepositedCollateralValue,` which is all  traders' margin (aka collaterals)

Now this probably means 
We can just pull `creditCapacity` since `delegatedCollateralValueAsInt`  is  just recalculating something that we already have?
also how do i get a read on this `totalWithdrawableUsd` in cannon, can't seem to be able to find this function in PerpMarketProxy

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -73,13 +73,14 @@ library GlobalPerpsMarket {
...
         lockedCredit = minimumCredit(self);
-        if (delegatedCollateralValue == 0) {
+        if (delegatedCollateralValueAsInt <= 0) {
             return (0, 0, lockedCredit);
```
kaleb-keny, line 80: If a market is undercollateralized, does it mean interest this rate factor here is zero? I think it might need to be set to1, because we kind of , need in this situation, folks to wind down their positions, so that the market becomes less volatile, for a "fix" to be proposed that re-establishes equilibrium is my intuitive take on it.

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -73,13 +73,14 @@ library GlobalPerpsMarket {
...
         }
 
-        rate = lockedCredit.divDecimal(delegatedCollateralValue).to128();
+        rate = lockedCredit.divDecimal(delegatedCollateralValueAsInt.toUint()).to128();
```
kaleb-keny, line 83: I dont think we need to re-uInt it 😅  since we already caught he negative case?

sunnyvempati: need to make it uint to divide by another uint

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
sunnyvempati: confirmed with afif, set it to 1

sunnyvempati (approved):

kaleb-keny (approved):

PR #2015, opened 2024-02-29
Update to cannon 2.11.18

mjlescano:


noisekit (approved):
Looks ok. perps tests (Liquidation specifically) dont really pass right now anywhere. Need refactoring

PR #2027, opened 2024-03-06
New deploy perps and Arbitrum gas oracle

sunnyvempati:


markets/perps-market/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@synthetixio/perps-market",
-  "version": "3.3.15",
+  "version": "3.3.16",
```
noisekit, line 3: Not a good idea to do it this way.
If we start changing manually all the versions - we must then remove lerna and start doing this manually. We cannot do both.

```suggestion
  "version": "3.3.15",
```

noahlitvin (approved):

PR #2028, opened 2024-03-07
Split setMaxMarketSize into two setters, adding setMaxMarketValue

noisekit:
- Partial revert "Cap on value (#1978)"
- Blocked by #2029


dbeal-eth (approved):
changes look good. I just dont understand why they are needed?

PR #2036, opened 2024-03-12
Add more views and events based on Troy's recommendation

davidvuong:


markets/bfp-market/test/integration/modules/MarginModule.test.ts
```diff
@@ -2546,6 +2546,102 @@ describe('MarginModule', async () => {
...
+      const { PerpMarketProxy } = systems();
+
+      const desiredMarginUsdDepositAmount = genOneOf([5000, 10_000, 20_000]);
+      const { trader, marketId } = await depositMargin(
+        bs,
+        genTrader(bs, { desiredMarginUsdDepositAmount })
+      );
+
+      const margin = await PerpMarketProxy.getWithdrawableMargin(trader.accountId, marketId);
+      assertBn.near(margin, bn(desiredMarginUsdDepositAmount), bn(0.000001));
```
0xjocke, line 2596: This can probably be equal instead of near?

0xjocke (approved):

markets/bfp-market/test/integration/modules/MarginModule.test.ts
davidvuong: Due to how collateral amounts are deposited, sometimes this can lead to values not totally exact so hence the `.near`

0xjocke: Ah that's right, we're specifying usd amount

PR #2043, opened 2024-03-14
Bfp market margin liq

davidvuong:


davidvuong:
All tests are passing barring an existing flaky test (to be fixed separately - already tracked as an issue to resolve).

> PerpMarketFactoryModule Utilization Market digest UtilizationRate should return current utilization rate when highUtilization

0xjocke (approved):
Nice!

PR #2045, opened 2024-03-15
More test retriers

noisekit:


.eslintrc.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 12,
   },
-  ignorePatterns: ['/*.js'],
+  ignorePatterns: ['!.*', '!.circleci/test-batch.js'],
```
0xjocke, line 18: Is this right?  I'm reading it as "not ignoring all files in this folder and not ignoring `circleci/test-batch"`. Friday afternoon and double negatives, maybe I'm missing something?

noisekit: yeah not ignore all of them and not ignore the one specific... idk, eslint is weird

noisekit: We need to switch to biome anyway

0xjocke: haha fair enough. I would've thought files were included by default.. But I'm sure there's reasons... Just wanted to confim it wasn't a mistake

0xjocke (approved):

PR #2059, opened 2024-03-21
Replace Rewards Distributor

leomassazza:


davidvuong (approved):
🔥

auxiliary/RewardsDistributor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@synthetixio/rewards-distributor",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
noisekit, line 3: Please don't do this. Lerna will increment this on the next release.
We publish with lerna as usual. It updates versions. Only after that we deploy with cannon (with updated versions) whatever we think needs to be deployed and published.

leomassazza: ok, will rollback to 0.0.2. Thanks for noticing

PR #2067, opened 2024-03-26
msg.sender -> ERC2771Context._msgSender()

0xjocke:
Use ERC2771Context_msgSender() instead of msg sender.
Except PerpRewardDistributor and mocks

### DIFF
--- a/markets/bfp-market/contracts/modules/LiquidationModule.sol
+++ b/markets/bfp-market/contracts/modules/LiquidationModule.sol
@@ -16,8 +16,7 @@ import {Position} from "../storage/Position.sol";
 import {ErrorUtil} from "../utils/ErrorUtil.sol";
 import {MathUtil} from "../utils/MathUtil.sol";
 import {Flags} from "../utils/Flags.sol";
-
-/* solhint-disable meta-transactions/no-msg-sender */
+import {ERC2771Context} from "@synthetixio/core-contracts/contracts/utils/ERC2771Context.sol";
 
 contract LiquidationModule is ILiquidationModule {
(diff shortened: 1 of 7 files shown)


markets/bfp-market/contracts/modules/LiquidationModule.sol
```diff
@@ -254,12 +253,18 @@ contract LiquidationModule is ILiquidationModule {
         liquidateCollateral(accountId, marketId, market, globalConfig);
 
         // Flag and emit event.
-        market.flaggedLiquidations[accountId] = msg.sender;
+        market.flaggedLiquidations[accountId] = ERC2771Context._msgSender();
```
davidvuong, line 256: You're calling `_msgSender` 3 times. Store in a local var once then reuse the var. Depending on how the contracts were optimised, this could lead to 3 storage loads to check for `isTrustedMulticallForwarder` as well as unnecessary rechecks to verify and decode `msg.data` over and over and over.

markets/bfp-market/contracts/modules/LiquidationModule.sol
```diff
@@ -298,14 +303,18 @@ contract LiquidationModule is ILiquidationModule {
         }
 
         // Pay the keeper
-        globalConfig.synthetix.withdrawMarketUsd(marketId, msg.sender, liqKeeperFee);
+        globalConfig.synthetix.withdrawMarketUsd(
+            marketId,
+            ERC2771Context._msgSender(),
```
davidvuong, line 308: Same here. It's called twice.

davidvuong (approved):

PR #2094, opened 2024-04-04
Cannon v2.12

noahlitvin:


dbeal-eth (approved):
test failures appear to be same as ones currently in main

dbeal-eth:
not sure about the bfp failure tho that seems pretty random. doesn't *seem* to be cannon related. ideas?

davidvuong:
errrrrrr guys, whatever is in this version of cannon is breaking the `bfp-market` tests.

Before this was merged, you can see in build, the `test-bfp-market` task has been passing without manual retries. However, now that this is merged, random tests are breaking.

PR #2097, opened 2024-04-05
Make sure we calculate mm with collateralUsd

0xjocke:
- Also only call getLiquidationMarginUsd once to save some gas

### DIFF
--- a/markets/bfp-market/contracts/storage/Position.sol
+++ b/markets/bfp-market/contracts/storage/Position.sol
@@ -136,42 +136,17 @@ library Position {
     }
 
     /**
-     * @dev Validates whether the `newPosition` is below initial margin req (IM).
-     *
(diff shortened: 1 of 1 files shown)


markets/bfp-market/contracts/storage/Position.sol
```diff
@@ -135,43 +135,18 @@ library Position {
         return MathUtil.max(marginUsd.toInt() - orderFee.toInt() - keeperFee.toInt(), 0).toUint();
     }
 
-    /**
-     * @dev Validates whether the `newPosition` is below initial margin req (IM).
-     *
-     * NOTE: We expect marginUsd here to be the discount adjusted margin due to liquidation checks as margin checks
-     * for liquidations are expected to discount the collateral.
-     */
-    function validateNextPositionIm(
```
davidvuong, line 144: 🔥

davidvuong (approved):

PR #2103, opened 2024-04-10
Rename library to prevent name collision in tooling

leomassazza:


markets/perps-market/storage.dump.sol
```diff
@@ -857,6 +892,7 @@ library PerpsMarketFactory {
         address spotMarket;
         uint128 perpsMarketId;
         string name;
+        address liquidationAssetManager;
```
noisekit, line 895: Can we decouple rename from adding new feature?

sunnyvempati (approved):

markets/perps-market/storage.dump.sol
sunnyvempati: build wasn't working on the parent branch and now that it is, thats why the storage changes were here.  should be fine to merge

PR #2107, opened 2024-04-11
Upgrade cannon 2.12.2

mjlescano:


noisekit (approved):
The failing tests need to be fixed separately, but cannon update should be fine. I can update ipfs url later when release is merged in

noisekit (approved):
3 failing tests are expected. One from bfp and 2 from perps. We will fix them separately

PR #2108, opened 2024-04-15
Dont allow skewScale 0 

0xjocke:


davidvuong:
Code looks fine but consider writing a test with zero skewScale. Also, failing test

> LiquidationModule "before each" hook: restore for "should remove all margin collateral on flag"

🤔

0xjocke:
> Code looks fine but consider writing a test with zero skewScale. Also, failing test
> 
> > LiquidationModule "before each" hook: restore for "should remove all margin collateral on flag"
> 
> 🤔

Yep will do, had problems running test at all yesterday. Main is also failing on this one. I just tried nuking cache on main an reran. But no success so far

markets/bfp-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -53,6 +54,10 @@ contract MarketConfigurationModule is IMarketConfigurationModule {
         // Only allow an existing per market to be configurable. Ensure it's first created then configure.
         PerpMarket.exists(marketId);
 
+        if (data.skewScale == 0) {
+            revert ErrorUtil.ZeroAmount();
```
davidvuong, line 58: I want to add more configuration validation checks here. Can we use a `error InvalidParameter(string parameter, string reason);` here instead?

0xjocke: Yep cool I was thinking about that too

0xjocke: Done!

davidvuong (approved):

PR #2110, opened 2024-04-16
SIP-378 Section C: Fix maker/taker set to zero bug

leomassazza:
There's an issue where maker or taker fees are set to zero that expresses when attempting to send 0 fees to a referrer. The fix just adds a conditional on zero fees collected.


kaleb-keny (approved):
simplest fix ever

markets/perps-market/contracts/storage/GlobalPerpsMarketConfiguration.sol
```diff
@@ -207,7 +207,7 @@ library GlobalPerpsMarketConfiguration {
         address referrer,
         PerpsMarketFactory.Data storage factory
     ) private returns (uint256 referralFeesSent) {
-        if (referrer == address(0)) {
+        if (fees == 0 || referrer == address(0)) {
```
sunnyvempati, line 210: nit: feel like this check should exist at the caller level.  in `collectFees()` we should return (0,0) if the orderFees == 0.  this function shouldn't even be called if there's no fees to collect.

sunnyvempati (approved):
small nit but lgtm

sunnyvempati:
@leomassazza i think rebase will fix the test

leomassazza:
Closed and commits cherrypicked to main SIP-378 branch. All SIP-378 is now only one PR

PR #2117, opened 2024-04-18
keeperFeeBufferUsd should not be bigger than margin

0xjocke:
Also improve cancellation keeper fee calulation by excluding keeperBufferFee + separate gas units


markets/bfp-market/contracts/utils/ErrorUtil.sol
```diff
@@ -124,4 +124,7 @@ library ErrorUtil {
 
     /// @notice Thrown when user trying to merge accounts with positions on opposite sides.
     error InvalidPositionSide();
+
+    /// @notice Thrown when user trying commit an order with `keeperFeeBufferUsd` larger than margin.
+    error KeeperBufferFeeTooLarge(uint256 keeperFeeBufferUsd, uint256 max);
```
davidvuong, line 129: `max` is not technically the max as it's more like `maxKeeperFee - fees`, I'd just revert here without any of the attributes as (1) `keeperBufferUsd` is already supplied on commit, (2) if it's failing on settlement then we already store in the order, and (3) max here isn't the true max.

markets/bfp-market/contracts/storage/Position.sol
```diff
@@ -188,6 +188,13 @@ library Position {
             params.oraclePrice
         );
 
+        if (params.keeperFeeBufferUsd > marginValuesForLiqValidation.discountedMarginUsd) {
```
davidvuong, line 191:
```suggestion
        // Prevents execution optionality with `keeperFeeBufferUsd` on settlement. We don't need to consider
        // maker/taker fees here as cancelations will not lead to charging order fees.
        if (params.keeperFeeBufferUsd > marginValuesForLiqValidation.discountedMarginUsd) {
```

davidvuong (approved):

davidvuong:
Contracts look fine but there are now failing unrelated tests due to the new revert.

markets/bfp-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -49,10 +50,14 @@ contract MarketConfigurationModule is IMarketConfigurationModule {
         OwnableStorage.onlyOwner();
 
         PerpMarketConfiguration.Data storage config = PerpMarketConfiguration.load(marketId);
-
+        PerpMarketConfiguration.GlobalData storage globalConfig = PerpMarketConfiguration.load();
```
davidvuong, line 53: Minor: You can pull this below PerpMarket.exists(marketId), just above the if check where its used.

markets/bfp-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -49,10 +50,14 @@ contract MarketConfigurationModule is IMarketConfigurationModule {
...
         // Only allow an existing per market to be configurable. Ensure it's first created then configure.
         PerpMarket.exists(marketId);
 
+        if (data.minMarginUsd < globalConfig.maxKeeperFeeUsd) {
+            revert ErrorUtil.InvalidMinMargin();
```
davidvuong, line 58: Can you use the same revert `error InvalidParameter(string parameter, string reason);`?

markets/bfp-market/test/generators.ts
```diff
@@ -113,7 +113,7 @@ export const genMarket = () => ({
     takerFee: bn(genNumber(0.0006, 0.0008)), // 1 - 8bps
     maxMarketSize: bn(genNumber(20_000, 50_000)),
     maxFundingVelocity: bn(genNumber(3, 9)),
-    minMarginUsd: bn(genNumber(50, 60)),
+    minMarginUsd: bn(genNumber(500, 100)),
```
davidvuong, line 116: `500, 100` looks wrong. 500, 600?

markets/bfp-market/test/helpers.ts
```diff
@@ -23,7 +24,7 @@ declare module 'ethers' {
     [customInspectSymbol]?: () => string;
   }
 }
-ethers.BigNumber.prototype[customInspectSymbol] = function (this: ethers.BigNumber) {
+hre.ethers.BigNumber.prototype[customInspectSymbol] = function (this: ethers.BigNumber) {
```
davidvuong, line 27: 🤔 What's this for again and why the change?

0xjocke: I think it's related to cannon using viem internally. I think that's when it stopped working. I guess now they grab ethers from hre instead?

markets/bfp-market/test/generators.ts
0xjocke: Opsie, thanks!

markets/bfp-market/contracts/modules/MarketConfigurationModule.sol
```diff
@@ -49,11 +50,19 @@ contract MarketConfigurationModule is IMarketConfigurationModule {
...
         // Only allow an existing per market to be configurable. Ensure it's first created then configure.
         PerpMarket.exists(marketId);
 
+        PerpMarketConfiguration.Data storage config = PerpMarketConfiguration.load(marketId);
+        PerpMarketConfiguration.GlobalData storage globalConfig = PerpMarketConfiguration.load();
+
+        if (data.minMarginUsd < globalConfig.maxKeeperFeeUsd) {
+            revert ErrorUtil.InvalidParameter(
+                "minMarginUsd",
+                "minMarginUsd can't be smaller than maxKeeperFeeUsd"
```
davidvuong, line 62:
```suggestion
                "minMarginUsd cannot be less than maxKeeperFeeUsd"
```

markets/bfp-market/contracts/modules/OrderModule.sol
```diff
@@ -449,10 +449,10 @@ contract OrderModule is IOrderModule {
             }
         }
 
-        // If `isAccountOwner` then 0 else chargeFee.
+        // If `isAccountOwner` then 0 else chargeFee. Also note that we pass 0 for `keeperFeeBufferUsd`, to avoid some edgecase scenarios that can lead to bad debt.
```
davidvuong, line 452:
```suggestion
        // If `isAccountOwner` then 0 else charge cancellation fee.
```

markets/bfp-market/contracts/storage/Order.sol
```diff
@@ -120,6 +120,27 @@ library Order {
         return boundedKeeperFeeUsd;
     }
 
+    function getCancellationKeeperFee() internal view returns (uint256) {
```
davidvuong, line 123:
```suggestion
    /// @dev Returns the keeper fee in USD for order cancellations.
    function getCancellationKeeperFee() internal view returns (uint256) {
```

markets/bfp-market/contracts/storage/PerpMarketConfiguration.sol
```diff
@@ -49,6 +49,8 @@ library PerpMarketConfiguration {
         uint128 keeperProfitMarginPercent;
         /// Number of gas units required to perform an order settlement by a keeper.
         uint128 keeperSettlementGasUnits;
+        /// Number of gas units required to perform a cancellation of an order.
```
davidvuong, line 52:
```suggestion
        /// Number of gas units required to perform an order cancellation by a keeper.
```

markets/bfp-market/test/integration/modules/MarketConfigurationModule.test.ts
```diff
@@ -83,6 +84,23 @@ describe('MarketConfigurationModule', async () => {
       );
     });
 
+    it('should revert when minMarginUsd is smaller then maxKeeperFeeUsd', async () => {
```
davidvuong, line 87:
```suggestion
    it('should revert when minMarginUsd less than maxKeeperFeeUsd', async () => {
```

markets/bfp-market/test/integration/modules/MarketConfigurationModule.test.ts
```diff
@@ -83,6 +84,23 @@ describe('MarketConfigurationModule', async () => {
...
+      const globalConfig = await BfpMarketProxy.getMarketConfiguration();
+      const { specific } = genMarket();
+      const config = {
+        ...specific,
+        minMarginUsd: globalConfig.maxKeeperFeeUsd.sub(bn(1)),
+      };
+
+      await assertRevert(
+        BfpMarketProxy.setMarketConfigurationById(marketId, config),
+        `InvalidParameter("minMarginUsd", "minMarginUsd can't be smaller than maxKeeperFeeUsd")`,
```
davidvuong, line 99:
```suggestion
        `InvalidParameter("minMarginUsd", "minMarginUsd cannot be less than maxKeeperFeeUsd")`,
```

markets/bfp-market/test/integration/modules/OrderModule.cancel.test.ts
```diff
@@ -24,9 +24,12 @@ import {
   getFastForwardTimestamp,
   getPythPriceDataByMarketId,
   isSusdCollateral,
+  setBaseFeePerGas,
   setMarketConfiguration,
   withExplicitEvmMine,
 } from '../../helpers';
+import { calcKeeperCancellationFee } from '../../calculations';
+import { BigNumber } from 'ethers';
```
davidvuong, line 32: Minor: Group 3rd party imports together above.

davidvuong (approved):
LGTM. Just messaging and comments. Also, cancelled vs canceled and cancellation vs cancelation.

> Both "cancelled" and "canceled" are correct spellings in English, but their usage can depend on the region. In American English, "canceled" is more common, while in British English, "cancelled" with two 'l's is preferred.

> Similarly, both "cancellation" and "cancelation" are acceptable, but "cancellation" is more widely used. Again, this preference may vary slightly depending on regional differences.

🤷

davidvuong:
Also optionally consider having a generic `getKeeperFee` where you simply pass in `ethPrice` and `gasUnits` because settlement, liquidation, and cancellation all follow the same math.

PR #2118, opened 2024-04-19
Rerun failed tests only

noisekit:
Now we have ability to rerun only failed tests, skipping all the successful ones.

_ 2024-04-19 at 21 09 47

_ 2024-04-19 at 21 11 04

There is also an optimisation that skips the full yarn install and few more steps if there is no tests in the job, so we have an early successful exit.

_ 2024-04-19 at 21 10 25

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,6 +266,22 @@ jobs:
       #CANNON_IPFS_URL: "http://127.0.0.1:5001"
     steps:
       - checkout
+
+      - run:
+          name: "Split tests"
+          command: |
+            _dir=$(yarn workspace "<< parameters.workspace >>" exec pwd)
+            echo "Workspace: $_dir"
+            cd $_dir
+
+            circleci tests glob 'test/**/*.test.ts' | circleci tests run --command="cat > /tmp/files.txt" --verbose
+
+            # if there are no tests, terminate execution after this step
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -543,7 +559,7 @@ jobs:
 workflows:
   version: 2.1
 
-  anvil-per-test:
+  skip:
```
0xjocke, line 562: Is this name change on purpose?

.circleci/config.yml
```diff
@@ -580,67 +596,57 @@ workflows:
...
-          workspace: "@synthetixio/core-utils"
+          dir: "./utils/core-utils"
           parallelism: 2
           batch-size: 5
 
-      #- test-contracts: TODO: enable when tests are fixed
-      #    name: "test-legacy-market"
-      #    requires: [build-testable]
-      #    workspace: "@synthetixio/legacy-market"
-
```
0xjocke, line 643: Maybe keep this? Or is legacy market not a thing anymore?

.circleci/config.yml
noisekit: Yeah just don't want it to be this oddly specific

.circleci/config.yml
noisekit: It has some old config anyway. We can copy paste any other block from above and update for legacy market

davidvuong (approved):

PR #2120, opened 2024-04-19
support allocation of debt by debt shares

dbeal-eth:
we have to have v3 record debt the same way that v2x does. Before this change, v3 was using the number of SNX tokens to allocate debt.

After this change, v3 will be using debt shares (which are redeemable into their equivalent snx) for collateral. there are some restrictions:
* the pool that is connected to the legacy market can be the only pool connected
* the pool can only accept this redeemable collateral

the pool can still be used to connect to other markets, such as the bfp market, in the future, and other pools can be connected to the bfp market, but this change is necessary to ensure that debt allocations are correctly maintained.


dbeal-eth:
this is no longer needed and overly complicated

PR #2129, opened 2024-04-30
Contract storage layout verification improvements

mjlescano:
## Usage
1. `yarn hardhat storage:layout`: Will pretty print the storage layout of the contracts at `contracts/` folder.
2. `yarn hardhat storage:validate`: Will perform simple validations (the old ones) on the `contracts/` folder. It checks that any contract is using state variables and that the storage slot keys are not re-used by multiple contracts.
3. `yarn hardhat storage:dump`: Will parse the contracts, generate a JSON format storage layout and save it to `storage.dump.json`. The steps taken are as follow:
  1. Parse local contracts
  2. Perform static validations (same as `yarn hardhat storage:validate`)
  3. Generate storage layout dump
  4. If a previous dump exists, it will verify for invalid storage mutation changes between the old dump and the new one (and throw an error in case of invalid result)
  5. If everything went well, save new dump at `storage.dump.json`
4. `yarn hardhat storage:verify storage.prev.dump.json storage.dump.json`: Performs storage mutation changes between the two given storage dumps.

## TODO
- [x] Add `yarn hardhat storage:layout` command for pretty printing the storage layout
- [x] Throw an error when changing the slot/offset/size of an existing variable
- [x] Do not allow the deletion of storage slots (do not allow deletion of variables)
- [x] Add info message when renaming a variable
- [ ] Add info message when adding values that could be re-ordered to share slots.
  * ensure that the user tries to order the storage variables and struct members such that they can be packed tightly. For example, declaring storage variables in the order of uint128, uint128, uint256 instead of uint128, uint256, uint128, as the former will only take up two slots of storage whereas the latter will take up three.
- [ ] Update README
- [ ] [discussion] better dump format to improve readability (we want the users to easily be able for manually checking the changes on a PR, so they don't have to rely 100% on the verification script)
- [ ] Fix failing `check:storage` CI job, which is not compatible with this PR

## Notes
* Full deletions of a struct are allowed, but in case they are re-added in the future maybe the slots can have old information, and the user does not know about it. Maybe the re-addition of the old slots have different layout, with unintended consequences. We cannot really do much about than showing that the struct/contract is being deleted.
* Right now, the verifier is strict in the sense that it doesn't let you delete or change types of any variables, maybe we can think on some specific cases were this would be useful to be added (e.g. if we want to let the user to change a type `address` for an `uint256` and it doesn't modify the slots of other variables in the same layout).


auxiliary/BuybackSnx/storage.dump.sol
```diff
@@ -1,114 +0,0 @@
-// SPDX-License-Identifier: UNLICENSED
```
noisekit, line 1: dump lost?

auxiliary/ERC4626ToAssetsRatioOracle/storage.dump.sol
```diff
@@ -1,109 +0,0 @@
-// SPDX-License-Identifier: UNLICENSED
```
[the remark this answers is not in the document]
noisekit: And in many other packages - no more storage

auxiliary/BuybackSnx/storage.dump.sol
mjlescano: Not lost, they are being replaced by `storage.dump.json` files, have the advantage on being much easier to parse/use, ofc on the cost of readability but I think is a good trade off here. 
Screenshot 2024-07-02 at 18 19 49

mjlescano: Also, this new implementation now stores in the dump the `size`, `slot` and `offset` for each variable, which point to the exact location on storage and it is what is used for validations.

noisekit: Like for example 

image

mjlescano: That's right, it is because on the previous version we were also dumping constants/immutables, but as they are not really required it is not generated anymore. Also, it dumped all the imported dependencies, regardless they were used or not. Now it just dumps local contracts that have `struct`'s defined, with the explicitly used imported values used by that struct.

- `utils/sample-project`: forgot to create the new dump for this one, going to add it.
- `auxiliary/BuybackSnx`: only had `immutable` constants, not storage
- `auxiliary/ERC4626ToAssetsRatioOracle`: only had `immutable` constants, not storage
- `auxiliary/SpotMarketOracle`: only had `immutable` constants, not storage
- `auxiliary/WstEthToStEthRatioOracle`: only had `immutable` constants, not storage
- `utils/core-modules`: this one didn't have the storage check task, which i think is ok as it is being checked by the used ones.

dbeal-eth:
this is ready for merge.

dbeal-eth (approved):
its time.

noisekit (approved):

PR #2135, opened 2024-05-03
Make sure getWithdrawableMargin returns uint

0xjocke:


markets/bfp-market/contracts/modules/MarginModule.sol
```diff
@@ -598,7 +598,11 @@ contract MarginModule is IMarginModule {
         // When there is no position then we can ignore all running losses/profits but still need to include debt
         // as they may have realized a prior negative PnL.
         if (size == 0) {
-            return marginValues.collateralUsd - Margin.load(accountId, marketId).debtUsd;
+            return
+                MathUtil.max(
```
davidvuong, line 602: Minor: A quick test for this?

You can use the `__test_addDebtUsdToAccountMargin` function in `TestHelperModule` to quickly add debt to an account.

davidvuong (approved):
LGTM but would like a quick test on the max check.

markets/bfp-market/contracts/modules/MarginModule.sol
0xjocke: Thanks, this actually exposed a missing cast!

PR #2138, opened 2024-05-05
SIP-369

noahlitvin:


protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -34,8 +34,6 @@ contract RewardsManagerModule is IRewardsManagerModule {
     using Distribution for Distribution.Data;
     using RewardDistribution for RewardDistribution.Data;
 
-    uint256 private constant _MAX_REWARD_DISTRIBUTIONS = 10;
```
noisekit, line 37: Just to make sure when we remove const here won't this change data structure and potentially lead to storage slot collisions?

noahlitvin: This is in the module (not an object in a storage library), so I don't think it could have any impact.

mjlescano: Confirming:
> The compiler does not reserve a storage slot for these variables, and every occurrence is replaced by the respective value.
More info: https://docs.soliditylang.org/en/latest/contracts.html#constants

noisekit (approved):

PR #2140, opened 2024-05-06
Bfp permissions

0xjocke:
We need these new permission to be included in the SIP for bfp. 
Holding off merging this so that CORE can have releases without releasing these new permissions


protocol/synthetix/storage.dump.sol
```diff
@@ -417,6 +417,8 @@ library AccountRBAC {
     bytes32 internal constant _PERPS_MODIFY_COLLATERAL_PERMISSION = "PERPS_MODIFY_COLLATERAL";
     bytes32 internal constant _PERPS_COMMIT_ASYNC_ORDER_PERMISSION = "PERPS_COMMIT_ASYNC_ORDER";
     bytes32 internal constant _BURN_PERMISSION = "BURN";
+    bytes32 internal constant _PERPS_PAY_DEBT_PERMISSION = "PERPS_PAY_DEBT";
+    bytes32 internal constant _PERPS_SPLIT_ACCOUNT_PERMISSION = "PERPS_SPLIT_ACCOUNT";
```
davidvuong, line 421: Can we prefix this with `_BFP_xxx` as this is bfp-market specific?

0xjocke: happy to do this if you want. Just a question if we want other perps market to potentially reuse them

davidvuong: I don't think it would be used in other perp markets, it's very bfp-market specific tbh but I'm ok if you want to keep it named the same.

davidvuong:
cc: @noahlitvin and @dbeal-eth so they're also aware of core changes.

davidvuong (approved):
Approved but would like to see a test that reverts when attempting to split another user's account or pay another account's debt.

protocol/synthetix/storage.dump.sol
0xjocke: Added BFP

0xjocke:
> Approved but would like to see a test that reverts when attempting to split another user's account or pay another account's debt.

Added!

PR #2144, opened 2024-05-09
Add snxUSD collateral to minCredit

sunnyvempati:
- snxUSD is escrowed in the core system and is added to the credit capacity which is what is compared with minimumCredit to determine the capacity required for the market.
- in this PR, we account for this by adding the snxUSD owed to traders to minimumCredit

- [x] ~Requires SIP~ https://sips.synthetix.io/sips/sip-390/


leomassazza (approved):
LGTM

PR #2145, opened 2024-05-09
Sip 366 implementation

leomassazza:


protocol/synthetix/contracts/interfaces/IMarketManagerModule.sol
```diff
@@ -221,18 +245,65 @@ interface IMarketManagerModule {
     ) external returns (bool finishedDistributing);
 
     /**
-     * @notice allows for a market to set its minimum delegation time. This is useful for preventing stakers from frontrunning rewards or losses
-     * by limiting the frequency of `delegateCollateral` (or `setPoolConfiguration`) calls. By default, there is no minimum delegation time.
-     * @param marketId the id of the market that wants to set delegation time.
-     * @param minDelegateTime the minimum number of seconds between delegation calls. Note: this value must be less than the globally defined maximum minDelegateTime
+     * @notice allows for a market to set its un-delegation delay time. (See SIP-366). By default, there is no delay for undelegation.
```
kaleb-keny, line 248: nit but delegate

protocol/synthetix/contracts/interfaces/IVaultModule.sol
```diff
@@ -56,14 +144,97 @@ interface IVaultModule {
...
+    function getAccountIntent(
+        uint128 accountId,
+        uint256 intentId
+    )
+        external
+        view
+        returns (
+            uint128 poolId,
+            address collateralType,
+            int256 deltaCollateralAmountD18,
```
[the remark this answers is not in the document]
leomassazza: Just this intent, not accumulated or anything

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +111,160 @@ contract VaultModule is IVaultModule {
...
+
+    /**
+     * @inheritdoc IVaultModule
+     */
+    function deleteIntents(uint128 accountId, uint256[] calldata intentIds) external override {
+        Account.loadAccountAndValidatePermission(accountId, AccountRBAC._DELEGATE_PERMISSION);
+        for (uint256 i = 0; i < intentIds.length; i++) {
+            DelegationIntent.Data storage intent = DelegationIntent.load(intentIds[i]);
+            if (intent.accountId != accountId) {
+                revert InvalidDelegationIntent();
```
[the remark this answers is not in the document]
leomassazza: The check is necesary to not allow for removal of non-expired intents

protocol/synthetix/contracts/interfaces/IVaultModule.sol
```diff
@@ -56,14 +144,97 @@ interface IVaultModule {
...
+     * @dev It will only delete expired intents.
+     * @dev Only the owner of the account, or given the DELEGATE permission can execute this call.
+     */
+    function deleteIntents(uint128 accountId, uint256[] calldata intentIds) external;
+
+    /**
+     * @notice Attempt to delete all expired delegation intents from an account.
+     * @param accountId The id of the account owning the intents.
+     * @dev It will only delete expired intents.
+     * @dev Only the owner of the account, or given the DELEGATE permission can execute this call.
```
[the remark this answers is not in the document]
leomassazza: Done

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +111,160 @@ contract VaultModule is IVaultModule {
...
+            DelegationIntent.Data storage intent = DelegationIntent.load(intentIds[i]);
+            AccountDelegationIntents.getValid(accountId).removeIntent(intent);
+        }
+    }
+
+    /**
+     * @inheritdoc IVaultModule
+     */
+    function deleteAllExpiredIntents(uint128 accountId) external override {
+        Account.loadAccountAndValidatePermission(accountId, AccountRBAC._DELEGATE_PERMISSION);
```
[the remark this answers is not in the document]
leomassazza: Done

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +111,160 @@ contract VaultModule is IVaultModule {
...
+        AccountDelegationIntents.getValid(accountId).cleanAllExpiredIntents();
+    }
+
+    /**
+     * @inheritdoc IVaultModule
+     */
+    function deleteIntents(uint128 accountId, uint256[] calldata intentIds) external override {
+        Account.loadAccountAndValidatePermission(accountId, AccountRBAC._DELEGATE_PERMISSION);
```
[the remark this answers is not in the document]
leomassazza: Done

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -34,49 +36,65 @@ contract VaultModule is IVaultModule {
...
+                accountIntents.netDelegatedAmountPerCollateral[collateralType],
+                accumulatedDelta,
+                currentCollateralAmount
+            );
+        }
 
-        // Conditions for collateral amount
+        uint256 newCollateralAmountD18 = accumulatedDelta > 0
+            ? currentCollateralAmount + (accumulatedDelta).toUint()
+            : currentCollateralAmount - (-1 * accumulatedDelta).toUint();
```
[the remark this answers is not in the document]
leomassazza: It's needed because currentCollateralAmount is an `uint` and accumulatedDelta is `int`. And at that point is negative, so we need to get the modulus and substract from the uint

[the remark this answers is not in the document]
kaleb-keny: maybe a really dumb suggestion 😅  but isn't it possible to just perform the following: 
```
        uint256 newCollateralAmountD18 =  currentCollateralAmount - (-1 * accumulatedDelta).toUint();
```
Since either way the minus is cancelled?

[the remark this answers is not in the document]
leomassazza: We need to use the two branches here since accumulatedDelta can be positive or negative and toUint() will revert if the number is negative. An alternative would be to do:
`uint256 newCollateralAmountD18 =  `(currentCollateralAmount.toInt() + accumulatedDelta).toUint();`

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +111,160 @@ contract VaultModule is IVaultModule {
...
+        uint256[] memory intentIds
+    ) public override {
+        FeatureFlag.ensureAccessToFeature(_DELEGATE_FEATURE_FLAG);
+        for (uint256 i = 0; i < intentIds.length; i++) {
+            DelegationIntent.Data storage intent = DelegationIntent.load(intentIds[i]);
+            if (!intent.isExecutable()) {
+                // Remove the intent.
+                if (intent.intentExpired()) {
+                    AccountDelegationIntents.getValid(accountId).removeIntent(intent);
+                    emit DelegationIntentRemoved(
```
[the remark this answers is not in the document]
leomassazza: If is not executable, it will skip it and emit an event to indicate it was Skipped.

And, if it's expired it will also remove the intent and emit an event to indicate it was removed. 

So, you'll only get the skip event (and not the removed) if the time is too soon to execute this intent, that is, initial delay not reached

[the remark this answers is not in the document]
leomassazza: if due to clarity, you prefer I move Skipped event before this if, I'll do it.

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -34,49 +36,65 @@ contract VaultModule is IVaultModule {
...
+        if (newCollateralAmountD18 > 0) {
+            CollateralConfiguration.requireSufficientDelegation(
+                collateralType,
+                newCollateralAmountD18
+            );
+        }
+        // Check the validity of the collateral amount to be delegated, respecting the caches that track outstanding intents to delegate or undelegate collateral.
         // If increasing delegated collateral amount,
         // Check that the account has sufficient collateral.
         else if (newCollateralAmountD18 > currentCollateralAmount) {
```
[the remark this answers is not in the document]
leomassazza: good call. Fixed

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,149 @@
...
+            self.delegatedCollaterals.add(delegationIntent.collateralType);
+        }
+    }
+
+    function removeIntent(
+        Data storage self,
+        DelegationIntent.Data storage delegationIntent
+    ) internal {
+        if (!self.intentsId.contains(delegationIntent.id)) {
+            return;
```
kaleb-keny, line 83: Shouldn't we revert here? Since the guy is trying to remove an intent that doesn't exist?

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,149 @@
...
+    function intentIdsByPair(
+        Data storage self,
+        uint128 poolId,
+        uint128 accountId
+    ) internal view returns (uint256[] memory intentIds) {
+        return self.intentsByPair[keccak256(abi.encodePacked(poolId, accountId))].values();
+    }
+
+    /**
+     * @dev Cleans all intents related to the account. This should be called upon liquidation.
```
kaleb-keny, line 119: Probably this isn't called upon liquidation, can be called when you want to clean up an account?

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,149 @@
...
+            if (intent.intentExpired()) {
+                removeIntent(self, intent);
+            }
+        }
+    }
+
+    /**
+     * @dev Cleans all intents related to the account. This should be called upon liquidation.
+     */
+    function cleanAllIntents(Data storage self) internal {
```
kaleb-keny, line 134: About this function, i remember speaking to jared about it, and remember him saying, that it would be possible to nuke it all, by just deleting the key, rather than iterating.

protocol/synthetix/contracts/storage/DelegationIntent.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    function processingEndTime(Data storage self) internal view returns (uint32) {
+        (uint32 requiredDelayTime, uint32 requiredWindowTime) = Pool
+            .loadExisting(self.poolId)
+            .getRequiredDelegationDelayAndWindow(self.deltaCollateralAmountD18 < 0);
+
+        // Apply default (forever) window time if not set
+        if (requiredWindowTime == 0) {
+            requiredWindowTime = 86400 * 360; // 1 year
```
kaleb-keny, line 115: I understand we discussed this before, and sorry to bring it up again, but in case a window is shortened, would it be possible to use the min between the shortened window and the one saved when the intent was specified?
Like my thoughts is that we raise the window, or misconfigure to 0, people now have to wait a year to expire. so we need owner to kil those events that can't be executed (due to issue)  nor expired... So we shorten the window... 
I understand the double edge sword... I'll think about this some more, but want to avoid the above scenario of peole having to wait a long time for removing / killing a non-executable intent.

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -420,27 +418,24 @@ library Pool {
...
-                requiredMinDelegateTime = marketMinDelegateTime;
+            Market.Data storage market = Market.load(self.marketConfigurations[i].marketId);
+            uint32 marketDelayTime = isUndelegation
+                ? market.undelegateCollateralDelay
+                : market.delegateCollateralDelay;
+
+            if (marketDelayTime > requiredDelayTime) {
+                requiredDelayTime = marketDelayTime;
+                // Also get the window from the more restrictive market
+                requiredWindowTime = isUndelegation
```
kaleb-keny, line 434: Am not sure does more restrictive window 🤔 ... if it's most strict based on delay, if that's the intention i dont see an issue with it... but maybe fix the comment. I understand the complexity if a market has a 0 window, it means it's less strict that other window configurations, but we have the edge case of 2 year window, and 0 which cause confusion, since 1 year is less strict than 2 years...  For that reason, i'm ok keeping it as is i feel.

protocol/synthetix/contracts/interfaces/IVaultModule.sol
```diff
@@ -56,14 +144,95 @@ interface IVaultModule {
...
+     * @notice Returns details of the requested intent.
+     * @param accountId The id of the account owning the intent.
+     * @param intentId The id of the intents.
+     * @return poolId The id of the pool associated with the position.
+     * @return collateralType The address of the collateral used in the position.
+     * @return deltaCollateralAmountD18 The delta amount of collateral delegated in the position, denominated with 18 decimals of precision.
+     * @return leverage The new leverage amount used in the position, denominated with 18 decimals of precision.
+     * @return declarationTime The time at which the intent was declared.
+     */
+    function getAccountIntent(
```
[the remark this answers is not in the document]
leomassazza: done

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
leomassazza: fixed the comment

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
leomassazza: added nukability

protocol/synthetix/contracts/storage/Pool.sol
leomassazza: done

protocol/synthetix/contracts/modules/core/LiquidationModule.sol
```diff
@@ -114,6 +117,9 @@ contract LiquidationModule is ILiquidationModule {
             liquidationData.amountRewarded
         );
 
+        // Clean any outstanding intents to delegate collateral
+        AccountDelegationIntents.getValid(accountId).cleanAllIntents();
```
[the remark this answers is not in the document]
leomassazza: done

protocol/synthetix/contracts/storage/DelegationIntent.sol
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        return self.declarationTime + requiredDelayTime + requiredWindowTime;
+    }
+
+    function checkIsExecutable(Data storage self) internal view {
```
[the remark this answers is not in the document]
leomassazza: added some views for the intents. Let me know if that is enough

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
leomassazza: Didn't want to revert in this internal function since is used on another ones

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -252,6 +365,277 @@ contract VaultModule is IVaultModule {
...
+            .collateralConfigurations[collateralType]
+            .issuanceRatioD18;
+
+        uint256 effectiveIssuanceRatioD18 = CollateralConfiguration
+            .load(collateralType)
+            .getEffectiveIssuanceRatio(minIssuanceRatioD18);
+
+        uint256 collateralValue = newCollateralAmountD18.mulDecimal(collateralPrice);
+
+        uint256 maxDebt = effectiveIssuanceRatioD18.mulDecimal(collateralValue);
```
kaleb-keny, line 524: I would just try and see if we can tackle the edge case of infinite issuance ratio, but capping the debt to be repaid to the current debt.

protocol/synthetix/storage.dump.sol
```diff
@@ -396,7 +395,7 @@ library Account {
         AccountRBAC.Data rbac;
         uint64 lastInteraction;
         uint64 __slotAvailableForFutureUse;
-        uint128 __slot2AvailableForFutureUse;
+        uint128 currentDelegationIntentsEpoch;
```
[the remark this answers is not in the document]
leomassazza: because of how the storage alignment works. `lastInteraction` being an `uint64`, if the next item is a `uint256` it will need to pad it to keep the alignment (and keep the rest unused). That gives us empty slots that can be reused later and why we had 
```
        uint64 lastInteraction;
        uint64 __slotAvailableForFutureUse;
        uint128 __slot2AvailableForFutureUse;
```
filling the 256 bits

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +110,169 @@ contract VaultModule is IVaultModule {
...
+     * @inheritdoc IVaultModule
+     */
+    function deleteAllExpiredIntents(uint128 accountId) external override {
+        AccountDelegationIntents.getValid(accountId).cleanAllExpiredIntents();
+    }
+
+    /**
+     * @inheritdoc IVaultModule
+     */
+    function deleteExpiredIntents(
```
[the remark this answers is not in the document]
leomassazza: I don't think it matters, just the user wasting some gas deleting discarded intents

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +110,169 @@ contract VaultModule is IVaultModule {
...
+
+                // skip to the next intent
+                continue;
+            }
+
+            // Ensure the intent is valid.
+            if (intent.accountId != accountId) revert InvalidDelegationIntent();
+
+            // Ensure the intent is within the processing window.
+            intent.checkIsExecutable();
```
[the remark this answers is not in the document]
leomassazza: done (need to test it, but makes sense)

protocol/synthetix/contracts/interfaces/IVaultModule.sol
```diff
@@ -56,14 +164,181 @@ interface IVaultModule {
...
+     * @return accumulatedIntentDelta The total amount of collateral intended to be delegated that is not expired, denominated with 18 decimals of precision.
+     */
+    function getExecutableDelegationAccumulated(
+        uint128 accountId,
+        uint128 poolId,
+        address collateralType
+    ) external view returns (int256 accumulatedIntentDelta);
+
+    /**
+     * @notice Returns the amount of debt to repay in order to an intent to reduce the delegated collateral will meet the issuance ratio.
```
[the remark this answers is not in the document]
leomassazza: done

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -93,73 +110,169 @@ contract VaultModule is IVaultModule {
...
+                    emit DelegationIntentRemoved(
+                        intent.id,
+                        accountId,
+                        intent.poolId,
+                        intent.collateralType
+                    );
+                }
+
+                // skip to the next intent
+                continue;
```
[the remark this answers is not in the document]
leomassazza: the expired check is to delete the intent that we just found out was expired.

protocol/synthetix/contracts/modules/core/VaultModule.sol
leomassazza: fixed in last commit here: https://github.com/Synthetixio/synthetix-v3/blob/05cd81fab29ae65d8711ec2a88e9ffa8f58317a1/protocol/synthetix/contracts/modules/core/VaultModule.sol#L520-L522

protocol/synthetix/contracts/interfaces/IVaultModule.sol
```diff
@@ -56,14 +164,181 @@ interface IVaultModule {
      *
      * Emits a {DelegationUpdated} event.
      */
-    function delegateCollateral(
```
noahlitvin, line 59: Is it possible to have the interfaces set up where atomic delegation can still be possible and the interface doesn't change? Like this function could be the "process" function if necessary a delay is set?

There's also an abstraction where everything is async, and atomic can be done with a batch of two tx's when the delay is configured to 0.

kaleb-keny: i belive switching the delay and window to 0 leads to atomic processing with the same interface

protocol/synthetix/contracts/modules/core/MarketManagerModule.sol
```diff
@@ -329,38 +328,101 @@ contract MarketManagerModule is IMarketManagerModule {
...
+    function getUndelegateCollateralDelay(
+        uint128 marketId
+    ) external view override returns (uint32) {
+        return Market.load(marketId).undelegateCollateralDelay;
+    }
+
+    /**
+     * @inheritdoc IMarketManagerModule
+     */
+    function setUndelegateCollateralWindow(
```
dbeal-eth, line 357: please combine these individual `setUndelegateCollateral...` functions into a single setter that takes a tuple/structure.

its likely that such values will be always set/changed at the same time, reduces surface area, and 

same goes for getter.

protocol/synthetix/contracts/modules/core/PoolModule.sol
```diff
@@ -148,7 +148,6 @@ contract PoolModule is IPoolModule {
     ) external override {
         Pool.Data storage pool = Pool.loadExisting(poolId);
         Pool.onlyPoolOwner(poolId, ERC2771Context._msgSender());
-        pool.requireMinDelegationTimeElapsed(pool.lastConfigurationTime);
```
dbeal-eth, line 151: this line existed in order to prevent pool owners (or, single user-pool owners, which is something users can do, create a pool for themselves) from being exempt from the min delegation timeout

in order to effectively apply the delegation limitations to protect the pool, the same intent system must be developed on the market side as well.

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,170 @@
...
+                accountDelegationIntents.delegationIntentsEpoch != delegationIntentsEpoch)
+        ) {
+            revert IVaultModule.InvalidDelegationIntent();
+        }
+    }
+
+    /**
+     * @dev Returns the account delegation intents stored at the specified account id. Checks if it's valid
+     */
+    function getValid(uint128 accountId) internal returns (Data storage accountDelegationIntents) {
```
dbeal-eth, line 72: the name of this function and understanding the difference between `getValid` and `loadValid` seems extremely difficult. maybe a better name is `loadWithInit` or something?

protocol/synthetix/contracts/storage/DelegationIntent.sol
```diff
@@ -0,0 +1,166 @@
...
+     * - processingStartTime: Timestamp from which the intent can start being processed.
+     * - processingEndTime: Timestamp after which the intent cannot be processed.
+     *
+     * The intent can be processed only between processingStartTime and processingEndTime.
+     */
+    struct Data {
+        /**
+         * @notice An incrementing id (nonce) to ensure the  uniqueness of the intent and prevent replay attacks
+         */
+        uint256 id;
```
dbeal-eth, line 37: this field could be done without--see other comments

delegation is already very expensive so every storage slot we can reduce is huge help

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,170 @@
...
+    using SetUtil for SetUtil.UintSet;
+    using SetUtil for SetUtil.AddressSet;
+    using DelegationIntent for DelegationIntent.Data;
+    using Account for Account.Data;
+
+    struct Data {
+        uint128 accountId;
+        uint128 delegationIntentsEpoch; // nonce used to nuke previous intents using a new era (useful on liquidations)
+        SetUtil.UintSet intentsId;
+        mapping(bytes32 => SetUtil.UintSet) intentsByPair; // poolId/collateralType => intentIds[]
```
dbeal-eth, line 26: can we just embed `DelegationIntent.Data[]` type right here? then we:
1. don't need to store `intentsId`
2. don't need to store `id` inside of the `DelegationIntent`
3. since intents are added in timestamp order, we can wipe out expired intents very cleanly without searching the full intents list by creating another field `lastActiveIntent` and incrementing that.
4. if we make it a rule that intents have to be processed in the order they are submitted, this same field can be used to invalidate intents as we go

also if an intent is "cancelled" by the owner before it can be processed, we can add a field to the intent "cancelled, which only incurs a cost on intent processing and nowhere else

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,170 @@
...
+        // accounting for the intents collateral delegated
+        // Per Collateral
+        SetUtil.AddressSet delegatedCollaterals;
```
dbeal-eth, line 29: if we can avoid storing this and only track it off-chain, save about 2-3 storage slots!

dbeal-eth: not sure if this one is possible tho

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
```diff
@@ -0,0 +1,170 @@
...
+        mapping(address => int256) netDelegatedAmountPerCollateral; // collateralType => net delegatedCollateralAmount
+    }
+
+    /**
+     * @dev Returns the account delegation intents stored at the specified account id.
+     */
+    function load(
```
dbeal-eth, line 36: if we got rid of this load function and instead embedded this data structure into the `Account`, it would reduce load complexity and heiarchally store data in a convenient manner

dbeal-eth (changes requested):
please see my storage optimization comments in particular.

protocol/synthetix/contracts/interfaces/IVaultModule.sol
dbeal-eth: i have same feedback

kaleb-keny:
> please see my storage optimization comments in particular.
> 
> the removal of `minDelegationTime` is likely to have ripple effects elsewhere.
> 
> also, I already raised this with kaleb, but due to the removal of the `minDelegationTime` check on the PoolModule, it is no longer possible to enforce delegation time for users who want to frontrun on a pool level. We need to make sure the same rules apply both to pool owners, and to vault delegators.

Hey @leomassazza , there needs to be a minimum delay / window enforced on pool owners as db mentioned, otherwise one can setup a pool with 0 delay/window parameters in order to front-run price updates. That said, if the protocol owner, can push any parameters regardless of the minimum specified, on specific LP's, it would be great, as it allows for special cases where 0 delay/window are possible. But this isn't necessary in the near term and can be thought about in the future, dealing with just the minimum delay/window is fine for now.

Hey @dbeal-eth thank just to be on the same page here, you are fine with having a minimum delay / window applied on pool owners by the core right?

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
kaleb-keny: All in favor optimizing the removal of id @dbeal-eth   and replacing with timestamps, in case the lp/account-id/market is the key to holding the stored intents... In order to avoid different timestamps stepping over each other. The the rule to make intents processed in the order they are submitted can lead to issues with vaults... Assume the delay / window are configured differently for delegate / undelegate and a user specifies a undelegation on a vault and another user a delegate.. They can lead to stepping over each other... So this would require 2 intent data's right? One for delegation and another for undelegation?
Also another scenario I was thinking about is assume you have an intent that can't be expired or executed, and is in limbo until expiry, Example caps. This leads to kind of bricking until parameters are updated... So enforcing sequentiality I thought about this and opted to not ask for it to be employed purposefully....

dbeal-eth: yea 3 is literally intended tobe a simplification for clients by allowing for the contract to be able to handle everything effieicntly.

regarding #4, if delegate/undelegate have different window or delay, then we can separate the lists by "operation" (its already separated by vault). Tho to simplify things I dont see a lot of use in allowing different window limits for delegate/undelegate...

regarding update of parameters... I mean these settings apply immediately to all pending intents regardless of when they are created and seems like it would be an issue with either scheme, so seems like a small issue if anything comparing to the benefits of the optimizations.

dbeal-eth:
> Hey @dbeal-eth thank just to be on the same page here, you are fine with having a minimum delay / window applied on pool owners by the core right, in order to do away with minDelegateTime?

yea I think this is fine. but I am *way* more concerned about the breaking interface changes. should not be made lightly...

protocol/synthetix/contracts/storage/AccountDelegationIntents.sol
kaleb-keny: There is also the situation of an intent that is neither executable and the window is not out for it's expiry, blocking the vault from executing any other intent. Aside from parameter changes, there are things that could go wrong and requiring sequential execution for async intents I feel is too fragile for the sake of optimization.

kaleb-keny:
> > Hey @dbeal-eth thank just to be on the same page here, you are fine with having a minimum delay / window applied on pool owners by the core right, in order to do away with minDelegateTime?
> 
> yea I think this is fine. but I am _way_ more concerned about the breaking interface changes. should not be made lightly...

Awesome thank you, yeah I get you, given that the core staking is yet to be integrated across different protocols I think we can risk it at this point, giving dhedge the necessary time. The reason i insist on this is that the intention is to lower the intent delay to 30 minutes and window to 1 hour. So having a minDelegateTime set to zero if it's kept. Which might lead to issues in having to deal with legacy code there, in case someone can setup their own pool without intents and just use the old interface for delegation, with 0 minDelegateTime...

protocol/synthetix/contracts/storage/Market.sol
```diff
@@ -138,11 +138,16 @@ library Market {
...
+         */
+        uint32 __unusedLegacyStorageSlot;
+        uint32 undelegateCollateralDelay;
+        uint32 undelegateCollateralWindow;
+        uint32 delegateCollateralDelay;
+        uint32 delegateCollateralWindow;
         uint32 __reservedForLater1;
         uint64 __reservedForLater2;
-        uint64 __reservedForLater3;
-        uint64 __reservedForLater4;
```
mjlescano, line 145: Listing storage changes as this looks risky:

1. Deprecating `uint32  minDelegateTime` as `uint32 __unusedLegacyStorageSlot` (This should be marked with something like "DO NOT USE", because maybe it has data). ⚠️ 
2. Renaming `uint32 __reservedForLater1` to `uint32 undelegateCollateralDelay` ✅ 
3. Expanding slot `uint64 __reservedForLater3` to `uint32 undelegateCollateralWindow` & `uint32 delegateCollateralDelay` ✅
4. Renaming and **Resizing** `uint64 __reservedForLater4` to `uint32 delegateCollateralWindow`.

Please confirm @leomassazza, but this needs to add a `uint32 __reservedForLaterX` so it doesn't leak into the next slot for `uint256 minLiquidityRatioD18`

protocol/synthetix/contracts/storage/VaultEpoch.sol
```diff
@@ -64,12 +64,7 @@ library VaultEpoch {
          * and directly when users mint or burn USD, or repay debt.
          */
         mapping(uint256 => int256) consolidatedDebtAmountsD18;
-        /**
-         * @dev Tracks last time a user delegated to this vault.
-         *
-         * Needed to validate min delegation time compliance to prevent small scale debt pool frontrunning
-         */
-        mapping(uint128 => uint64) lastDelegationTime;
+        mapping(uint128 => uint64) __unused_legacy_slot;
```
mjlescano, line 67: Should be marked as `do not use`, because maybe in the future someone reads "unused" and think it has no data.

mjlescano (changes requested):

protocol/synthetix/contracts/storage/VaultEpoch.sol
mjlescano: Ah! For consistency it should be __camelCase

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -34,8 +36,12 @@ contract VaultModule is IVaultModule {
     using SafeCastU256 for uint256;
     using SafeCastI128 for int128;
     using SafeCastI256 for int256;
+    using AccountDelegationIntents for AccountDelegationIntents.Data;
+    using DelegationIntent for DelegationIntent.Data;
+    using Account for Account.Data;
 
     bytes32 private constant _DELEGATE_FEATURE_FLAG = "delegateCollateral";
+    bytes32 private constant _TWO_STEPS_DELEGATE_FEATURE_FLAG = "twoStepsDelegateCollateral";
```
dbeal-eth, line 44: really we should be naming all feature flags the same as the function they guard

```
bytes32 private constant _DECLARE_DELEGATE_FEATURE_FLAG = "declareIntentToDelegateCollateral";
bytes32 private constant _PROCESS_DELEGATE_FEATURE_FLAG = "processIntentToDelegateCollateral";
```

having a flag to disable both of them at the same time icnraseses code complexity and can be done almost as eaily by simply specifying both of the function names for deny all feature flag.
that way its straightforward in an emergency situation to block any function by simply using its name.

dbeal-eth (approved):
seems ok

we need to do some gas profiling tho because the way this has been designed may work well enough to cover the somewhat edge case of the needs of dhedge, but it does so by using an extreme amoun tof gas and storing excessive amounts of data imo. I am very certain there is a better way.

protocol/synthetix/contracts/modules/core/VaultModule.sol
leomassazza: fixed, though couldn't use the full function name due to naming length (bytes32)

PR #2153, opened 2024-05-20
Testnet deployment fixes

0xjocke:
Some minor changes discovered while working on testnet deployemnet.

-  Make `marketId` part of `ConfigureByMarketParameters`, this makes the cannon file in synthetix-deployments behave a bit nicer. 
-  Include `setRewardDistributorImplementation` in the production cannonfile. We already have it in the cannon test file, and I think it make sense to have it here rather than in synthetix-deployment as it should behave differently on different networks


davidvuong (approved):
LGTM

PR #2154, opened 2024-05-21
fix for pdao incident

jmzwar:


protocol/synthetix/subgraph/mainnet/handleRewardsDistributed.ts
```diff
@@ -22,12 +23,26 @@ export function handleRewardsDistributed(event: RewardsDistributed): void {
...
   );
-  let rewardsDistributor = RewardsDistributor.load(event.params.distributor.toHex());
+
+  let rewardsDistributor: RewardsDistributor | null;
+
+  // Pdao situation
+  const snxDistributor = Address.fromString('0x45063dcd92f56138686810eacb1b510c941d6593');
+  const pdao = Address.fromString('0xbb63ca5554dc4ccaca4edd6ecc2837d5efe83c82');
+
+  if (event.params.start.toString() == '1716127200' && event.params.distributor == pdao) {
```
noisekit, line 33: I believe we can get rd from event itself. No need to hardcode blocks.

jmzwar: The pdao or snx one? That was the point of this we had pdao when it should have been snx.
I think it is better to leave it this way for readability sake

noisekit: Yeah ok. We will fix it in core I guess. Hopefully this is a one-off situation

noisekit (approved):

PR #2156, opened 2024-05-24
feat(Vault): added vault snapshots by day and week

fritzschoff:


noisekit (changes requested):
1. Don't see where we subscribe to event
2. Need tests

fritzschoff:
> 1. Don't see where we subscribe to event
> 2. Need tests

1. In the the handleDelegatinoUpdated event
2. alright, incoming

protocol/synthetix/subgraph/mainnet/handleDelegationUpdated.ts
```diff
@@ -51,4 +53,7 @@ export function handleDelegationUpdated(event: DelegationUpdated): void {
   vault.updated_at_block = event.block.number;
   vault.save();
   position.save();
+
+  createVaultSnapshotByDay(vault);
```
noisekit, line 57: Ok can see here.
Still need tests

protocol/synthetix/subgraph/mainnet/handleDelegationUpdated.ts
```diff
@@ -9,19 +14,50 @@ export function handleDelegationUpdated(event: DelegationUpdated): void {
...
+    vault.created_at = event.block.timestamp;
+    vault.created_at_block = event.block.number;
+    vault.collateral_amount = event.params.amount.toBigDecimal();
+    vault.collateral_type = event.params.collateralType.toHex();
+    vault.pool = event.params.poolId.toString();
+  } else {
+    if (position) {
+      const isIncreasing = event.params.amount.toBigDecimal().gt(position.collateral_amount);
+      if (isIncreasing) {
+        log.info('INCREASING {} {}', [
```
noisekit, line 34: guess we dont really need logs here

protocol/synthetix/subgraph/mainnet/handleDelegationUpdated.ts
```diff
@@ -9,19 +14,50 @@ export function handleDelegationUpdated(event: DelegationUpdated): void {
...
+          event.params.amount.toString(),
+          position.collateral_amount.toString(),
+        ]);
+        vault.collateral_amount = vault.collateral_amount.plus(
+          event.params.amount.toBigDecimal().minus(position.collateral_amount)
+        );
+      } else {
+        vault.collateral_amount = vault.collateral_amount.plus(event.params.amount.toBigDecimal());
```
noisekit, line 42: when we modify collateral we actually set value, not "add" collateral.
So the vault value needs to increase by difference between (`event.params.amount` and previous collateral amount).

To simplify logic and reduce ifs I suggest setting "previous amount" to 0 if position does not exist and then all the calculations will be identical.

```js
if (!vault) {
 // create new vault *without* collateral_amount
  previous_vault_amount = 0
} else {
  previous_vault_amount = vault.collateral_amount
}

if (!position) {
 // create new position *without* collateral_amount
 previous_position_amount = 0
} else {
  previous_position_amount = position.colateral_amount 
}

// now fill all the values that are common for update/create

amount_delta = event.params.amount - previousCollateral

vault.collateral_amount = previous_vault_amount + amount_delta
position.collateral_amount = event.params.amount 

```

nested if/else are now so complicated it is pretty much impossible to understand what is going on.

protocol/synthetix/subgraph/mainnet/vaultSnapshotByDay.ts
```diff
@@ -0,0 +1,35 @@
...
+    vaultSnapshotByDay.collateral_type = vaultWithLatestValues.collateral_type;
+    vaultSnapshotByDay.pool = vaultWithLatestValues.pool;
+  }
+  vaultSnapshotByDay.updated_at = event.block.timestamp;
+  vaultSnapshotByDay.updated_at_block = event.block.number;
+  vaultSnapshotByDay.updates_in_period = vaultSnapshotByDay.updates_in_period.plus(
+    BigInt.fromI32(1)
+  );
+  vaultSnapshotByDay.collateral_amount = vaultWithLatestValues.collateral_amount;
+  log.info('{}', [vaultSnapshotByDay.updates_in_period.toString()]);
```
noisekit, line 32: Do we need logs?

protocol/synthetix/subgraph/tests/CoreProxy.test.ts
```diff
@@ -46,6 +49,11 @@ describe('CoreProxy', () => {
   test('handleMarketDeposits', handleMarketDepositsTest);
   test('handleMarketWithdrawals', handleMarketWithdrawalsTest);
   test('getISOWeekNumber', getISOWeekNumberTest);
+  test('vaultSnapshotByDay', vaultSnapshotByDayTest);
+  test('vaultSnapshotByWeek', vaultSnapshotByWeekTest);
+  test('vaultSnapshotByMonth', vaultSnapshotByMonthTest);
+  test('vaultSnapshotByYear', vaultSnapshotByYearTest);
+  test('caultSnapshotByWeek', marketSnapshotByWeekTest);
```
noisekit, line 56:
```suggestion
```
there is already vaultSnapshotByWeek above and marketSnapshotByWeek below

protocol/synthetix/subgraph/mainnet/vaultSnapshotByDay.ts
fritzschoff: wut, dont have that?

protocol/synthetix/subgraph/mainnet/handleDelegationUpdated.ts
noisekit: "my version" is not tested and just a short summary of the idea.
Please remove 3 levels of `if`s, comments are not enough to understand what is going on there. Logic is too complex.
DelegationUpdated event has actual new amount of what is delegated, not the delta. So `position.collateral_amount = event.params.amount` should be correct in this case. If tests are failing then tests need to be fixed.

See the source: https://github.com/Synthetixio/synthetix-v3/blob/main/protocol/synthetix/contracts/interfaces/IVaultModule.sol#L30

fritzschoff: gotcha, updated the tests

protocol/synthetix/subgraph/mainnet/handleDelegationUpdated.ts
```diff
@@ -9,21 +14,43 @@ export function handleDelegationUpdated(event: DelegationUpdated): void {
...
+    vault.collateral_amount = BigDecimal.fromString('0');
+    vault.collateral_type = event.params.collateralType.toHex();
+    vault.pool = event.params.poolId.toString();
+  }
+
+  let previous_position_amount = BigDecimal.fromString('0');
+  if (position) {
+    previous_position_amount = position.collateral_amount;
+    let amount_delta = event.params.amount.toBigDecimal().minus(previous_position_amount);
+    vault.collateral_amount = previous_position_amount.plus(amount_delta);
```
noisekit, line 36:
```suggestion
    vault.collateral_amount = vault.collateral_amount.plus(amount_delta);
```
Vault is global and tracks all positions, not just one.

protocol/synthetix/subgraph/tests/vaultSnapshotByDay.ts
```diff
@@ -0,0 +1,166 @@
...
+      BigInt.fromI32(1),
+      Address.fromString(sender),
+      BigInt.fromI32(500),
+      BigInt.fromI32(1),
+      now + oneDay,
+      now
+    )
+  );
+
+  // assert.fieldEquals(
```
noisekit, line 130: uncomment tests maybe?

protocol/synthetix/subgraph/tests/handleDelegationUpdated.ts
```diff
@@ -61,5 +62,18 @@ export default function test(): void {
   assert.fieldEquals('Position', `1-1-${address}`, 'collateral_amount', '10000');
   assert.fieldEquals('Vault', `1-${address}`, 'updated_at', (now + 1000).toString());
   assert.fieldEquals('Vault', `1-${address}`, 'updated_at_block', now.toString());
-  assert.fieldEquals('Vault', `1-${address}`, 'collateral_amount', '-5354');
+  assert.fieldEquals('Vault', `1-${address}`, 'collateral_amount', '10000');
+
+  const newDelegatioNUpdatedEvent3 = createDelegationUpdateEvent(
```
noisekit, line 67: odd casing... here and above too
```suggestion
  const newDelegationUpdatedEvent3 = createDelegationUpdateEvent(
```

noisekit: can probably just wrap it in `handleDelegationUpdated` and not add new variables

protocol/synthetix/subgraph/tests/handleDelegationUpdated.ts
```diff
@@ -61,5 +62,18 @@ export default function test(): void {
...
+    BigInt.fromI32(1),
+    Address.fromString(address),
+    BigInt.fromI32(5000),
+    BigInt.fromI32(10),
+    now + 1000,
+    now
+  );
+  handleDelegationUpdated(newDelegatioNUpdatedEvent3);
+
+  assert.fieldEquals('Vault', `1-${address}`, 'collateral_amount', '15000');
```
noisekit, line 78:
```suggestion
  assert.fieldEquals('Vault', `1-${address}`, 'collateral_amount', '5000');
```
We have basically one staker that updated delegation from 10000 to 5000 (so delta is `-5000`, vault was 10000 before, so new vault amount should be 5000 now, not 15000

noisekit: Ideally this test now needs to have at least two different addresses with different positions so we can see that vault collateral amount is different from each individual position amount

fritzschoff: no, different account id

fritzschoff: so I think it should add it

fritzschoff: > Ideally this test now needs to have at least two different addresses with different positions so we can see that vault collateral amount is different from each individual position amount

yeah with account id set to 2 in event 3 we have that case

noisekit: Yeah ok. It is not clear what all the args are. Would be nice to have it written down like

```
var accountId1 = 2
createDelegationUpdateEvent(
    accountId1,
    smth,
    address1,
    ...
```

Another thing to do is to remove all the type conversions from test itself. They should be done in event creator. You can see all the example in existing event creators and tests. They do not have type casting in test code itself

noisekit: image

noisekit: look at spot/perps subgraph tests. This particular one was quite old and was not rewritten by me at the time of all the refactorings.

noisekit: Reading test now is hard, I could not figure out from test code that there were multiple positions. And all the new additions need to follow much more readable standard instead of adding a lot of a much less understandable legacy code.

fritzschoff: > look at spot/perps subgraph tests. This particular one was quite old and was not rewritten by me at the time of all the refactorings.

cant follow you? This is not about blaming someone. Refactoring the type conversions is not super urgent and important and can be tackled later on

noisekit (approved):

protocol/synthetix/subgraph/tests/handleDelegationUpdated.ts
noisekit: Not adding code that needs refactoring and adding code that does not need immediate refactoring is the best approach.

noisekit: "not super urgent" means it will stay like this forever

fritzschoff:
https://linear.app/ccsnx/issue/SNX-164/clean-up-subgraph-tests ticket for ref

PR #2161, opened 2024-06-03
Feature/collateral sg

jmzwar:
Read name, symbol and decimals into collateral type definition


protocol/synthetix/subgraph/optimism-mainnet/handleCollateralConfigured.ts
```diff
@@ -0,0 +1,33 @@
...
+  collateralType.updated_at = event.block.timestamp;
+  collateralType.updated_at_block = event.block.number;
+
+  const tokenDefinition = TokenDefinition.fromAddress(event.params.collateralType);
+
+  if (tokenDefinition !== null) {
+    collateralType.name = tokenDefinition.name;
+    collateralType.symbol = tokenDefinition.symbol;
+    collateralType.decimals = tokenDefinition.decimals;
+  }
```
0xjocke, line 30: Maybe we want an else with an error log here?

protocol/synthetix/subgraph/base-sepolia-andromeda/collateralDefinition.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// Initialize a Token Definition with the attributes
+export class TokenDefinition {
+  address: Address;
+  symbol: string;
+  name: string;
+  decimals: BigInt;
+
+  // Get all tokens with a static defintion
+  static getStaticDefinitions(): Array<TokenDefinition> {
```
noisekit, line 11: Why do we need static definitions for tokens? This instantly adds the hard dependency on graph for any new collateral additions. We should pull collateral configurations in the ui and do the matching in the ui and not in subgraph

noisekit (changes requested):

protocol/synthetix/subgraph/optimism-mainnet/handleCollateralConfigured.ts
noisekit: We should jot have this info in collateral type in subgraph. This info can be retrieved on chain an can be done in the ui directly without introduction of this gard dependency on subgraph updates for any collateral changes

protocol/synthetix/subgraph/subgraph.base-mainnet-andromeda.yaml
```diff
@@ -52,10 +52,6 @@ dataSources:
           handler: handlePoolConfigurationSet
         - event: MarketRegistered(indexed address,indexed uint128,indexed address)
           handler: handleMarketCreated
-        - event: MarketUsdDeposited(indexed uint128,indexed address,uint256,indexed address,int128,int128,uint256,uint256)
-          handler: handleMarketUsdDeposited
-        - event: MarketUsdWithdrawn(indexed uint128,indexed address,uint256,indexed address,int128,int128,uint256,uint256)
-          handler: handleMarketUsdWithdrawn
```
0xjocke, line 58: Was this removed on purpose? Should it be removed from here too https://github.com/Synthetixio/synthetix-v3/blob/fa085945b21b2b1a23ca953d45b28d5389112e95/protocol/synthetix/subgraph/base-mainnet-andromeda/index.ts#L10C27-L10C51

jmzwar: Yes, these events weren't part of the CoreProxy

protocol/synthetix/subgraph/optimism-mainnet/handleCollateralConfigured.ts
jmzwar: Yeah good idea

protocol/synthetix/subgraph/base-sepolia-andromeda/collateralDefinition.ts
jmzwar: Yes it does. 
We are pulling them in the UI now already, this simplifies things for loading across diff networks.

noisekit: It does not. Because it means that we now lock in multiple things in the chain of updates. After deployment we would need to redeploy subgraph, do full indexing and then things will start working in ui. If we just pull data from contracts then after deployment update ui will pick things up automatically

noisekit: I am 100% against this change.

protocol/synthetix/subgraph/mainnet/collateralDefinition.ts
```diff
@@ -0,0 +1,39 @@
+import { Address, BigInt } from '@graphprotocol/graph-ts';
+
+// Initialize a Token Definition with the attributes
+export class TokenDefinition {
```
0xjocke, line 4: This file can be removed now?

jmzwar: Yep

0xjocke (approved):

protocol/synthetix/subgraph/mainnet/handleCollateralConfigured.ts
```diff
@@ -17,5 +20,18 @@ export function handleCollateralConfigured(event: CollateralConfigured): void {
   collateralType.min_delegation = event.params.config.minDelegationD18.toBigDecimal();
   collateralType.updated_at = event.block.timestamp;
   collateralType.updated_at_block = event.block.number;
+
+  const erc20Contract = ERC20.bind(event.params.collateralType);
+
+  if (erc20Contract) {
+    collateralType.name = erc20Contract.name();
```
noisekit, line 27: Normally I would say we should avoid calling contracts from event handlers as it is incredibly slow comparing to taking events values.
In this case it is less of an issue as this particular event happens very infrequently.

I am still convinced we should not have this info in subgraph though as this info can be easily retrieved from the UI and we should avoid using subgraphs for anything that is possible (and feasible) to do with contracts directly from the UI.

Still not convinced this is necessary.

noisekit: The idea is simple: only use event values in graph handlers. And add contrast calls only if there is absolutely no other way accessing data.
We already have v2 subgraph that did not consider any restrictions.

PR #2165, opened 2024-06-09
RewardsDistributor management improvements + new features

dbeal-eth:
need to write tests and etc but this look like it will work


protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -79,11 +81,14 @@ contract RewardsManagerModule is IRewardsManagerModule {
...
 
-				if (collateralType == address(0)) {
-						pool.rewardsToVaults[rewardId].distributor = IRewardDistributor(distributor);
-				} {
-						pool.vaults[collateralType].rewards[rewardId].distributor = IRewardDistributor(distributor);
-				}
+        if (collateralType == address(0)) {
+            pool.rewardsToVaults[rewardId].distributor = IRewardDistributor(distributor);
+        }
+        {
```
noisekit, line 87: do we expect `else` here?

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -372,6 +385,65 @@ library Pool {
...
+    ) internal returns (uint256[] memory rewards, address[] memory distributors) {
+        bytes32[] memory poolRewardIds = new bytes32[](self.rewardIds.length());
+
+        {
+            bytes32 actorId = pos.collateralType.toBytes32();
+            uint256 totalSharesD18 = self.vaultsDebtDistribution.totalSharesD18;
+            uint256 vaultSharesD18 = self.vaultsDebtDistribution.getActorShares(actorId);
+
+            uint256 numRewards = self.rewardIds.length();
+            for (uint256 i = 0; i < numRewards; i++) {
```
noisekit, line 403: As `updateRewardsToVaults` gets called for any vault "action" this is going to increase gas by a bit for every vault added to the pool? So the more collaterals we have the more and more gas we are going to consume for every delegation change

protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
dbeal-eth: Yes

protocol/synthetix/contracts/storage/Pool.sol
dbeal-eth: No the loop is on number of RDs in the pool, not number if collateral types/vaults in the pool. so the gas increase is caused by the addition of any pool level RDs, which is inevitable considering this is the ask for the change

noisekit: Yeah sure. I meant to write rd not collateral.

noisekit: would be very good to see a gas impact for arbitrum and base deployments (with 4 and 2 distributors deployed on those).

noisekit (approved):
This looks good to me. Approved, but please get the auditors approval before merge.

protocol/synthetix/test/integration/modules/core/RewardsManagerModule.test.ts
```diff
@@ -668,6 +670,39 @@ describe('RewardsManagerModule', function () {
...
+            rewardAmount,
+            ethers.utils.parseEther('1')
+          )
+      ).wait();
+
+      const availableRewards = await systems()
+        .Core.connect(user1)
+        .getAvailableRewards(2222, poolId, collateralAddress(), RewardDistributorPoolLevel.address);
+
+      console.log('POST AVAILABLE REWARDED', availableRewards);
```
noisekit, line 705: Asert?

auxiliary/RewardsDistributor/cannonfile.toml
```diff
@@ -38,7 +33,6 @@ artifact = "RewardsDistributor"
 args = [
     "<%= settings.rewardManager %>",
     "<%= settings.poolId %>",
-    "<%= settings.collateralType %>",
```
noisekit, line 41: Can we keep args without breaking change and accept zero address instead please

auxiliary/RewardsDistributor/src/RewardsDistributor.sol
```diff
@@ -11,15 +11,18 @@ import {ERC20Helper} from "@synthetixio/core-contracts/contracts/token/ERC20Help
...
     error NotEnoughRewardsLeft(uint256 amountRequested, uint256 amountLeft);
     error NotEnoughBalance(uint256 amountRequested, uint256 currentBalance);
 
     using ERC20Helper for address;
+    using SafeCastU256 for uint256;
+    using SafeCastI256 for int256;
 
     address public rewardManager;
     uint128 public poolId;
-    address public collateralType;
```
noisekit, line 22: We still want to have per collateral distributors available. Please dont remove this functionality

protocol/synthetix/test/integration/modules/core/RewardsManagerModule.test.ts
dbeal-eth: sorry no that was just for sanity debug. I removed the console log and the getavailablerewards call

auxiliary/RewardsDistributor/cannonfile.toml
dbeal-eth: its not a breaking change because its for the constructor of the contract, and the functionality of the contract has significantly changed in an of itself. It might as well be a new contract. since the contract no longer pays any attention to collateral type, I went ahead and removed the arg. On the contrary, I think this should be useful in creating awareness of the change downstream in case its forgotten about.

auxiliary/RewardsDistributor/src/RewardsDistributor.sol
dbeal-eth: before removal, the storage of the value here was only for verification against when calling the `distributeRewards` function. the `distributeRewards` function below still takes in `collateralType` as an argument, allowing for collateral to be distributed to any vault in the pool or at the pool which it is registered to on the core system. since it was only used for that verification before, the storage has been removed because the value isnt used anywhere anymore.

noisekit: Ok makes sense thanks 🙏

auxiliary/RewardsDistributor/cannonfile.toml
noisekit: Got it, thanks

noisekit (approved):

protocol/synthetix/contracts/modules/core/RewardsManagerModule.sol
```diff
@@ -217,30 +300,35 @@ contract RewardsManagerModule is IRewardsManagerModule {
...
-        bytes32 rewardId = _getRewardId(poolId, collateralType, rewardsDistributor);
         if (!rewardIds.contains(rewardId)) {
             revert ParameterError.InvalidParameter(
                 "poolId-collateralType-distributor",
                 "reward is not registered"
             );
         }
 
-        RewardDistribution.Data storage reward = pool.vaults[collateralType].rewards[rewardId];
+        //RewardDistribution.Data storage reward = pool.vaults[collateralType].rewards[rewardId];
```
noisekit, line 319:
```suggestion
```

protocol/synthetix/contracts/storage/Pool.sol
```diff
@@ -372,6 +385,87 @@ library Pool {
...
+        Vault.PositionSelector memory pos
+    ) internal returns (uint256[] memory rewards, address[] memory distributors, uint256) {
+        bytes32[] memory poolRewardIds = new bytes32[](self.rewardIds.length());
+
+        {
+            uint256 actorId = pos.collateralType.to256();
+            uint256 totalSharesD18 = self.vaultsDebtDistribution.totalSharesD18;
+            uint256 vaultSharesD18 = self.vaultsDebtDistribution.getActorShares(bytes32(actorId));
+
+            //uint256 numRewards = self.rewardIds.length();
```
noisekit, line 402:
```suggestion
```

dbeal-eth:
auditor has greenlit

PR #2166, opened 2024-06-09
Multi-collateral support for Rewards Distributor

dbeal-eth:
- update rewards distributor to work with many collateral types in same pool
- cancelled amount support

would also work with pool level distribution


auxiliary/RewardsDistributor/foundry.toml
```diff
@@ -1,7 +1,7 @@
 [profile.default]
 src = "src"
 out = "out"
-libs = []
+libs = ["lib"]
```
noisekit, line 4: This is unnecessary as we only use node_modules and not using git submodules.

```suggestion
libs = []
```

auxiliary/RewardsDistributor/test/RewardsDistributorTest.sol
```diff
@@ -68,24 +76,23 @@ contract RewardsDistributorTest is Test {
...
             poolId,
-            collateralType,
             payoutToken,
             SNX.decimals(),
             name
         );
     }
 
-    function test_constructor_arguments() public {
+    function test_constructor_arguments() public view {
```
noisekit, line 93:
```suggestion
    function test_constructor_arguments() public {
```

auxiliary/RewardsDistributor/test/RewardsDistributorTest.sol
```diff
@@ -267,13 +237,33 @@ contract RewardsDistributorTest is Test {
...
+    }
+
     function test_onPositionUpdated() public {
         SNX.mint(address(rewardsDistributor), 1000e18);
         uint256 actorSharesD18 = 123;
         rewardsDistributor.onPositionUpdated(accountId, poolId, collateralType, actorSharesD18);
     }
 
-    function test_supportsInterface() public {
+    function test_supportsInterface() public view {
```
noisekit, line 266:
```suggestion
    function test_supportsInterface() public {
```

auxiliary/RewardsDistributor/test/RewardsDistributorTest.sol
```diff
@@ -267,13 +237,33 @@ contract RewardsDistributorTest is Test {
...
+				rewardsManager.setCancelledAmount(onceCancelledAmount);
+        rewardsDistributor.distributeRewards(poolId, address(0), amount, start, duration);
+        vm.stopPrank();
+
+        assertEq(rewardsManager.poolId(), poolId);
+        assertEq(rewardsManager.collateralType(), address(0));
+        assertEq(rewardsManager.amount(), amount);
+        assertEq(rewardsManager.start(), start);
+        assertEq(rewardsManager.duration(), duration);
+				assertEq(rewardsDistributor.rewardedAmount(), 2 * amount - onceCancelledAmount);
```
noisekit, line 257: Don't math in tests please
```suggestion
				assertEq(rewardsDistributor.rewardedAmount(), 150e18);
```

protocol/synthetix/contracts/storage/RewardDistribution.sol
```diff
@@ -109,7 +124,18 @@ library RewardDistribution {
             self.lastUpdate = 0;
 
             diffD18 += updateEntry(self, totalSharesD18);
-        }
+				} else {
```
noisekit, line 127: With adding another kind of condition about scheduling, would be good to be more explicit about all the conditions, rather than just `else`, with default case throwing error of sorts.

smth like
```
if (true1) {
  do_stuff1
  return
}

if (true2) {
  do_stuff2
  return
}

revert SomethingOff
```

protocol/synthetix/contracts/storage/RewardDistribution.sol
```diff
@@ -216,17 +242,29 @@ library RewardDistribution {
...
+							curUpdateDistributedD18 = self.scheduledValueD18;
+							self.start = self.nextStart;
+							self.duration = self.nextDuration;
+							self.scheduledValueD18 = self.nextScheduledValueD18;
+							self.nextStart = 0;
+							self.nextDuration = 0;
+							self.nextScheduledValueD18 = 0;
+
+							// start processing the new entry also
+							valuePerShareChangeD18 = updateEntry(self, totalSharesAmountD18);
```
noisekit, line 263: Recursive call? 
Can we extract functionality into internal `_updateEntry` and reuse it depending on current conditions?

auxiliary/RewardsDistributor/test/RewardsDistributorTest.sol
dbeal-eth: resolved in private chat

protocol/synthetix/contracts/storage/RewardDistribution.sol
dbeal-eth: the call itself already acts as an internal role that also happens to be used externally whenever `totalShares` of a distribution changes.

there is no benefit that I can see because:
* all the functionality of the existing `updateEntry` function is being used by this recursive call
* this funtion in and of itself is used both internally (through this recursive call, and call through `distribute` function in this file), as well as externally (when the distributions `totalShares` changes, it is supposed to be called by users of the library)

protocol/synthetix/contracts/interfaces/IRewardsManagerModule.sol
```diff
@@ -111,14 +111,16 @@ interface IRewardsManagerModule {
...
+     * the duration period being interrupted by the start of this new distribution
      */
     function distributeRewards(
         uint128 poolId,
         address collateralType,
         uint256 amount,
         uint64 start,
         uint32 duration
-    ) external;
+    ) external returns (uint256 cancelledAmount);
```
noisekit, line 123: just realised this is a breaking change. Can we have a separate view for cancelled amount instead of updating this function signature please?

dbeal-eth: add a return value actually does not change the signature. try it out!

dbeal-eth: previously no functions used any return value from this function so adding one here wont break previous impls

dbeal-eth:
closing because this branch was merged into #2165

PR #2168, opened 2024-06-11
Audit / other fixes

sunnyvempati:
- [x] Add fees to debt instead of selling synths.
- [x] track global debt to subtract from reported debt.
- [x] Clear debt on account margin liquidation.
- [x] Fix other non-high severity audit issues: https://www.notion.so/iosiro/SIP-378-Perps-V3-Multi-collateral-9f20659dd11147ee8ea1edc840bfecc9

### DIFF
--- a/markets/perps-market/contracts/interfaces/IAsyncOrderSettlementPythModule.sol
+++ b/markets/perps-market/contracts/interfaces/IAsyncOrderSettlementPythModule.sol
@@ -62,6 +62,8 @@ interface IAsyncOrderSettlementPythModule {
         uint256 synthDeductionIterator;
         uint128[] deductedSynthIds;
         uint256[] deductedAmount;
+        int256 chargedAmount;
+        uint256 newAccountDebt;
     }
 
     /**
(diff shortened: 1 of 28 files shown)


markets/perps-market/contracts/interfaces/IAsyncOrderSettlementPythModule.sol
```diff
@@ -52,8 +52,6 @@ interface IAsyncOrderSettlementPythModule {
         int256 pnl;
         uint256 chargedInterest;
         int256 accruedFunding;
-        uint256 pnlUint;
-        uint256 amountToDeduct;
```
sunnyvempati, line 56: @leomassazza just to verify here, these are not used anymore, should be safe to remove right since this struct is only instantiated during runtime?

leomassazza: no problem here, the issue is for storage structs, not runtime.

leomassazza (approved):
LGTM

PR #2172, opened 2024-06-13
SIP-360: Wormhole Cross Chain Integration

moss-eth:
Changes cross-chain functionality to using Wormhole instead of CCIP (as defined in [SIP-360](https://sips.synthetix.io/sips/sip-341/)](https://sips.synthetix.io/sips/sip-360/).

## Description
- Modifies the crosschain libraries and contracts to use Wormhole instead of CCIP
- Modifies test suite to work with the new contracts and delivery flow
- Imports relevant Wormhole interfaces (Relayer and Core)
- Edits Cannonfiles to initialize the election modules properly


protocol/governance/cannonfile.toml
moss-eth, line 1: for reviewers: suggestions on how we could condense the number of cannonfiles and repeat code would be helpful

mjlescano (approved):

PR #2173, opened 2024-06-14
Avoid division by zero when totalShares is zero

noisekit:


dbeal-eth (changes requested):
It seems very straightforward, but we should add a test to verify outcome of condition. It's probably relatively easy to test

noisekit:
> It seems very straightforward, but we should add a test to verify outcome of condition. It's probably relatively easy to test

Yeah it is not really straightforward. We do not have any tests around this stuff at all at the moment.

PR #2183, opened 2024-06-25
[Gov v3] add grace period for preparing ballot

mjlescano:
Minimum delegation time always needs to be greater than 0 on all chains that use collateral delegation to establish voting power. This is to prevent a malicious user from using a flashloan to artificially inflate their amount delegated and then calling takeVotePowerSnapshot().


mjlescano:
This PR should not be merged as it really does not prevent flashloan usage. The `minDelegationTime > 0` should be setted on the Pool that's used for vote counting, not here:
> this PR and it does not seem to protect against the flashloan attack described. This is because a flashloan can still be used when calling takeSnapshot to artificially increase the vote power when the snapshot is taken. Preventing takeSnapshot and prepareBallotWithSnapshot from being executed in the same block does not prevent the snapshot from being inflated. If it was possible to detect that a flashloan was being used that would stop the attack, so for example using tx.origin == msg.sender (i.e. being called by an EOA) would be one option. However, this is not a strong control as account abstraction might bypass this check at some point in the future.  It is quite difficult to mitigate at the Election Module level, because the potential to use a flashloan to attack it largely depends on the source of the vote power. In the case of v3 governance, where delegated collateral is used as the sole source of vote power, a strong control is to ensure that minDelegationTime > 0 for any pool that has vote power. This works because the collateral cannot be delegated and undelegated atomically, so flashloans do no work. We recommend adding strong warnings in the comments of the takeSnapshot and setSnapshotContract  stating that atomic inflation attacks are possible with snapshot functionality, so developers should implement a mitigation if additional snapshot contracts are used.

PR #2187, opened 2024-07-01
Ensure we handle realisation when pnl matches debt

0xjocke:


markets/bfp-market/contracts/storage/Margin.sol
```diff
@@ -101,13 +101,18 @@ library Margin {
...
+                accountMargin.debtUsd = 0;
+                market.updateDebtAndCollateral(-previousDebt.toInt(), 0, addresses.sUsd);
+            }
             return;
         }
 
         uint256 availableUsdCollateral = accountMargin.collaterals[addresses.sUsd];
-        uint128 previousDebt = accountMargin.debtUsd;
 
         if (amountDeltaUsd >= 0) {
```
davidvuong, line 117: Minor: You just need to check for `> 0` here because the `== 0` is checked above.

markets/bfp-market/contracts/storage/Margin.sol
```diff
@@ -101,13 +101,18 @@ library Margin {
         int256 amountDeltaUsd,
```
davidvuong, line 101: It's been a super long time since I've touched this and probably my fault for not using descriptive names but does it make sense to call this `profitOrLossDeltaUsd`? I was scratching my head wondering why zero or more `amountDeltaUsd` meant we can clear the debt but positive is profit, zero is profit that perfectly balances the debt and negative means more debt where the calculation of profit/loss is performed by the caller.

If you don't want to rename it, just add a comment would be useful to

davidvuong (approved):
LGTM. Just left 2 comments

PR #2191, opened 2024-07-04
oraclemannager: add arguments to staleness node error

dbeal-eth:
should make it much easier to understand a stale price error when one comes up


noisekit (approved):
❤️

noahlitvin (approved):
I like it

PR #2195, opened 2024-07-11
Fix for underflow error when offchain price timestamp is bigger than current block timestamp

noisekit:
On arbitrum sepolia block is minted with a reasonable delay so after getting the fresh price update txn - the timestamp of the update will be "in the future" for the current block and we get underflow error

Example: https://www.tdly.co/shared/simulation/8e997e25-04d6-47a9-b566-56a779efed4c

### DIFF
--- a/protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
+++ b/protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
@@ -31,6 +31,10 @@ library StalenessCircuitBreakerNode {
             runtimeValues
         );
 
+        if (block.timestamp <= priceNodeOutput.timestamp) {
+            return priceNodeOutput;
+        }
+
         if (block.timestamp - priceNodeOutput.timestamp <= stalenessTolerance) {
             return priceNodeOutput;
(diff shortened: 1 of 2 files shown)


protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
```diff
@@ -31,6 +31,10 @@ library StalenessCircuitBreakerNode {
             runtimeValues
         );
 
+        if (block.timestamp <= priceNodeOutput.timestamp) {
```
dbeal-eth, line 34: It looks like this if statement and the if statement after words could be merged

dbeal-eth (changes requested):

protocol/oracle-manager/contracts/nodes/StalenessCircuitBreakerNode.sol
noisekit: yeah simplified with
```js
block.timestamp - stalenessTolerance <= priceNodeOutput.timestamp
```
As `block.timestamp - stalenessTolerance` cannot get negative.

dbeal-eth (approved):

PR #2198, opened 2024-07-15
Various Perps v3 audit fixes

sunnyvempati:
Guardian Audit Fixes:

- [x] Fix utilization calculation to exclude snxUSD [Issue](https://www.notion.so/guardianaudits/C-02-cb3aa2725b2f41d5bc2e22ab22b37459?pvs=4)
- [x] Fix `calculateStartingPnl` using new position size instead of size delta [Issue](https://www.notion.so/guardianaudits/H-03-60ef6bbde43d46898d8a06dbfe4b221a?pvs=4)
- [x] Check market capacity during order validation to ensure market stays solvent. [Issue](https://www.notion.so/guardianaudits/H-02-9766797413ab46389b0f10ef50e1f216?pvs=4)

### DIFF
--- a/markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
+++ b/markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
@@ -97,9 +97,6 @@ library GlobalPerpsMarket {
 
             accumulatedMinimumCredit += PerpsMarket.requiredCredit(marketId, priceTolerance);
         }
-
-        // add the sUSD collateral value to the minimum credit since it's used as escrow
-        accumulatedMinimumCredit += self.collateralAmounts[SNX_USD_MARKET_ID];
     }
 
     function totalCollateralValue(Data storage self) internal view returns (uint256 total) {
--- a/markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
+++ b/markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
@@ -50,6 +50,11 @@ library GlobalPerpsMarket {
         uint256 withdrawAmount
     );
 
+    /**
+     * @notice Thrown when an order puts the market above its credit capacity by checking it's utilization
+     */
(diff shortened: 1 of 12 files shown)


leomassazza (approved):

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -70,13 +75,24 @@ library GlobalPerpsMarket {
         }
     }
 
+    /**
+     * @notice Check whether, given the new locked credit delta, if the market is still solvent given its credit capacity from the core system.
+     */
+    function validateMarketCapacity(Data storage self, int256 lockedCreditDelta) internal view {
+        int256 delegatedCollateralValue = getDelegatedCollateralValue(self);
+        int256 lockedCredit = minimumCredit(self, PerpsPrice.Tolerance.DEFAULT).toInt() +
```
GuardianAudits, line 83: I believe we'll just need to add the sUSD collateral amount to the `lockedCredit` here as this has been moved to the `PerpsMarketFactoryModule.minimumCredit function`.

GuardianAudits: Ah I believe the getDelegatedCollateralValue should remove this amount correctly, verifying...

markets/perps-market/contracts/storage/PerpsMarket.sol
```diff
@@ -426,6 +426,17 @@ library PerpsMarket {
         return positionPnl + fundingPnl - self.debtCorrectionAccumulator;
     }
 
+    function requiredCreditForSize(
```
GuardianAudits, line 429: Makes sense, maybe just a note that in the future this should only be used for delta's in position size -- as short positions should still have a positive required credit.

Or perhaps this is more of a `requiredCreditForDelta` function.

GuardianAudits (approved):

PR #2200, opened 2024-07-17
Fill premium should only be applied to delta

0xjocke:


markets/bfp-market/contracts/storage/Position.sol
```diff
@@ -172,19 +172,20 @@ library Position {
      */
     function validateNextPositionEnoughMargin(
         Position.Data memory newPosition,
+        int128 oldPositionSize,
         uint256 oraclePrice,
         uint256 im,
-        uint256 mm,
         uint256 nextMarginUsd
     ) internal pure {
+        int128 sizeDelta = newPosition.size - oldPositionSize;
```
davidvuong, line 180:
```suggestion
        // Compute the net size difference as the fillPremium is only applied on the change in size and not total.
        int128 sizeDelta = newPosition.size - oldPositionSize;
```

davidvuong:
LGTM although perhaps an explicit test for this would be good as well.

davidvuong (approved):

0xjocke:
> LGTM although perhaps an explicit test for this would be good as well.

The test I made changes to is testing the fill premium, so I'll merge

PR #2202, opened 2024-07-19
PayDebt fixes

sunnyvempati:
- [x] Fix debt wipe when attempting to pay with more amount than debt.
- [x] update global debt when debt is paid.
- [x] update interest rate on debt payment.
- [x] add `PERPS_SYSTEM` feature flag

Some notable considerations:
- When `payDebt()` is called with an amount larger than the debt, only upto the debt amount is transferred from caller, so we have eliminated adding excess to USD margin since that would require global collateral validations and other considerations.
- All deposit/withdraw margin interactions are done via `modifyCollateral` for a trader.
- Only way to repay their debt is to call `payDebt`.
- Anyone can pay someone else's debt, no reason to have only trader pay their own debt

### DIFF
--- a/markets/perps-market/contracts/modules/PerpsAccountModule.sol
+++ b/markets/perps-market/contracts/modules/PerpsAccountModule.sol
@@ -9,11 +9,13 @@ import {SetUtil} from "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
 import {ITokenModule} from "@synthetixio/core-modules/contracts/interfaces/ITokenModule.sol";
 import {PerpsMarketFactory} from "../storage/PerpsMarketFactory.sol";
 import {IPerpsAccountModule} from "../interfaces/IPerpsAccountModule.sol";
+import {IGlobalPerpsMarketModule} from "../interfaces/IGlobalPerpsMarketModule.sol";
 import {PerpsAccount, SNX_USD_MARKET_ID} from "../storage/PerpsAccount.sol";
 import {Position} from "../storage/Position.sol";
 import {AsyncOrder} from "../storage/AsyncOrder.sol";
 import {PerpsMarket} from "../storage/PerpsMarket.sol";
 import {GlobalPerpsMarket} from "../storage/GlobalPerpsMarket.sol";
+import {InterestRate} from "../storage/InterestRate.sol";
 import {PerpsPrice} from "../storage/PerpsPrice.sol";
 import {MathUtil} from "../utils/MathUtil.sol";
 import {Flags} from "../utils/Flags.sol";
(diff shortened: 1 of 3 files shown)


markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -249,24 +249,32 @@ library PerpsAccount {
...
 
+        /*
+            1. if the debt is less than the amount, set debt to 0 and only deposit debt amount
+            2. if the debt is more than the amount, subtract the amount from the debt
+            3. excess amount is ignored
+        */
+
         PerpsMarketFactory.Data storage perpsMarketFactory = PerpsMarketFactory.load();
+
+        if ((self.debt.toInt() - amount.toInt()) > 0) {
```
leomassazza, line 265: Why don't just `if (self.debt > amount) { }` ? That way reading is direct with the comment above.

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -249,24 +249,32 @@ library PerpsAccount {
...
+            updateAccountDebt(self, -amount.toInt());
```
leomassazza, line 266: np: order `debtPaid` and `updateAccountDebt` in the same order on both branches. 
In fact, if you do the assignemt before, you can do the updateAccountDebt outside of the if with just `updateAccountDebt(self, -debtPaid.toInt());`

markets/perps-market/contracts/storage/PerpsAccount.sol
sunnyvempati: lol yea much better

markets/perps-market/contracts/storage/PerpsAccount.sol
sunnyvempati: yep, thx.  fixed.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -94,14 +96,21 @@ contract PerpsAccountModule is IPerpsAccountModule {
 
     // 1. call depositMarketUsd and deposit amount directly to core system
     // 2. look up account and reduce debt by amount
-    // 3b. quoteUnwrap() -> inchQuote -> returnAmount
     function payDebt(uint128 accountId, uint256 amount) external override {
+        FeatureFlag.ensureAccessToFeature(Flags.PERPS_SYSTEM);
```
GuardianAudits, line 100: Worth noting/considering that the BFP market validates that the account has permission to modify the account's collateral: https://github.com/Synthetixio/synthetix-v3/blob/8aff01938913983b97faa5ce082c15b86db32e0d/markets/bfp-market/contracts/modules/MarginModule.sol#L498

GuardianAudits: Additionally, BFP will revert on 0 amounts, don't see any bad outcomes here, but avoids unexpected behaviors/event emissions.

GuardianAudits: I suppose even if you reverted on 0 amounts here, the debtPaid could still be capped to 0 too. So if added, perhaps the validation should be based on the `debtPaid` result.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -94,14 +96,21 @@ contract PerpsAccountModule is IPerpsAccountModule {
...
         Account.exists(accountId);
+
```
GuardianAudits, line 102: Another recommended validation we had for BFP was to ensure no orders were pending for the account. This was to prevent potential gaming with respect to delaying orders and then allowing them to go through via manipulating debt in a single transaction.

Finding for BFP: https://www.notion.so/guardianaudits/M-02-0b54ea9c1dfe439b8bd9ecc3c39ac8fc?pvs=4

The BFP finding leveraged the merge feature, which doesn't exist here, but it still may be a good validation to consider to cut off any other potential manipulations.

https://github.com/Synthetixio/synthetix-v3/pull/2216/files#diff-f2bde65b1d25fef9e34a2acc04220ffc7cdcb2d3c8e28e1dc9ac92396f45cc8eR507

leomassazza (approved):

PR #2204, opened 2024-07-20
legacymarket: allow migration of liquidatable accounts

dbeal-eth:
we remove the code that allows for the legacymarket to "liquidate" accounts, and instead allow even liquidatable accounts to be migrated so that the v3 liquidation system can be used instead.

this change works by selectively removing a restriction on the v3 core system which prevented the assignment of debt which would cause an account to become liquidatable, a tactic that would allow a market to mint effectively unlimited debt.


markets/legacy-market/contracts/LegacyMarket.sol
```diff
@@ -222,25 +224,7 @@ contract LegacyMarket is ILegacyMarket, Ownable, UUPSImplementation, IMarket, IE
         uint256 cratio = (collateralMigrated * v3System.getCollateralPrice(address(oldSynthetix))) /
             debtValueMigrated;
 
-        // if the account needs to be liquidated, liquidate it here by unlocking the debt to all accounts and moving to a
         if (
-            cratio < v3System.getCollateralConfiguration(address(oldSynthetix)).liquidationRatioD18
```
kaleb-keny, line 227: If we delete this line how does the liquidation workflow work? I mean how can one force a v2x liquidation, without hitting reverts due to not owning the account when calling migrate?

kaleb-keny (approved):

markets/legacy-market/cannonfile.toml
```diff
@@ -103,6 +96,18 @@ extra.marketId.event = "MarketRegistered"
 extra.marketId.arg = 1
 depends = ["invoke.setFeatureFlagMarket", "invoke.setSystemAddresses"]
 
+# tell the v3 system its ok for this market to assume unlimited debt (to allow underwarter stakers to be converted and then liquidated)
+[invoke.setConfigAssociateUnlimitedDebt]
+target = ["v3.CoreProxy"]
+fromCall.func = "owner"
+func = "setConfig"
+args = [
+    "<%= keccak256(defaultAbiCoder.encode(['bytes32', 'uint256'], [formatBytes32String('associateDebtRatioExclusion'), settings.marketId])) %>",
```
barrasso, line 105: should be `settings.market_id`

utils/core-contracts/contracts/utils/RevertUtil.sol
```diff
@@ -0,0 +1,15 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.11 <0.9.0;
+
+/*
+		Utilities related to reverts and error handling
+*/
+
+library RevertUtil {
```
barrasso, line 8: Did this get added by mistake? Harmless, but seems out of scope/unused

barrasso (changes requested):

utils/core-contracts/contracts/utils/RevertUtil.sol
dbeal-eth: ah whoops. I think I checked it in because I thought I forgot to include the file. probably goes with a different PR :/

barrasso (approved):

PR #2209, opened 2024-07-24
L-13, L-12, L-11, L-09

davidvuong:


markets/bfp-market/contracts/storage/Position.sol
```diff
@@ -605,6 +605,11 @@ library Position {
         Margin.MarginValues memory marginValues,
         AddressRegistry.Data memory addresses
     ) internal view returns (uint256) {
+        // A zero sized position means there is no position.
+        if (size == 0) {
+            return 2 ** 256 - 1;
```
0xjocke, line 610:
```suggestion
            return type(uint256).max;
```

Maybe this is cleaner?

PR #2211, opened 2024-07-27
C-02: fix collateral discount logic

sunnyvempati:
- Fix min/max collateral discount logic [Issue](https://www.notion.so/guardianaudits/C-02-af4dc12dc2164d5e8ec5f10f2372c286?pvs=4)

### DIFF
--- a/markets/perps-market/contracts/storage/PerpsAccount.sol
+++ b/markets/perps-market/contracts/storage/PerpsAccount.sol
@@ -2,7 +2,6 @@
 pragma solidity >=0.8.11 <0.9.0;
 
 import {ERC2771Context} from "@synthetixio/core-contracts/contracts/utils/ERC2771Context.sol";
-import {Price} from "@synthetixio/spot-market/contracts/storage/Price.sol";
 import {DecimalMath} from "@synthetixio/core-contracts/contracts/utils/DecimalMath.sol";
 import {SafeCastI256, SafeCastU256, SafeCastU128} from "@synthetixio/core-contracts/contracts/utils/SafeCast.sol";
 import {SetUtil} from "@synthetixio/core-contracts/contracts/utils/SetUtil.sol";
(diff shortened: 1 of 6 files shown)


markets/perps-market/test/integration/Account/Margins.multiCollateral.test.ts
```diff
@@ -58,7 +58,7 @@ const ethDiscountConfig = {
   upperLimitDiscount: bn(0.05),
   lowerLimitDiscount: bn(0.01),
   discountScalar: bn(2),
-  skewScale: bn(10_000),
+  skewScale: bn(50_000),
```
sunnyvempati, line 61: this forces the skew impact to be really small which is where the bug lies otherwise the tests passed cuz the discount would always be higher than the upper limit.

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -164,7 +163,23 @@ library PerpsAccount {
         Data storage self,
         PerpsPrice.Tolerance stalenessTolerance
     ) internal view returns (bool isEligible, int256 availableMargin) {
-        availableMargin = getAvailableMargin(self, stalenessTolerance);
+        // calculate keeper costs
+        KeeperCosts.Data storage keeperCosts = KeeperCosts.load();
+        uint256 totalLiquidationCost = keeperCosts.getFlagKeeperCosts(self.id) +
```
GuardianAudits, line 168: Since this check is relevant for margin only liquidations, should these costs correspond to margin only liquidations?

E.g. maybe this function can use:

```solidity
uint256 keeperReward = Margin.getMarginLiquidationOnlyReward(
            marginValues.collateralUsd,
            PerpMarketConfiguration.load(marketId),
            globalConfig,
            addresses
        );
 ```
 
 Instead

sunnyvempati: yeah this function (`isEligibleForMarginLiquidation`) is only used for margin liquidations.  yes can create a separate function but i'd imagine non blocking for now

leomassazza (approved):

PR #2213, opened 2024-07-30
H8 - Skip distribution if collateral is zero

leomassazza:


leomassazza:
CPed directly on poc-multi-coll

PR #2215, opened 2024-07-31
Oracle manager process many

dbeal-eth:
adds a capability for oracle manager to process many oracles at the same time, and if one oracle fails, it saves the error and returns them all as a batch that can be conveniently read by the client all at once

it is expected that this change will significantly improve the UX of the frontend and other places that are dependent upon the use of ERC7412 by reducing the number of callbacks required to simulate the txn to just 1.

notes:
* there are many places where this new function can be used, but for right now I am just focusing in 1 place
* the way that the errors are aggregated together is through a new error called `Errors(bytes[])` which basically is just an array of errors. it can be nested too, which allows for sub-nodes to send errors to super nodes without being a disaster in solidity code


markets/perps-market/contracts/modules/PerpsMarketFactoryModule.sol
```diff
@@ -120,12 +120,15 @@ contract PerpsMarketFactoryModule is IPerpsMarketFactoryModule {
             uint256 collateralValue = globalMarket.totalCollateralValue();
             int256 totalMarketDebt;
 
-            SetUtil.UintSet storage activeMarkets = globalMarket.activeMarkets;
-            uint256 activeMarketsLength = activeMarkets.length();
-            for (uint256 i = 1; i <= activeMarketsLength; i++) {
-                uint128 marketId = activeMarkets.valueAt(i).to128();
-                totalMarketDebt += PerpsMarket.load(marketId).marketDebt(
-                    PerpsPrice.getCurrentPrice(marketId, PerpsPrice.Tolerance.DEFAULT)
+            uint256[] memory activeMarkets = globalMarket.activeMarkets.values();
```
dbeal-eth, line 123: this is how the usage of the new oracle manager generally looks in the code: instaed of having a loop over every price which repetedly calls oracle manager, we cal oracle manager once and it processes all the prices at once.

protocol/oracle-manager/contracts/interfaces/INodeModule.sol
```diff
@@ -16,6 +16,16 @@ interface INodeModule {
      */
     error InvalidNodeDefinition(NodeDefinition.Data nodeType);
 
+    /**
+     * @notice An array of revert reasons when an array of nodes is processed, but some of the nodes failed
+     */
+    error Errors(bytes[] revertReasons);
```
noisekit, line 22:
```suggestion
    error OracleNodeErrors(bytes[] revertReasons);
```

noisekit (commented):
One small recommendation for now. But as this PR is pretty large it will take more time to go over it in full.

protocol/oracle-manager/contracts/interfaces/INodeModule.sol
dbeal-eth: Its unknown for sure if the errors are going to be Oracle node related or some other error. so it makes more sense to just call them errors and let the client decide what type of errors they are and if there's any action that can or should be taken

dbeal-eth: Oh, I see what you're saying with this suggestion. My concern is that we may find this feature to be useful and use it elsewhere. I think it would be good to treat this as something more standardized and know how to recursively parse groups of errors like this, not just for erc7412

protocol/oracle-manager/test/integration/nodes/PriceDeviationCircuitBreakerNode.test.ts
```diff
@@ -44,11 +44,11 @@ describe('PriceDeviationCircuitBreakerNode', function () {
     });
 
     it('expect process to revert for prices = [1, 0.5]', async () => {
-      await assertRevert(NodeModule.process(node1), 'DeviationToleranceExceeded', NodeModule);
+      await assertRevert(NodeModule.process(node1), '0xdcac0919', NodeModule);
```
noisekit, line 47: can we not use hex errors in tests please

protocol/oracle-manager/test/integration/modules/NodeModule.test.ts
```diff
@@ -60,4 +52,66 @@ describe('NodeModule', function () {
...
+      await assertRevert(
+        NodeModule.processManyWithRuntime([nodeId2(), nodeId1(), reducer2Id], [], []),
+        'Errors("0xac47be2100000000000000000000000000000000000000000000000000000000000004d2,0x0b42fd17000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001040b42fd1700000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000024ac47be2100000000000000000000000000000000000000000000000000000000000004d2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")',
```
noisekit, line 112: can you please "encode" readable values in here to match against instead of hardcoding full response as is

protocol/oracle-manager/test/integration/modules/NodeModule.test.ts
```diff
@@ -60,4 +52,66 @@ describe('NodeModule', function () {
...
+      assertBn.equal(priceData.price, ethers.utils.parseEther('1'));
+
+      priceData = await NodeModule.process(nodeId2());
+      assertBn.equal(priceData.price, ethers.utils.parseEther('0.9'));
+    });
+
+    it('passes through errors', async () => {
+      await assertRevert(
+        NodeModule.process(failingNodeId()),
+        '0xac47be2100000000000000000000000000000000000000000000000000000000000004d2'
```
noisekit, line 68: need a readable value here too

noisekit (approved):
I'd really love to see readable asserts in tests before the merge, but not blocker

protocol/oracle-manager/contracts/interfaces/INodeModule.sol
noisekit: I can see that the same format used in other places. So maybe it is fine after all. Is it worth extracting as separate reusable interface maybe instead of duplicating?

utils/core-utils/src/utils/solidity/compiler.ts
```diff
@@ -88,11 +88,7 @@ async function _getCompiler(version: string): Promise<any> {
   const isDownloaded = await downloader.isCompilerDownloaded(version);
 
   if (!isDownloaded) {
-    await downloader.downloadCompiler(
-      version,
-      async () => {},
-      async () => {}
-    );
+    await downloader.downloadCompiler(version);
```
barrasso, line 91: Looks like this is causing CI to fail: https://app.circleci.com/pipelines/github/Synthetixio/synthetix-v3/3296/workflows/ce27a610-0b90-48ad-b23f-47e44855398e/jobs/56098

barrasso (approved):
Please fix CI failures, otherwise LGTM.

utils/core-utils/src/utils/solidity/compiler.ts
noisekit: Oh yeah I remember fixing that somewhere. hold on

noisekit: haha indeed
image

dbeal-eth:
sending this into audit, will look into improving the error checking next week cc @noisekit 

thanks everyone for the reviews

dbeal-eth:
audits passed.

PR #2216, opened 2024-08-01
bfp-market m-02

davidvuong:


markets/bfp-market/test/integration/modules/MarginModule.debt.test.ts
```diff
@@ -151,6 +153,23 @@ describe('MarginModule Debt', async () => {
       );
     });
 
+    it('should revert when an order is present', async () => {
```
0xMithrandir, line 156: I'd say should revert **paying debt** when an order is present

0xMithrandir (approved):
Looks good to me, only have 1 comment about naming the test

markets/bfp-market/test/integration/modules/MarginModule.debt.test.ts
davidvuong: It's part of the `payDebt` describe block so no need? Same name structure as all other reverts in this test suite.

PR #2221, opened 2024-08-10
Fix H-01 (partial of previous H-04)

leomassazza:
check if a pending order exists at payDebt

### DIFF
--- a/markets/perps-market/contracts/modules/PerpsAccountModule.sol
+++ b/markets/perps-market/contracts/modules/PerpsAccountModule.sol
@@ -100,6 +100,8 @@ contract PerpsAccountModule is IPerpsAccountModule {
         FeatureFlag.ensureAccessToFeature(Flags.PERPS_SYSTEM);
         Account.exists(accountId);
 
+        AsyncOrder.checkPendingOrder(accountId);
+
         PerpsAccount.Data storage account = PerpsAccount.load(accountId);
         uint256 debtPaid = account.payDebt(amount);
 


markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -100,6 +100,8 @@ contract PerpsAccountModule is IPerpsAccountModule {
         FeatureFlag.ensureAccessToFeature(Flags.PERPS_SYSTEM);
         Account.exists(accountId);
 
+        AsyncOrder.checkPendingOrder(accountId);
```
wasp-in-bush, line 103: Given this validation check is critical to the market, I recommend adding a single line comment that briefly informs the reader of its importance.

markets/perps-market/contracts/modules/PerpsAccountModule.sol
```diff
@@ -100,6 +100,8 @@ contract PerpsAccountModule is IPerpsAccountModule {
         FeatureFlag.ensureAccessToFeature(Flags.PERPS_SYSTEM);
```
wasp-in-bush, line 100: Although out of scope for this specific commit, I noticed `payDebt()` and `debt()` override existing functions defined within the interface `IPerpsAccountModule`.

Please use the `@inheritdoc` tag to inherit the interface documentation as is done elsewhere within the contract.

ex:
```
/**
  * @inheritdoc IPerpsAccountModule
  */
```

markets/perps-market/contracts/modules/PerpsAccountModule.sol
wasp-in-bush, line 103: Calling `checkPendingOrder()` will _throw_ if an order exists that has not yet expired. However, if the order has expired, the debt can still be _safely paid_ despite the existing order.

This behavior is perfectly fine, but integrators referring to the interface might not expect an error to be thrown when calling this function simply because an order exists.

I recommend that functions that throw errors via a nested call to another function should include that information in the interface.

For example: `f(x) -> g(x)` where `g(x)` may revert with `e`; `e` should be mentioned within the interface containing `f(x)`

> I encountered this issue myself while working from the integrator’s perspective 😅

wasp-in-bush, line 103: ❓ Question:

I see in the `bfp-market` [implementation](https://github.com/Synthetixio/synthetix-v3/blob/75975f9f8ef04a7d44186abffd8a096ddca0d0b8/markets/bfp-market/contracts/modules/MarginModule.sol#L494) of `payDebt()` the strategy for checking order existence solely relies on size and does not concern itself with the order's expiration status. Why is this?

⛳️ Unrelated to this commit:

@davidvuong @0xjocke I noticed when reading the `bfp-market` version of `payDebt()`:

```
PerpMarket.Data storage market = PerpMarket.exists(marketId);
Margin.Data storage accountMargin = Margin.load(accountId, marketId);
Order.Data storage order = market.orders[accountId];
if (order.sizeDelta != 0) {
    revert ErrorUtil.OrderFound();
}
```

loading account margin (`Margin.load(accountId, marketId);`) can be safely moved _after_ the order size conditional logic. This will reduce gas lost if the call throws with `OrderFound`

wasp-in-bush (approved):
I am approving now because the only things that really need to be "changed" are the interface and adding the `@inheritdoc` tag.

Otherwise, the addition looks great 💪

PR #2234, opened 2024-08-22
Allow Pool-level Rewards distributors in Perps

noisekit:


dbeal-eth (approved):
all we need to do is add a test "can add RD at zero address" or whatever

markets/perps-market/contracts/storage/LiquidationAssetManager.sol
```diff
@@ -53,9 +53,6 @@ library LiquidationAssetManager {
             revert ParameterError.InvalidParameter("collateralTypes", "must not be empty");
         }
         for (uint256 i = 0; i < collateralTypesLength; ) {
-            if (self.poolDelegatedCollateralTypes[i] == address(0)) {
-                revert AddressError.ZeroAddress();
```
noisekit, line 57: later in the code there is an iteration over all the collaterals configured and then their prices are used. So we end up having error that Zero-address oracle node is used.

noisekit: so to use pool level distributors with perps we would need quite a bit more refactoring

kaleb-keny (changes requested):
Blocked until issues resolved

PR #2235, opened 2024-08-23
ability to change token names from within synthetix core proxy

dbeal-eth:
for example we want to be able to change the symbol of the synthetix usd token

### DIFF
--- a/protocol/synthetix/cannonfile.toml
+++ b/protocol/synthetix/cannonfile.toml
@@ -152,7 +152,7 @@ target = ["CoreProxy"]
 fromCall.func = "owner"
 func = "initOrUpgradeNft"
 args = [
-    "0x6163636f756e744e667400000000000000000000000000000000000000000000",
+    "<%= formatBytes32String('accountNft') %>",
     "<%= settings.account_token_name %>",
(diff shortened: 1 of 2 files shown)


protocol/synthetix/cannonfile.toml
```diff
@@ -158,8 +170,8 @@ fromCall.func = "owner"
 func = "initOrUpgradeToken"
 args = [
     "0x555344546f6b656e000000000000000000000000000000000000000000000000",
-    "Synthetic USD Token v3",
-    "snxUSD",
+    "<%= settings.usd_token_name %>",
```
noisekit, line 173: image
we can already do it I guess.
Also can you replace `0x555344546f6b656e000000000000000000000000000000000000000000000000` (and account token too) with this please

```
[setting.usd_token_name]
defaultValue = "<%= formatBytes32String('USDToken') %>"
```

noisekit (approved):

protocol/synthetix/cannonfile.toml
dbeal-eth: yea true, but we should really be doing it through the official function. so I guess we can call this cleanup to bring upstream as we go

will do the replacements

PR #2238, opened 2024-08-27
M-01, M-02, L-01

davidvuong:


markets/bfp-market/contracts/interfaces/IPerpMarketFactoryModule.sol
```diff
@@ -94,9 +94,15 @@ interface IPerpMarketFactoryModule is IMarket, IBasePerpMarket {
 
     // --- Views --- //
 
-    /**
-     * @notice Returns a digest of an existing market given the `marketId`.
-     */
+    /// TODO: Why is this necessary?
```
davidvuong, line 97: @0xMithrandir External function on the PerpMarketFactoryModule but I don't see any usage of it anywhere. What was your original intent for exposing this and if not needed, I'd rather we remove this.

0xMithrandir: I'm pretty sure I've added it because of tests but now that I checked it's not used anywhere. 
I've removed it.

0xMithrandir (approved):

PR #2239, opened 2024-08-28
🔴 H-01: Required Credit Check Can DoS Decrease Orders

wasp-in-bush:
## **Description**

An Order can not be settled if the delegated liquidity by the pool to this market is smaller than the required liquidity to back the open interest + this order sufficiently.

Because of price changes, it could be that the current minimum credit is already above the delegated amount of the pool. In that case, it could be that reducing the OI and therefore improving the situation is prevented if this order does not sufficiently reduce the OI.

For example: minimum credit = 500 delegated amount = 400

A decrease order of 50 lockedCreditDelta will revert because: delegated amount < minimum credit - 50 But a decrease order of 150 lockedCreditDelta will be allowed because: delegated amount > minimum credit - 150

Both orders would reduce the open interest and therefore improve the situation and should be allowed.
## **Recommendation**

Do not perform this check if the order decreases the open interest.


markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
```diff
@@ -113,26 +131,22 @@ contract AsyncOrderSettlementPythModule is
...
+        // order fees are total fees minus settlement reward
+        uint orderFees = runtime.totalFees - runtime.settlementReward;
+        GlobalPerpsMarketConfiguration.Data s = GlobalPerpsMarketConfiguration.load();
+        s.collectFees(orderFees, asyncOrder.request.referrer, factory);
 
         // trader can now commit a new order
         asyncOrder.reset();
 
-        // Note: new event for this due to stack too deep adding it to OrderSettled event
+        /// @custom:question two events here seems bad. Can we not combine or find a better way?
```
fredsnax, line 147: combining the two events might be possible. we can take a look at this next week and sync up with the data and UI teams to see what sort of data they are using from this event in the data repo or clients and make sure they are aware of any changes so they can update their code as well

markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
```diff
@@ -61,39 +61,57 @@ contract AsyncOrderSettlementPythModule is
...
+        /// @dev runtime stores order settlement data; prevents stack too deep error
         SettleOrderRuntime memory runtime;
+
         runtime.accountId = asyncOrder.request.accountId;
         runtime.marketId = asyncOrder.request.marketId;
-        // check if account is flagged
+        runtime.sizeDelta = asyncOrder.request.sizeDelta;
+
+        // ensure account has not been flagged for liquidation
+        /// @custom:question is this check necessary given validateRequest() calls isEligibleForLiquidation()?
```
fredsnax, line 82: sent longer response in dm but checkLiquidation checks if already flagged. and isEligibleForLiquidation checks if account has enough margin. let's leave it alone for now and investigate next week in more depth

fredsnax (approved):

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -348,11 +350,18 @@ library AsyncOrder {
...
-            MathUtil.abs(runtime.newPositionSize).toInt() - MathUtil.abs(oldPosition.size).toInt(),
-            PerpsPrice.Tolerance.DEFAULT
-        );
-        GlobalPerpsMarket.load().validateMarketCapacity(lockedCreditDelta);
+        /// @custom:magnitude determines if more market credit is required
+        /// when a position's magnitude is increased, more credit is required
+        /// when a position's magnitude is decreased, less credit is required
+        uint newMagnitude = MathUtil.abs(runtime.newPositionSize);
+        uint oldMagnitude = MathUtil.abs(oldPosition.size);
+        if (newMagnitude > oldMagnitude) {
```
kaleb-keny, line 358: i think a bit more than that please, they need to be lower magnitude and have same side, to be able to skip the validation on min credit.
note the tricky part is that 0 is same side as both long and short

kaleb-keny: To reiterate, is someone flips from 10 eth long to 9 eth short. this is basically allowed since magnitude works, but i would prefer that the guy closes completely rather than continue trading... since the market does not have enough lp to back it.

markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
kaleb-keny: You want to avoid stack too deep issues, hence 2 events

markets/perps-market/contracts/storage/AsyncOrder.sol
kaleb-keny: `MathUtil.isSameSideReducing`

markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
wasp-in-bush: yes, but the above stack limitation can be gracefully circumvented without need of two events (bad code).

From [soliditylang abi specs for events](https://docs.soliditylang.org/en/latest/abi-spec.html#events):

> Events are an abstraction of the Ethereum logging/event-watching protocol. Log entries provide the contract’s address, a series of up to four topics and some arbitrary length binary data. 

Given: "arbitrary length binary data"
We can say: Any number of variables can be encoded and thus emitted without concern of the stack

wasp-in-bush: this shouldnt be changed here regardless but it is a code quality issue that should be logged... somewhere.....? @kaleb-keny 😅

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -348,11 +350,28 @@ library AsyncOrder {
...
+            /// @custom:magnitude determines if more market credit is required
+            /// when a position's magnitude is increased, more credit is required and risk increases
+            /// when a position's magnitude is decreased, less credit is required and risk decreases
+            uint newMagnitude = MathUtil.abs(runtime.newPositionSize);
+            uint oldMagnitude = MathUtil.abs(oldPosition.size);
+
+            /// @custom:side reflects if position is long or short; if side changes, further validation required
+            /// given new position size cannot be zero, it is inconsequential if old size is zero; 
+            /// magnitude will necessarily be larger
+            bool sameSide = runtime.newPositionSize > 0 == oldPosition.size > 0;
```
kaleb-keny, line 364: if `oldPosition.size` is 100 eth and `runtime.newPositionSize` is zero, this result in False, right? While it should True?

markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
kaleb-keny: Yeah i think sunny might have mentioned this to me once long time back, that splitting events was required for avoiding that stack-too-deep error.

markets/perps-market/contracts/storage/AsyncOrder.sol
wasp-in-bush: you are correct. however, given previous logic, `runtime.newPositionSize` will never be zero.

i.e.,

```
/// @dev if new position size is not 0, further credit validation required
if (runtime.newPositionSize != 0) {
    (...)
}
```

I also made not of this in the comment above the `sameSide` assignment: "given new position size cannot be zero"

markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -82,11 +82,17 @@ library GlobalPerpsMarket {
...
-        int256 lockedCredit = minimumCredit(self, PerpsPrice.Tolerance.DEFAULT).toInt() +
-            lockedCreditDelta;
+
+        /// @dev establish amount of credit needed to collateralize outstanding perp markets
+        int256 credit = minimumCredit(self, PerpsPrice.Tolerance.DEFAULT).toInt();
+
+        /// @dev calculate new accumulated credit following the addition of the new locked credit
+        credit += lockedCreditDelta;
 
         if (delegatedCollateralValue < lockedCredit) {
```
fredsnax, line 97: lockedCredit is now undefined

markets/perps-market/contracts/storage/AsyncOrder.sol
leomassazza: np. If I'm not wrong, there's a sameSide helper that can be used here to prevent duplicated code (though is not A LOT of code)

markets/perps-market/test/integration/Insolvent.test.ts
```diff
@@ -136,7 +136,7 @@ describe('Insolvent test', () => {
       });
     });
 
-    it('reverts when attempting to open position above market credit capacity', async () => {
+    it.only('reverts when attempting to open position above market credit capacity', async () => {
```
leomassazza, line 139: rogue `.only`

JaredBorders: fixed

wasp-in-bush: removed rogue `.only` 🚮

PR #2240, opened 2024-08-28
arbitrum multi collateral fee collector with fees going to TC

fredsnax:
…date the tests


auxiliary/OwnedFeeCollector/storage.dump.sol
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: UNLICENSED
```
fredsnax, line 1: note this file should be updated soon once the codebase is working with the open zeppelin package

fredsnax:
Everything seems to be working except the yarn storage:verify command won't generate a storage.dump.sol file. Maybe I have to delete the one that exists so it will write a new one. will try but not sure the issue

leomassazza (approved):

PR #2245, opened 2024-08-29
M-04: getKeeperRewardsAndCosts uses only flagging rewards

fredsnax:
…eral reward


markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -569,7 +569,7 @@ library AsyncOrder {
             runtime.oldRequiredMargin;
 
         (runtime.accumulatedLiquidationRewards, runtime.maxNumberOfWindows) = account
-            .getKeeperRewardsAndCosts(marketId);
+            .getKeeperRewardsAndCosts(marketId, PerpsPrice.Tolerance.DEFAULT);
```
kaleb-keny, line 572: I'm not sure if this is valid because `getPossibleLiquidationReward` does the flag+liquidate*windows, right?

fredsnax: getKeeperRewardsAndCosts is called in 2 different places. once in isEligibleForLiquidation and once for opening a new position. the method takes all open positions and calculates the rewards. when calling it in commitOrder it also adds the method below on the newPositionSize as well which is separate from the existing positions so it should be good to go with these changes

fredsnax: actually I do wonder if we should add the new position to the loop of flag costs and then compare that to the collateral reward instead of comparing a loop of the existing positions to the collateral reward and then adding the new position on top. what do you think?

markets/perps-market/contracts/storage/PerpsAccount.sol
```diff
@@ -489,20 +492,24 @@ library PerpsAccount {
             PerpsMarketConfiguration.Data storage marketConfig = PerpsMarketConfiguration.load(
                 marketId
             );
-
-            uint256 numberOfWindows = marketConfig.numberOfLiquidationWindows(
+            maxNumberOfWindows = marketConfig.numberOfLiquidationWindows(
```
leomassazza, line 495: By doing that you are assuming the latest market in the loop is the one that will require more liquidation windows (depends on market config). 
Previously, it picked the longest to process as # of windows

leomassazza (approved):

barrasso (approved):

PR #2257, opened 2024-09-03
Adjust dependabot

noisekit:
Looks like the idea to have catch-all after specifying groups did not play out. So we are back to individual deps.


dbeal-eth (approved):
That's a bummer. I thought the den dependabot groups were working well

noisekit:
> That's a bummer. I thought the den dependabot groups were working well

They do work well but I wanted to have a "rest" catch-all bucket. Looks like it is greedy and takes all

PR #2270, opened 2024-09-04
M-05: Flagged accounts should return true for canLiquidate

leomassazza:


markets/perps-market/contracts/modules/LiquidationModule.sol
```diff
@@ -164,6 +164,11 @@ contract LiquidationModule is ILiquidationModule, IMarketEvents {
      * @inheritdoc ILiquidationModule
      */
     function canLiquidate(uint128 accountId) external view override returns (bool isEligible) {
+        // If an account is already flagged can be liquidated, no matter other conditions
+        if (GlobalPerpsMarket.load().liquidatableAccounts.contains(accountId)) {
+            return true;
```
fredsnax, line 169: there may be some benefit to putting this logic in the last line of isEligibleForLiquidation. I still need to dive a bit deeper into some liquidations code today and can share more detailed thoughts on this later. Basically in validateRequest this would also add the check if an account has been flagged which seems like a better change 

`isEligible = (requiredMaintenanceMargin + liquidationReward).toInt() > availableMargin || GlobalPerpsMarket.load().liquidatableAccounts.contains(accountId);`

fredsnax (approved):

PR #2271, opened 2024-09-04
M07: Cap utilizationRate to 100%

leomassazza:


markets/perps-market/contracts/storage/GlobalPerpsMarket.sol
```diff
@@ -106,6 +106,11 @@ library GlobalPerpsMarket {
         delegatedCollateralValue = delegatedCollateralValueInt.toUint();
 
         rate = lockedCredit.divDecimal(delegatedCollateralValue).to128();
+
+        // Cap at 100% utilization
+        if (rate > DecimalMath.UNIT_UINT128) {
+            rate = DecimalMath.UNIT_UINT128;
+        }
```
fredsnax, line 113: upon confirmation by Guardian we will close this PR for 2 reasons:

1) it is an unnecessary change given it's usage with interest rate params. the system works fine without this change
2) interest rates are 0 so this change is not only unnecessary but it will not affect the system

GuardianAudits: Ultimately up to you guys, but it is possible for the market to go below the minimumCredit threshold and thus the result of `rate = lockedCredit.divDecimal(delegatedCollateralValue).to128();` be greater than 100%.

I’m not familiar with what the current interest configurations are, but if in the future they were enabled then this could become something noteworthy. It is also worth adding that the BFP market caps the rate to 100%.

kaleb-keny: @barrasso would you mind please assigning a ressource to re-instate this PR, @GuardianAudits is right and the cap need to be enforced in case  those parameters are updated

barrasso (approved):

PR #2276, opened 2024-09-06
H-07: Minimum delegate time not set

fredsnax:


fredsnax:
The core system has `setMarketMinDelegateTime()` function which has to be called by integrating markets to set their minimum waiting time for the LPs to withdraw their collateral after they have delegated it.

The PerpsMarket doesn't call this function which lets LPs delegate and undelegate collateral in the same block. This can lead to some unexpected behaviors such as interest rate manipulation or risk-free yield opportunity by sandwiching a settlement to gain from fees by depositing a large amount of collateral and the withdrawing it.

fredsnax:
Note that this PR passes CI but breaks all the local anvil tests due to time traveling issues

barrasso (approved):

PR #2286, opened 2024-09-11
OwnedFeeCollector for Arbitrum perps IFeeCollector - audit modifications 

fredsnax:
…0) in constructor and slight gas improvement in claimFees removing modifier


auxiliary/OwnedFeeCollector/test/integration/OwnedFeeCollector.test.ts
```diff
@@ -6,13 +6,14 @@ import { ethers } from 'ethers';
...
-  const { getContract, owner, user } = bootstrapOwnedFeeCollector();
+  const { getContract, owner, user, feeShareController } = bootstrapOwnedFeeCollector();
 
   let OwnedFeeCollector: ethers.Contract;
   let UsdToken: ethers.Contract;
 
   const usdAmount = bn(5000);
-  const ownerFeeShareRatio = bn(0.5);
+  const feeShareRatio = bn(0.5);
+  const zeroAddress = '0x0000000000000000000000000000000000000000';
```
leomassazza, line 16: you can use `ethers.constants.AddressZero`

fredsnax:
Changes in this PR: 

- according to the audit notes using this error pattern instead of the modifiers is slightly more gas efficient and we already use the error pattern throughout so I changed it although its not that important imo

- audit suggestion to add the ability to change the fee share. at first I thought it was better immutable but this is actually important since we are changing the fee share for the first month of arbitrum perps. Kaleb agreed we should be able to change it in the contract rather than deploy a new one.

- ability to change the fee share controller address with this new functionality of changing the fee share. think the pDAO controls the fee share and the TC claims the fees

- adds zero address check in the controller

fredsnax:
further changes after review by Leo: pdao is the owner. tc is the fee share recipient and can claim fees. pdao handles all other admin and can also claim fees to the fee share recipient or change the recipient address

auxiliary/OwnedFeeCollector/cannonfile.toml
```diff
@@ -3,9 +3,12 @@ version = "<%= package.version %>"
...
+defaultValue = "0xBE0eB53F46cd790Cd13851d5EFf43D12404d33E8"
+
+[setting.fee_share_recipient]
 defaultValue = "0x0000000000000000000000000000000000000000"
 
-[setting.owner_fee_share]
+[setting.fee_share]
 defaultValue = "500000000000000000"
 
 [setting.fee_token]
```
barrasso, line 14: `fee_token` should be nonzero as it is set in the constructor

barrasso:
thanks fred, will take this over from here 🫡

auxiliary/OwnedFeeCollector/cannonfile.toml
```diff
@@ -3,9 +3,12 @@ version = "<%= package.version %>"
 description = "Colelcts fees on behalf of an owner"
 
 [setting.owner]
+defaultValue = "0xBE0eB53F46cd790Cd13851d5EFf43D12404d33E8"
+
+[setting.fee_share_recipient]
 defaultValue = "0x0000000000000000000000000000000000000000"
```
barrasso, line 9: `fee_share_recipient` cannot b 0

barrasso (approved):

PR #2292, opened 2024-09-16
M-06 - Risk-Free Trades By Credit Manipulation

leomassazza:


fredsnax:
**Description**

- Orders can be executed after commit time + delay and till the expiry of the order. The price at commit time + delay is used even if the order is executed later than that (somewhere between this timestamp and the expiry timestamp)
- An Order can not be settled if the delegated liquidity by the pool to this market is smaller than the required liquidity to back the open interest + this order sufficiently.

These conditions can be abused for risk-free trades:

- Attacker is an LP of the pool behind the perps market
- The attacker successfully creates an order
- Attacker mints snxUSD to reduce the delegated amt to the perps market
- The order can not be settled as there is no longer enough liquidity to back the perps market sufficiently if this trade goes through
- The attacker waits if the price changes so that the order would be instant profit and if so burns the snxUSD tokens to increase the delegated amount again and settles the order

**Recommendation**

Cancel orders that fail because the delegated amount is smaller than the required credit.

PR #2298, opened 2024-09-17
Make foundry great again

noisekit:


dbeal-eth (approved):
Let's do it.

noisekit:
BFP tests failing regardless of which version of foundry we use (even using the old one that had tests green before does not help). This needs to be investigated separately and will be sorted in a separate PR.

PR #2307, opened 2024-09-24
add warning when no contracts found for storage dump.

dbeal-eth:


noisekit (approved):
Id fail with error maybe but this is good too

PR #2308, opened 2024-09-25
Cannon Update action

noisekit:
The idea is to let us easily trigger a cannon update to the `latest`, `hotfix` or `alpha` version and create a PR that will run all the CI tests. 

This should allow cannon devs to verify there are no breaking changes by doing alpha releases


dbeal-eth (approved):
i think this can be seriously beneficial if we can get this button to the cannon devs, so that they can test against synthetxi ci whenver :+1:

PR #2310, opened 2024-09-26
M-03: change new req margin to use current price instead of fill

fredsnax:
… price to match liquidations


fredsnax:
In `getRequiredMarginWithNewPosition`, `newRequiredMargin` is calculated with fillPrice. This is inaccurate as fillPrice includes a premium/discount, and therefore should only be used for PnL calculations.

Otherwise, a new position might be within the required margin (based on fillPrice) but immediately after settling it could be eligible for liquidation (which calculates margin based on oracle price).

This is also inconsistent with how `oldRequiredMargin` is calculated in the next step, which uses currentPrice.

**Recommendation**

Use currentPrice to calculate `newRequiredMargin`.

markets/perps-market/contracts/storage/AsyncOrder.sol
```diff
@@ -587,7 +587,7 @@ library AsyncOrder {
         // get initial margin requirement for the new position
         (, , runtime.newRequiredMargin, ) = marketConfig.calculateRequiredMargins(
             newPositionSize,
-            fillPrice
+            PerpsPrice.getCurrentPrice(marketId, PerpsPrice.Tolerance.DEFAULT)
```
wasp-in-bush, line 590: this result is used again on line 596. would be more efficient to assign it to a variable and then reuse it.

markets/perps-market/test/integration/Orders/Order.marginValidation.test.ts
```diff
@@ -99,7 +99,7 @@ describe('Orders - margin validation', () => {
     });
 
     it('reverts if not enough margin', async () => {
-      const fillPrice = calculateFillPrice(wei(0), wei(10_000), wei(3), wei(2000));
+      const fillPrice = calculateFillPrice(wei(0), wei(10_000), wei(3), ETH_MARKET_PRICE);
```
wasp-in-bush, line 102: constants are *so much better* than "naked" `wei(10_000)`, `wei(3)`, `wei(2000)` values. 

nj

wasp-in-bush (approved):
Suggested one optimization but otherwise LGTM

Test updates are also great.

Nice work!

PR #2314, opened 2024-09-30
Expand erc7412 errors

dbeal-eth:
hoping to get this audited and out in ultra speed.

fixes https://linear.app/ccsnx/issue/SNX-842/expand-usage-of-new-erc7412-changes-to-all-places-possible


protocol/synthetix/contracts/modules/core/MarketCollateralModule.sol
```diff
@@ -60,15 +60,22 @@ contract MarketCollateralModule is IMarketCollateralModule {
         collateralEntry.amountD18 += systemAmount;
         collateralType.safeTransferFrom(marketData.marketAddress, address(this), tokenAmount);
 
+        (uint256 depositedCollateralValue, bytes memory possibleError1) = marketData
+            .getDepositedCollateralValue();
+        (uint256 reportedDebt, bytes memory possibleError2) = marketData.getReportedDebt();
+
+        RevertUtil.revertIfError(possibleError1);
+        RevertUtil.revertIfError(possibleError2);
```
noisekit, line 68: is it intentional to not combine these errors in a list and revert one by one? (here and in other places)

dbeal-eth: in this case I thought it would be simpler to just return errors one by one since the probability that error 1 and error 2 are caused by the same missing oracle is high.

That being said it probably wouldn't be that hard to combine them in the revert util

also would like to know of other places you are referring to.

protocol/synthetix/contracts/modules/core/MarketCollateralModule.sol
```diff
@@ -107,9 +114,16 @@ contract MarketCollateralModule is IMarketCollateralModule {
         // Account for transferring out collateral
         collateralEntry.amountD18 -= systemAmount;
 
+        (uint256 depositedCollateralValue, bytes memory possibleError1) = marketData
+            .getDepositedCollateralValue();
+        (uint256 reportedDebt, bytes memory possibleError2) = marketData.getReportedDebt();
+
+        RevertUtil.revertIfError(possibleError1);
+        RevertUtil.revertIfError(possibleError2);
```
noisekit, line 122: another place

barrasso (approved):

PR #2319, opened 2024-10-04
Increment version to 3.8.1

0xMithrandir:


barrasso (approved):

noisekit (commented):
This is not right. Publishing guide is in readme.
We need to have versions incremented properly by using lerna and not manually.

PR #2327, opened 2024-10-16
L-01: Unexpected sizeDelta Emitted

barrasso:
**Description**

In the `_settleOrder` function the `MarketUpdated` event emits `runtime.sizeDelta` as the sizeDelta for the market, however this may be inaccurate for the market size when the order moves a position from long to short.

**Recommendation**

Consider whether the `sizeDelta` in the `MarketUpdated` should represent the change in market size or the change in position size. If it should represent the change in market size, then update the emitted value to be the new market size - the original market size.

### DIFF
--- a/markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
+++ b/markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
@@ -82,6 +82,10 @@ contract AsyncOrderSettlementPythModule is
 
         Position.Data storage oldPosition;
 
+        // Load the market before settlement to capture the original market size
+        PerpsMarket.Data storage market = PerpsMarket.loadValid(runtime.marketId);
+        uint256 originalMarketSize = market.size;
+
         // validate order request can be settled; call reverts if not
         (runtime.newPosition, runtime.totalFees, runtime.fillPrice, oldPosition) = asyncOrder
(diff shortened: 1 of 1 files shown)


markets/perps-market/contracts/modules/AsyncOrderSettlementPythModule.sol
```diff
@@ -103,41 +108,74 @@ contract AsyncOrderSettlementPythModule is
...
+        uint256 orderFees = runtime.totalFees - runtime.settlementReward;
+        GlobalPerpsMarketConfiguration.Data storage s = GlobalPerpsMarketConfiguration.load();
+        s.collectFees(orderFees, asyncOrder.request.referrer, factory);
+    }
 
-        /// @dev two events emitted to avoid stack limitations
+    /// @dev Emit settlement events in a helper function to reduce stack depth
+    function _emitSettlementEvents(
+        SettleOrderRuntime memory runtime,
+        AsyncOrder.Data storage asyncOrder
```
wasp-in-bush, line 177: this can be marked as memory too. would be interesting to see gas comparison bc most if not all of these slots are warm.

wasp-in-bush (approved):
One NIT comment but LGTM. 

Are there no tests asserting `marketSizeDelta` is as expected

PR #2329, opened 2024-10-16
L-02: Missing Fee Data In Event

barrasso:
**Description**

In the `_settleOrder` function the `runtime.referralFees` and `runtime.feeCollectorFees` are no longer assigned to from the `collectFees` call.

Therefore these values will be the default 0 value in the `OrderSettled` event.


wasp-in-bush (approved):
LGTM! 
---
It would be nice to see a test for this but given it is an event, very low prio

PR #2330, opened 2024-10-17
M-08: Liquidation rewards for first window is ignored

barrasso:
**Description**

`PerpsAccount.getPossibleLiquidationReward` computes the liquidation reward as `costOfLiquidation * (numOfWindows - 1)`.

There is a wrong assumption that the first liquidation window is already accounted for in the `liquidateAndFlagCost` variable.

It is wrong because when that variable is calculated, only `costOfFlagging` is passed to `keeperReward()` .

In result, the liquidation cost for the first window will not be added towards the total liquidation cost.

**Recommendation**

When calling `keeperReward()` to compute `liquidateAndFlagCost`, pass `costOfFlagging + costOfLiquidation` as second parameter.


wasp-in-bush (approved):
LGTM
---
very concerning no tests break with this change

PR #2335, opened 2024-10-28
Forkability improvements

dbeal-eth:
This PR introduces a new core utility contract, `ForkDetector`, which
provides a way for developer forks (such as Tenderly, Anvil) to
distinguish themselves from the canonical chain safely. This can be
useful as there are some functionalities which are difficult to operate
on a fork (ex. offchain prices or precompiled contracts), and the
behavior in those places can be selectively mocked to prevent
interruption of testing.

Design Considerations:
* The idea of these changes is to provide a one-line way for integrators
  or anyone trying to fork the synthetix system a way to make Synthetix
  fork-friendly setup. In the future, who knows, this functionality could be
  enabled by default.
* Original functionality of the chain can be restored by removing or
  modifying the code at `0x1234123412341234123412341234123412341234`, in
  case your tests require it

Specific changes:
* add `ForkDetector` library
* in op gas price oracle, replace calls to the OP precompiled contract
  with mock values or skip
* in arb gas price oracle, replace calls to the arbitrum precompiled
  contract with mock values or skip
* in pyth ERC7412 adapter proxy, use whatever the latest price available
  on-chain is regardless of its staleness. Additionally, prices can be
  overridden by calling the contract (note: it was previously possible
  to simply switch out the oracle node on the oracle manager as well,
  but sometimes this lever is much more convenient to pull)

Tests to be written tomorrow


noisekit (approved):
❤️

noisekit (approved):

PR #2338, opened 2024-11-20
when checking capacity locked, also add the deposited collateral value

dbeal-eth:
this is a much more correct calculation for determining if capacity is sufficient or not.

Its good that this issue turned out to be so benign.


dbeal-eth:
the expected breaking changes are that ERC7412 is going to be affected for the following functions:

* VaultModule.delegateCollateral: function already depends on the same offchain prices prior to capacity locked check
* PoolModule.setPoolConfiguration: same as `delegateCollateral
* MarketManagerModule.isMarketCapacityLocked: as far as I know nobody actually uses this view function on an actual app,

noisekit (approved):

PR #2339, opened 2024-11-21
SIP-406 Add trading enabled feature flag

0xMithrandir:
Adds a feature flag `tradingEnabled` which has to be enabled to be trading/wrapping synths.

https://sips.synthetix.io/sips/sip-406/


markets/spot-market/contracts/modules/AtomicOrderModule.sol
```diff
@@ -31,6 +34,11 @@ contract AtomicOrderModule is IAtomicOrderModule {
         SpotMarketFactory.Data storage spotMarketFactory = SpotMarketFactory.load();
         spotMarketFactory.validateMarket(marketId);
 
+        ITokenModule synth = SynthUtil.getToken(marketId);
+        FeatureFlag.ensureAccessToFeature(
+            bytes32(abi.encodePacked(Flags.TRADING_ENABLED, marketId))
```
dbeal-eth, line 39: note: while this works fine for progmatic use cases, afaict its not like the human readable text of the flag would be:

```
tradingEnabledSynthId75
```

if the pool ID was 75. we would need to first encode the integer as a string i am pretty sure.

dbeal-eth (approved):
please see comments

while we are at it, do we also have a feature flag to turn on/off *all* markets?

and do we have a feature flag to only turn off withdrawals/deposits? I feel like this is a level of granularity we might actuall ywant to have, but I understand we probably don't want to expand the scope.

0xMithrandir:
> please see comments
> 
> while we are at it, do we also have a feature flag to turn on/off _all_ markets?
> 
> and do we have a feature flag to only turn off withdrawals/deposits? I feel like this is a level of granularity we might actuall ywant to have, but I understand we probably don't want to expand the scope.

This SIP is quite specific so I'd like to avoid expanding the scope but maybe @kaleb-keny  can chime in here when he gets back on adding more flags.

markets/spot-market/contracts/modules/AtomicOrderModule.sol
0xMithrandir: You are right, it's hard to decipher which `FeatureUnavailable` is thrown with this setup.

0xMithrandir: I've updated this line to encode the `marketId` to string as well.

kaleb-keny:
> > please see comments
> > while we are at it, do we also have a feature flag to turn on/off _all_ markets?
> > and do we have a feature flag to only turn off withdrawals/deposits? I feel like this is a level of granularity we might actuall ywant to have, but I understand we probably don't want to expand the scope.
> 
> This SIP is quite specific so I'd like to avoid expanding the scope but maybe @kaleb-keny can chime in here when he gets back on adding more flags.

hey @0xMithrandir i am happy to alter the sip, i think it makes sense to add a master control for killing the entire spot market...
For deposit withdrawal, not sure what we are depositing withdrawing, since this is more of an exchange part of the code base... I think this is possible to in asyncOrders with that isEnabled flag on the settle strategy... But not possible with atomic orders and wrappers, hence this sip...

So it would be more like a `atomicOrderFeatureFlag=1/0` and `wrapperFlag=1/0`
Let me know of your thoughts and happy to make the necessary changes to the sip

0xMithrandir:
> > > please see comments
> > > while we are at it, do we also have a feature flag to turn on/off _all_ markets?
> > > and do we have a feature flag to only turn off withdrawals/deposits? I feel like this is a level of granularity we might actuall ywant to have, but I understand we probably don't want to expand the scope.
> > 
> > 
> > This SIP is quite specific so I'd like to avoid expanding the scope but maybe @kaleb-keny can chime in here when he gets back on adding more flags.
> 
> hey @0xMithrandir i am happy to alter the sip, i think it makes sense to add a master control for killing the entire spot market... For deposit withdrawal, not sure what we are depositing withdrawing, since this is more of an exchange part of the code base... I think this is possible to in asyncOrders with that isEnabled flag on the settle strategy... But not possible with atomic orders and wrappers, hence this sip...
> 
> So it would be more like a `atomicOrderFeatureFlag=1/0` and `wrapperFlag=1/0` Let me know of your thoughts and happy to make the necessary changes to the sip

You'd prefer having separate flags for each module like:
- atomicOrdersEnabled per synth
- wrapperEnabled per synth
- general tradingEnabled flag that is for all markets?

kaleb-keny:
replied in discord, but the context is 
- `spotMarketEnabled` would enable the entire spot market for all synths, i.e. atomic, async and wrappers
- `atomicOrdersEnabled` would enable atomic orders on a specific synth, where by the non sUSD need to be enabled for the transaction to go through
- `wrapperEnabled` would enable wraps & unwraps on a specific synth, where by the non sUSD need to be enabled for the transaction to go through
As for async orders per synth they can be enabled / disabled with the settle strategy flag. So no changes required on those
I will update the sip accordingly

0xMithrandir:
I've updated the new feature flags, I'll add more specific tests but lmk if this is looking good now @dbeal-eth

dbeal-eth:
@0xMithrandir lgtm! do we have a audit completed for this change?

0xMithrandir:
> @0xMithrandir lgtm! do we have a audit completed for this change? normally we wait for audit before merging.

I'll submit it today for audit, after their approval it can be merged.

PR #2343, opened 2024-12-04
staleness circuit breaker node responds to fork

dbeal-eth:
does not follow staleness check if the node is a fork


noisekit (approved):

protocol/oracle-manager/test/integration/nodes/StalenessCircuitBreakerNode.OffchainLookup.test.ts
```diff
@@ -57,6 +57,27 @@ describe('StalenessCircuitBreakerNode: Offchain Lookup orchestration', function
...
+      revertNodeId,
+    ]);
+    // enable fork mode
+    await getProvider().send('hardhat_setCode', [
+      '0x1234123412341234123412341234123412341234',
+      ethers.utils.hexlify(ethers.utils.toUtf8Bytes('FORK')),
+    ]);
+    const parentOutput = await NodeModule.process(staleNodeId);
+    const nodeOutput = await NodeModule.process(nodeId);
+    assertBn.equal(parentOutput.price, nodeOutput.price);
```
ProblematicP, line 78: timestamp should be equal too right? check for that too

protocol/oracle-manager/contracts/nodes/pyth/PythOffchainLookupNode.sol
```diff
@@ -64,6 +64,8 @@ library PythOffchainLookupNode {
                 priceIds
             )
         );
+
+        return (nodeOutput, possibleError);
```
ProblematicP, line 68: i dont get what this function does. WHere is nodeOutput modified?

dbeal-eth: `nodeOutput` is not modified. However, this is apparently required to silence a solidity compiler warning about `nodeOutput` not being set, to explicitly return it.

this function is quite ackward, I agree. maybe we can rethink this implementation a bit later.

noisekit (approved):

PR #2344, opened 2024-12-04
use `updatePriceFeeds` without if necessary

dbeal-eth:
this ensures price feeds can always be updated without getting an annoying error that really doesn't matter


dbeal-eth:
yea, basically we are making this change because its required for proper client functioning.

the gas savings are probably pretty minimal because the price update pushes are only sent when the client determines they are needed. the only case where the gas "waste" would likely happen is in the somewhat unlikely case that a price update is pushed before the client is able to submit their txn, in which case the clients update should  be silently ignored.

auxiliary/PythERC7412Wrapper/contracts/PythERC7412Wrapper.sol
```diff
@@ -122,29 +122,15 @@ contract PythERC7412Wrapper is IERC7412 {
...
-            }
-
-            try
-                pyth.updatePriceFeedsIfNecessary{value: msg.value}(
-                    updateData,
-                    priceIds,
-                    publishTimes
-                )
-            {} catch (bytes memory reason) {
+            try pyth.updatePriceFeeds{value: msg.value}(updateData) {} catch (bytes memory reason) {
```
noisekit, line 133: yeah this is the one we now use on frontend to avoid reverts

PR #2348, opened 2024-12-13
Cancellation cost calculation for keepers

0xMithrandir:


dbeal-eth (approved):
might want to note that the `setCosts` call signature has changed, and this will likely required deployments repo changes on the next release.

0xMithrandir:
> might want to note that the `setCosts` call signature has changed, and this will likely required deployments repo changes on the next release.

@kaleb-keny actually simplified this PR quite a lot, we are now just reusing settlementRewardCost instead of adding cancel gas units.

PR #2354, opened 2025-01-19
SIP-420: Treasury market

dbeal-eth:
this implements sip-420.

notes:
* in addition to implementing the actual contract for SIP-420, it was also necessary to add the `migrateDelegation` function to the core system. This is becuase it needs to be possible for a user to actually migrate their account from the current v3 SC pool without repaying their debt. (on optimism the legacymrket can be directly conneted)
* the treasury market is designed to be able to work with other markets, preserving the portability of liquidity
* testing a new way of testing contracts in the v3 repo. we generate a foundry script with cannon, and then run tests against the generated script. So far this seems to be working well. Coverage and everything works wrt tests. This script can even be copied anywhere since it has no dependencies (outside of forge-std).

some deployment invariants exist WRT the creation of the pool (ex. limit to only one collateral type, no minting through v3, etc.). For more information see the corresponding deployments PR https://github.com/Synthetixio/synthetix-deployments/pull/619


markets/perps-market/storage.dump.json
```diff
@@ -1,63 +1,35 @@
 {
-  "contracts/modules/LiquidationModule.sol:LiquidationModule": {
-    "name": "LiquidationModule",
+  "contracts/modules/BookOrderModule.sol:BookOrderModule": {
```
noisekit, line 2: unexpected change

protocol/oracle-manager/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@synthetixio/oracle-manager",
-  "version": "3.10.1",
+  "version": "3.10.2",
```
noisekit, line 3: this looks unexpected

markets/treasury-market/script/Deploy.sol
```diff
@@ -0,0 +1,1076 @@
+// SPDX-License-Identifier: UNLICENSED
```
noisekit, line 1: this script is strange, includes a lot of things like core system deployement and a lot of extras, would expect it to only have treasury market.

markets/treasury-market/test/TreasuryMarket.test.sol
```diff
@@ -0,0 +1,427 @@
...
+
+        // delegate another account to cause the avg c-ratio to shoot way up
+        v3System.delegateCollateral(
+            accountId + 1,
+            poolId,
+            address(collateralToken),
+            5 ether,
+            1 ether
+        );
+        market.saddle(accountId);
```
noisekit, line 131: `saddle`? is it a typo?

markets/treasury-market/contracts/interfaces/external/IV3CoreProxy.sol
```diff
@@ -0,0 +1,732 @@
+// SPDX-License-Identifier: UNLICENSED
+// !! THIS FILE WAS AUTOGENERATED BY abi-to-sol v0.7.1. SEE SOURCE BELOW. !!
+pragma solidity ^0.8.4;
+
+interface IV3CoreProxy {
```
noisekit, line 5: we can include specific modules from node modules and avoid generating interface that does not automatically follow actual contracts

noisekit: smth like 
```ts

import {ERC2771Context} from "@synthetixio/core-contracts/contracts/utils/ERC2771Context.sol";
import {ICollateralModule} from "@synthetixio/main/contracts/interfaces/ICollateralModule.sol";
import {IIssueUSDModule} from "@synthetixio/main/contracts/interfaces/IIssueUSDModule.sol";
import {IVaultModule} from "@synthetixio/main/contracts/interfaces/IVaultModule.sol";
import {IAccountModule} from "@synthetixio/main/contracts/interfaces/IAccountModule.sol";
import {IERC20} from "@synthetixio/core-contracts/contracts/interfaces/IERC20.sol";
import {IERC721Receiver} from "@synthetixio/core-contracts/contracts/interfaces/IERC721Receiver.sol";
import {IERC721Enumerable} from "@synthetixio/core-contracts/contracts/interfaces/IERC721Enumerable.sol";
import {IERC721} from "@synthetixio/core-contracts/contracts/interfaces/IERC721.sol";
import {IWrapperModule} from "@synthetixio/spot-market/contracts/interfaces/IWrapperModule.sol";
import {IAtomicOrderModule} from "@synthetixio/spot-market/contracts/interfaces/IAtomicOrderModule.sol";
import {Price} from "@synthetixio/spot-market/contracts/storage/Price.sol";
import {IMarketManagerModule} from "@synthetixio/main/contracts/interfaces/IMarketManagerModule.sol";

contract PositionManagerAndromedaUSDC {
    error NotEnoughAllowance(
        address walletAddress, address tokenAddress, uint256 requiredAllowance, uint256 availableAllowance
    );
```

markets/treasury-market/script/Deploy.sol
dbeal-eth: it's a standalone script that deploys everything needed to get a testable treasury market from scratch. so this is as expected.

markets/treasury-market/cannonfile.toml
```diff
@@ -0,0 +1,83 @@
...
+factory.Proxy.artifact = "SynthetixTreasuryProxy"
+factory.Proxy.event = "Upgraded"
+factory.Proxy.arg = 0
+factory.Proxy.highlight = true
+
+[invoke.setFeatureFlagMarket]
+target = ["v3.CoreProxy"]
+func = "addToFeatureFlagAllowlist"
+args = [
+    "0x72656769737465724d61726b6574000000000000000000000000000000000000",
```
noisekit, line 62: why not use smth like `formatBytes32String("registerMarket")` here?

markets/treasury-market/test/TreasuryMarket.test.sol
dbeal-eth: not a typo 

definition 2
image

markets/treasury-market/contracts/interfaces/external/IV3CoreProxy.sol
dbeal-eth: i can do that. its just kind of inconvenient to pull invidual contracts and cast them all the time, I strongly prefer things this way as they are.

markets/treasury-market/cannonfile.toml
dbeal-eth: because I copied and pasted from somewhere else and I thik that code didnt do that lol. will fix.

markets/perps-market/storage.dump.json
dbeal-eth: will fix

protocol/oracle-manager/package.json
dbeal-eth: for some reason oracle manager version was not incremented on the recent 3.10.2 change on main https://github.com/Synthetixio/synthetix-v3/commit/2bf9865f34af6613366302f244d72675a6bb079a

so I had to do it here

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -0,0 +1,375 @@
...
+        int256 targetDebt = vaultDebtValue;
+        if (artificialDebt == 0) {
+            // if there is no artificial debt, it means that this is the first user and we should create a bunch
+            // to bring to low cratio
+            targetDebt = vaultCollateralValue.divDecimal(targetCratio).toInt();
+        } else if (vaultCollateralValue > accountCollateralValue) {
+            targetDebt =
+                (accountCollateralValue.toInt() * vaultDebtValue) /
+                (vaultCollateralValue + saddledCollateral[accountId] - accountCollateralValue)
+                    .toInt();
```
CalabashSquash, line 160: Could you explain the intuition behind this formula?

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -0,0 +1,375 @@
...
+            revert ParameterError.InvalidParameter("amount", "must be less than current loan");
+        } else if (amount < currentLoan) {
+            v3System.depositMarketUsd(marketId, sender, currentLoan - amount);
+        } else {
+            return; // nothing to do
+        }
+
+        // fractionally modify the original loan amount--this will continue the repayment schedule where the user left off without resetting it
+        loans[accountId].loanAmount = ((loans[accountId].loanAmount * amount) / currentLoan)
+            .to128();
```
CalabashSquash, line 265: As discussed, Make sure this is thoroughly tested

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -0,0 +1,375 @@
...
+        (, uint256 vaultCollateralValue) = v3System.getVaultCollateral(poolId, collateralToken);
+        int256 vaultDebtValue = v3System.getVaultDebt(poolId, collateralToken);
+
+        int256 targetDebt = vaultCollateralValue.divDecimal(targetCratio).toInt();
+
+        if (artificialDebt > vaultDebtValue - targetDebt) {
+            artificialDebt += targetDebt - vaultDebtValue;
+        } else {
+            return;
+        }
```
CalabashSquash, line 343: As discussed, we should set `artificialDebt` to 0

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -0,0 +1,392 @@
...
+            .getPosition(accountId, poolId, collateralToken);
+
+        // get the actual collateralization of the pool now
+        (, uint256 vaultCollateralValue) = v3System.getVaultCollateral(poolId, collateralToken);
+        int256 vaultDebtValue = v3System.getVaultDebt(poolId, collateralToken);
+
+        //
+        uint256 newlySaddledValue = accountCollateralValue -
+            (saddledCollateral[accountId] * accountCollateralValue) /
+            accountCollateral;
```
CalabashSquash, line 153: nit: should there be a comment here? I think it would be worth adding anyway

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    function loanedAmount(uint128 accountId) external view returns (uint256) {
+        return _loanedAmount(accountId, block.timestamp);
+    }
+
+    function setDebtDecayFunction(uint32 power, uint32 time) external onlyOwner {
+        if (power > 100) {
+            revert ParameterError.InvalidParameter("debtDecayPower", "too high");
+        }
```
CalabashSquash, line 292: Nit
```suggestion
        if (power > 100) {
            revert ParameterError.InvalidParameter("power", "too high");
        }
```

CalabashSquash (commented):
just two nitpicks

markets/treasury-market/contracts/interfaces/external/IV3CoreProxy.sol
```diff
@@ -0,0 +1,739 @@
...
```
noisekit, line 5: We no longer need `IV3CoreProxy.sol` and can use `@synthetixio/v3-contracts/1-main/CoreProxy.sol`

But we can replace this later

noisekit (approved):
lgtm

markets/treasury-market/test/TreasuryMarket.test.sol
```diff
@@ -0,0 +1,885 @@
...
+        (uint64 startTime, , , ) = market.loans(accountId);
+
+        assertEq(market.loanedAmount(accountId), 1 ether);
+        vm.warp(startTime + 1000000);
+        assertEq(market.loanedAmount(accountId), 1 ether);
+    }
+
+    function test_LoanDecayLinearConfig() external {
+        vm.warp(100000000);
+        vm.expectEmit();
```
noisekit, line 354: This is not a test. Need to expect specific emit and call contract function below that emits it. Not emit the event in the test itself.

dbeal-eth: ...how is the check here not matching how it works in the examples given? https://book.getfoundry.sh/cheatcodes/expect-emit#examples

we don't check the emitting address, but we check the event we expect emitted by emitting in the following line

noisekit: Ok sorry. This is really weird way of testing that I would never guess works like that by just reading code without reading documentation. 
Turns out you are expected to emit the event as an example what expect should look for in the next contract call. Expecting errors works in a completely different way (very conventional and intuitive).
It is kinda cool tho ngl, now I want expect revert to work with revert example like that

PR #2355, opened 2025-01-20
add delegation migrate function

dbeal-eth:
as part of SIP-420, it is necessary to have a function to move user's collateral and debt to a new pool without repaying debt

### DIFF
--- a/protocol/synthetix/contracts/modules/core/VaultModule.sol
+++ b/protocol/synthetix/contracts/modules/core/VaultModule.sol
@@ -36,6 +36,7 @@ contract VaultModule is IVaultModule {
     using SafeCastI256 for int256;
 
     bytes32 private constant _DELEGATE_FEATURE_FLAG = "delegateCollateral";
+    bytes32 private constant _MIGRATE_DELEGATION_FEATURE_FLAG = "migrateDelegation";
 
     /**
      * @inheritdoc IVaultModule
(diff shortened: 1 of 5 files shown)


CalabashSquash (approved):

protocol/synthetix/contracts/modules/core/VaultModule.sol
```diff
@@ -158,6 +154,109 @@ contract VaultModule is IVaultModule {
...
+        // old pool must not be left in capacity locked state
+        _verifyNotCapacityLocked(oldPoolId);
+
+        // newly migrated position must not be liquidatable
+        uint256 minIssuanceRatioD18 = Pool
+            .loadExisting(oldPoolId)
+            .collateralConfigurations[collateralType]
+            .issuanceRatioD18;
+
+        CollateralConfiguration.load(collateralType).verifyIssuanceRatio(
```
noisekit, line 228: do we need to add similar check for `newPoolId` or is it already performed when `recalculateVaultCollateral` or `assignDebtToAccount` are called?

noisekit (approved):

protocol/synthetix/contracts/modules/core/VaultModule.sol
dbeal-eth: the issuance ratio for the account does not change when the user moves from the old pool to the new pool, so we only need to do the check once and whenever it is convenient for the computation.

PR #2362, opened 2025-02-26
420 phase1

dbeal-eth:
adds ability to distribute token rewards to stakers which vest over a year with penalty, assuming they have no debt.

once complete, the amount rewarded is transparently added to their v3 account.


markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -198,14 +197,60 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
...
+                loans[accountId] = LoanInfo(
+                    uint64(block.timestamp),
+                    debtDecayPower,
+                    debtDecayTime,
+                    accountDebt.toUint().to128()
+                );
+                emit LoanAdjusted(accountId, accountDebt.toUint(), 0);
+            } else {
+                // this depositor is eligible for possible rewards
+                //DepositRewardConfiguration[] memory drc = depositRewardConfigurations;
```
noisekit, line 211:
```suggestion
```

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -198,14 +197,60 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
...
+                    if (rewardAmount > availableDepositRewards[config.token]) {
+                        revert InsufficientAvailableReward(
+                            config.token,
+                            rewardAmount,
+                            availableDepositRewards[config.token]
+                        );
+                    }
+
+                    // stack was too deep to set this as a local variable. annoying.
+                    //LoanInfo memory depositRewardData =
```
noisekit, line 226:
```suggestion
```

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -198,14 +197,60 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
...
+                    LoanInfo(
+                        uint64(block.timestamp),
+                        config.power,
+                        config.duration,
+                        rewardAmount.to128()
+                    );
```
noisekit, line 232:
```suggestion
```

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -198,14 +197,60 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
...
+                );
+                emit LoanAdjusted(accountId, accountDebt.toUint(), 0);
+            } else {
+                // this depositor is eligible for possible rewards
+                //DepositRewardConfiguration[] memory drc = depositRewardConfigurations;
+                for (uint256 i = 0; i < depositRewardConfigurations.length; i++) {
+                    DepositRewardConfiguration memory config = depositRewardConfigurations[i];
+                    uint256 rewardAmount = (((accountCollateral *
+                        oracleManager.process(config.valueRatioOracle).price.toUint()) / 1 ether) *
+                        config.percent) / 1 ether;
```
noisekit, line 216: I assume that config.percent here is in 0...1 range, so maybe the name "percent" is not the best.

markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -198,14 +197,60 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
...
+
+                    emit DepositRewardIssued(
+                        accountId,
+                        config.token,
+                        LoanInfo(
+                            uint64(block.timestamp),
+                            config.power,
+                            config.duration,
+                            rewardAmount.to128()
+                        )
```
noisekit, line 250:
```suggestion
                        depositRewards[accountId][config.token]
```

noisekit (approved):

markets/treasury-market/contracts/TreasuryMarket.sol
dbeal-eth: i left the comment in because i would prefer if the coed could look like this,but it cant because of solidity stack too deep stuff. idk seemed right to keep it.

markets/treasury-market/contracts/TreasuryMarket.sol
dbeal-eth: same as below comment

markets/treasury-market/contracts/TreasuryMarket.sol
dbeal-eth: yea probably right. maybe "ratio" is a better name?

PR #2364, opened 2025-03-05
add constructor args so that cannon can verify proxy contracts

dbeal-eth:
should fix an error during contract verification with `cannon verify` where it doesn't have the correct constructor arguments to do the verification.


noisekit (approved):
Great

PR #2368, opened 2025-03-18
add new liquidation to treasury capability

dbeal-eth:
allows for sending of liquidation rewards and debts to another account in another pool, rather than to the liquidator or socialized.

still need to work out how to automatically claim a new position for users


protocol/synthetix/contracts/modules/core/LiquidationModule.sol
```diff
@@ -134,6 +147,111 @@ contract LiquidationModule is ILiquidationModule {
...
+            liquidationData.debtLiquidated.toInt()
+        );
+
+        // Remove the debt assigned to the liquidated account
+
+        // The collateral is reduced by `amountRewarded`, so we need to reduce the stablecoins capacity available to the markets
+        Pool.load(poolId).recalculateVaultCollateral(collateralType);
+
+        // Send amountRewarded to the specified account
+        emit Liquidation(
```
noisekit, line 245: Would be good to store this information per account in storage so we can use it later in frontends

noisekit (approved):

noisekit:
Base off this branch from "fixes" branch to get the tests running

PR #2371, opened 2025-03-27
ability to cap rewards granted for deposits at certain percentages

dbeal-eth:


noisekit (approved):
I still think we can simplify it to only have one cap and not an array, but other than that - looks good

PR #2374, opened 2025-04-26
few final fixes

dbeal-eth:
* fix remaining debug code left in the TM contract
* make the staking contract upgradable (which also means it needs to be ownable)


markets/treasury-market/contracts/TreasuryMarket.sol
```diff
@@ -397,9 +397,6 @@ contract TreasuryMarket is ITreasuryMarket, Ownable, UUPSImplementation, IMarket
             )
         ) {
             // if their loan hasnt been repaid by reset time, it gets completely reset
-
-            //emit UpdateAuxTokenRequirement(auxTokenRequiredRatios[firstInsufficientIdx].timestamp, 0);
-            emit UpdateAuxTokenRequirement(loanLastUpdate, block.timestamp);
```
noisekit, line 402: this intended?

noisekit (approved):

PR #2377, opened 2025-05-27
Collateral Deprecation

dbeal-eth:
ability to deprecate deposited collaterals and move them to a new system for claim. The owner of the contract can specify a deprecation target, and collateral will be sent to this target for further deprecation related processing outside of v3.

notes:
* once a collateral is deprecated, it is also automatically disabled, which will prevent further usage of it. It should never be re-enabled or configured again.
* the owner is expected to act rationally wrt to usage of deprecation (ex. do not use deprecated collaterals to fund markets which are actively needed for protocol operation)

### DIFF
--- a/protocol/synthetix/test/integration/modules/core/CollateralConfigurationModule/CollateralConfigurationModule.config.test.ts
+++ b/protocol/synthetix/test/integration/modules/core/CollateralConfigurationModule/CollateralConfigurationModule.config.test.ts
@@ -217,7 +217,7 @@ describe('CollateralModule', function () {
             it('fails with correct error when trying to withdraw', async () => {
               await assertRevert(
                 systems().Core.connect(user1).withdraw(1, AnotherCollateral.address, 1000),
-                'CollateralDeprecated',
+                'DeprecatedDeposit',
                 systems().Core
               );
             });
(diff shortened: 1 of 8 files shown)


protocol/synthetix/test/integration/modules/core/CollateralConfigurationModule/CollateralConfigurationModule.config.test.ts
```diff
@@ -160,6 +166,63 @@ describe('CollateralModule', function () {
...
+            before('deposit and deprecate the collateral', async () => {
+              await AnotherCollateral.mint(await user1.getAddress(), 1000);
+              await AnotherCollateral.connect(user1).approve(systems().Core.address, 1000);
+              await systems().Core.connect(user1).deposit(1, AnotherCollateral.address, 1000);
+              const tx = await systems()
+                .Core.connect(systemOwner)
+                .deprecateCollateral(AnotherCollateral.address, await systemOwner.getAddress());
+              await tx.wait();
+            });
+            it('sends to deprecation target', async () => {
```
noisekit, line 209: add a check that depositing is diabled

dbeal-eth: the check is on the other branch of the tests, figured it may be a bit redundant to copy them since the same code executes either way.

